Issue #1, opened 2020-01-22
Project Creation


hwong0305:
1. Next.js template
2. Typescript
3. ReadMe

hwong0305:
Pushed boilerplate setup into `initial` branch

hwong0305:
Closed by PR #2

PR #2, opened 2020-01-23
Initial Template Setup

hwong0305:
#1 

- Setup Next.JS with typescript.
- Created EsLint configuration


tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "jsx": "preserve",
+    "lib": ["dom", "es2017"],
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noEmit": true,
+    "noFallthroughCasesInSwitch": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
```
bryanjenningz: These compiler options look pretty good. Is there a reason why `"skipLibCheck": true` is being used?

hwong0305: There's no particular reason. I kept the default settings generated by next.js.

PR #11, opened 2020-02-02
Grid Layout

Wolfy64:
**This is a test to compare a grid layout between TailwindCSS and Grid CSS** 
[Live Demo: https://c0d3.contactdewulf.now.sh/grid](https://c0d3.contactdewulf.now.sh/grid)

Screen Shot 2020-02-02 at 1 12 11 AM

I setup a  12 column grid that I think we should implement and wrap every page with it.
- In light grey I used CSS GRID ❤️ _(I would prefer use module.css instead of css-in-js for clarity)_
- In dark grey I used TailwindCSS 🤔 _(I did not use any css with Tailwind)_

Please look at the code and give me your feedback.

**Summary:**
I feel have more control with CSS Grid by only set `grid-column: start / end` in my component it make really easy to place it. 
There is less `<div />` and less `className` the code seems easier to read.
Tailwind is good to set up quickly some basic CSS rule.

ps: For some reason `module.css` or `import style-jsx` from another file was not working... [The doc](https://nextjs.org/docs/basic-features/built-in-css-support#adding-component-level-css) say there are supported  out the box.

### DIFF
--- a/components/UI/Layout.tsx
+++ b/components/UI/Layout.tsx
@@ -1,0 +1,75 @@
+import React from 'react'
+
+const Layout: React.FC = ({ children }) => {
+  return (
+    <div className="container">
+      {children}
+      <style jsx>{`
+        .container {
+          display: grid;
+          grid-template-columns: repeat(12, 1fr);
+          grid-gap: 1em;
+          max-width: 1440px;
+          margin: auto;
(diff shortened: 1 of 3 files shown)


Wolfy64:
We opted to use Bootstrap.
Following this PR #12

PR #12, opened 2020-02-04
Landing Page Demo with Bootstrap

hwong0305:
https://c0d3com.hwong0305.now.sh


package.json
```diff
@@ -13,8 +13,11 @@
   },
   "dependencies": {
     "@zeit/next-css": "^1.0.1",
```
songz: remove since not using

PR #19, opened 2020-02-08
Landing Page - Part 3

hwong0305:
- Added Journey Section
- Added Footer
- Added Link to Chat in NavBar


components/LandingPage.tsx
```diff
@@ -2,97 +2,216 @@ import * as React from 'react'
...
+                reviewed curriculum and our community chat. You will be able to
+                setup c0d3.com CLI which will allow you to submit code for
+                review. The CLI allows you to code in Javascript from any code
+                editor of your choice.
+              </p>
+            </div>
+          </div>
+          <div className="row mt-5 pb-5">
+            <div className="col-md-5 text-center">
+              <img src="journey-02.svg" className="mb-3 img-fluid" />
```
bryanjenningz: missing an alt tag here

components/LandingPage.tsx
```diff
@@ -2,97 +2,216 @@ import * as React from 'react'
...
+          <h1
+            id="journey"
+            className="mt-5 display-4 font-weight-bold text-center"
+          >
+            Your Learning Journey
+          </h1>
+          <h4
+            className="mt-5 font-weight-light col-md-8 offset-md-2 pb-5 text-center"
+            style={{ lineHeight: 1.5 }}
+          >
```
bryanjenningz: I think this would be a <p> instead of <h4>, it seems more like a paragraph than a heading.

components/LandingPage.tsx
bryanjenningz: if you don't want to have an any alt, then use an empty alt attribute like this `alt=""` so that screen readers don't read out "journey-02.svg"

components/LandingPage.tsx
```diff
@@ -2,97 +2,216 @@ import * as React from 'react'
...
-          <img src="header-01.svg" alt="Hero Image"></img>
+    <React.Fragment>
+      <div className="container text-center pb-5">
+        <div className="row justify-center mt-5">
+          <div className="col-md-6 offset-md-3">
+            <img
+              src="header-01.svg"
+              alt="Hero Image"
+              className="img-fluid"
+            ></img>
```
bryanjenningz: use `` instead of `</img>`

components/LandingPage.tsx
```diff
@@ -2,97 +2,216 @@ import * as React from 'react'
...
+          >
             Our learning process is interactive and follows the same practice as
             a well functioning engineering team. From the first line of code you
             write, you will be code reviewed by engineers to ensure you are
-            always following industry best practices.
+            always following industry best practice
           </h4>
+          <div className="row mt-5 pb-5">
+            <div className="col-md-5 offset-md-2 order-md-2 text-center">
+              <img src="journey-01.svg" className="mb-3 img-fluid"></img>
```
bryanjenningz: I think screen readers might read "journey-01.svg". If you want to silence this, you can add an empty alt attribute like this: `alt=""`

Quote: "You should use CSS background images for decorative images, but if you must use HTML, add a blank alt="". If the image isn't part of the content, a screen reader shouldn't waste time reading it."

source: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Images_in_HTML

components/LandingPage.tsx
hwong0305: I changed it to a p and I added a `font-size` as inline style. Please let me know if there's any better way to do this.

components/LandingPage.tsx
hwong0305: Fixed.

components/LandingPage.tsx
hwong0305: I decided to add `alt` as the same as the section header. I can't seem to decide whether it should be `alt=""` or `alt="title of section"`. 

I didn't want to change it use background image since each image has a different height and I would have to specify the height of each div.

components/LandingPage.tsx
bryanjenningz: Since we're using Bootstrap, you can use `<p className="h4">` if you want the font-size to be the same size.

Source: https://getbootstrap.com/docs/4.0/content/typography/#headings

components/LandingPage.tsx
bryanjenningz: I'm not 100% sure either, but I think `alt=""` might be good because I think the images seem like redundant content to a screen reader if they're the same as the headings.

I think it's fine to not use a background image and instead use `alt=""`.

public/journey-01.svg
```diff
@@ -0,0 +1,53 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="403" height="293" viewBox="0 0 403 293">
```
bryanjenningz: We don't have to do this now but I think it might be useful to minify the SVGs in the future. We can probably use something like svgo to do that. This can possibly make the page load faster, but it's not a big deal right now, we can do this later. https://github.com/svg/svgo

components/LandingPage.tsx
hwong0305: Thanks! I was looking for that section...

components/LandingPage.tsx
hwong0305: Thanks. I just changed to `alt=""`

public/journey-01.svg
hwong0305: I made the change and added the minify step to the autofix script.

PR #21, opened 2020-02-09
Signup - Card Layout

Wolfy64:
Before to start the Signup Component I created the **Card Layout** that will be re-usable with other components like _Signin_  _Success_ ...etc


components/Card.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { ReactComponent as Checked } from '../assets/images/checked.svg'
+
+type Props = {
+  children?: React.ReactNode
+  success?: boolean
+  text?: string
+  title: string
```
songz: I would also accept variable columns in case we want to customize it in the future:

```
col: number
offset: number
```

components/Card.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+const Card: React.FC<Props> = props => (
+  <div className="row mt-5">
+    <div className="col-sm-5 m-auto" style={{ maxWidth: '604px' }}>
```
songz: `maxWidth` -> I don't think we should hardcode a width for a reusable component.

songz:
```
const colClass = `col-sm-${props.col || 6}`
const offsetClass = `col-offset-${props.offset || 0}`
```

Wolfy64: I added it to follow  the actual design mockup... I will remove it
Screen Shot 2020-02-08 at 11 09 42 PM

Wolfy64: Offset do not work with margin- auto (`m-auto`) and do not center properly in the middle.
I improved the responsive of the **Card** and instead of just changing `col` and `offset` user can use his own Boostrap classes.

components/Card.tsx
```diff
@@ -0,0 +1,31 @@
...
+  children?: React.ReactNode
+  success?: boolean
+  text?: string
+  classes?: string
+  title: string
+}
+
+const Card: React.FC<Props> = props => {
+  const classes = `col-sm-${props.classes ||
+    'col-sm-8 col-md-7 col-lg-6 col-xl-4 m-auto'}`
```
songz: so if `props.classes` is null, then `classes` variable is:

`col-sm-col-sm8` <- what is that?

assets/images/checked.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
```
bryanjenningz: you could minify this SVG with svgo

Wolfy64: I didn't know that feature I will take a look thank Bryan

PR #22, opened 2020-02-10
Layout component

Wolfy64:
Add Layout Component that should wrap all pages except the landingPage.
Setup the default Background
Add viewport addon for StoryBook
Fix the nav bar on big screen:
Screen Shot 2020-02-09 at 7 15 05 PM


.storybook/main.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
-  addons: ['@storybook/addon-actions/register']
+  addons: [
+    '@storybook/addon-actions/register',
+    '@storybook/addon-viewport/register'
```
songz: NICE!!!

This allows us to view on mobile

__tests__/pages/components/__snapshots__/Landing.js.snap
```diff
@@ -2,71 +2,77 @@
...
+        </a>
+        <a
+          className="nav-item nav-link"
+          href="#journey"
+        >
+          Journey
+        </a>
+        <a
+          className="nav-item nav-link"
+          href="https://chat.c0d3.com"
```
songz: hmmm... `help` should really me an email to `song@garagescript.org` temporarily. In the future, it should go to `contact@c0d3.com` that will get posted into our chatroom for us to respond back. but until then.... `chat.c0d3.com` does not help because they don't have account.

components/LandingPage.tsx
```diff
@@ -199,12 +199,12 @@ const LandingPage: React.FC = () => {
           Get Started
         </a>
       </div>
-      <footer className="mt-5 font-weight-light text-center mb-5 text-secondary">
+      <footer className="mt-5 font-weight-light text-center pb-5 text-secondary">
```
songz: did you mean to move these into the `footer` component?

hwong0305: What's the purpose of changing margin to padding?

songz: colors I'd imagine. If you do a padding the footer takes the footer color, so I think the best practice is to use padding.

Wolfy64: Yes it was because of the background

PR #23, opened 2020-02-10
(Fix) - Add Default Color for Button

hwong0305:


components/Button.tsx
```diff
@@ -1,25 +1,47 @@
...
 type Props = {
-  color: string
+  color?: string
   text: string
   initial?: string | undefined
 }
 
 const Button = ({ color, initial, text }: Props) => {
+  let btnClass
+  switch (color) {
```
songz: To make this extensible, I think `type` is a better prop name. Then you can generate class like this:

```
const btnClass = `btn-${type}`
```

PR #31, opened 2020-02-13
Finished NavBar for Curriculum

hwong0305:
- decided to change layout to use navbar in courses and lessons.


pages/signup.tsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react'
 import Card from '../components/Card'
-import Layout from '../components/Layout'
+import LandingNav from '../components/LandingNav'
 
 const Signup: React.FC = () => {
   return (
-    <Layout>
+    <div>
+      <LandingNav />
```
Wolfy64: Why did you removed `<Layout />`
Layout should wrap all pages (except the landing page)

components/Layout.tsx
```diff
@@ -1,16 +1,31 @@
...
   return (
     <>
-      <LandingNav />
-      <div className="container">{props.children}</div>
+      <AppNav
+        loggedIn={loggedIn}
+        username={username}
+        firstName={firstName}
+        lastName={lastName}
+      />
```
Wolfy64: We might `useContext()` to pass `Auth data` to the <NavBar /> but maybe it's over-engineering

pages/signup.tsx
hwong0305: No. The Login / Signup page share the `Nav` with the landing page,. I changed `Layout` to use the Nav for courses. My implementation makes Layout uses the `secondNav` which isn't compatible with signup and login.

components/AppNavBar.tsx
```diff
@@ -1,29 +1,100 @@
...
+  firstName,
+  lastName
+}) => {
+  const initial =
+    firstName && lastName
+      ? firstName[0].toUpperCase() + lastName[0].toUpperCase()
+      : ''
+  if (!loggedIn) {
+    return (
+      <div className="bg-white">
```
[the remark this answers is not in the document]
songz: At PayPal we did this and the components start having all sorts of logic inside and it became difficult to read. I'd rather have:

```
if (isLoggedIn) {
  return <NavBarAth>
}
return <NavBarWithoutAuth />
```

Which is kind of what this PR is already doing.

hwong0305:
Need to update PR to add comment for why there are duplicate navbars.

yarn.lock
```diff
@@ -1710,28 +1710,7 @@
     react-inspector "^4.0.0"
     uuid "^3.3.2"
 
-<<<<<<< HEAD
```
songz: lol.

PR #32, opened 2020-02-14
Signup Validation with Formik

Wolfy64:
I will add Jest tests this weekend to improve the code coverage


helpers/formValidation.tsx
```diff
@@ -0,0 +1,29 @@
...
+const PASSWORD_MIN = 6
+
+const signupValidation = Yup.object({
+  email: Yup.string()
+    .email('Invalid email address')
+    .required('Required'),
+  userName: Yup.string()
+    .min(TEXT_MIN, `Must be at least ${TEXT_MIN} characters`)
+    .max(TEXT_MAX, `Must be ${TEXT_MAX} characters or less`)
+    .required('Required'),
```
hwong0305: To be consistent with the current naming scheme, I think we should enforce the first letter of the username be a lowercase letter.

There should be a check that only alphanumeric characters are used.

helpers/formValidation.tsx
```diff
@@ -0,0 +1,29 @@
...
+  password: Yup.string()
+    .min(PASSWORD_MIN, `Must be at least ${PASSWORD_MIN} characters`)
+    .max(TEXT_MAX, `Must be ${TEXT_MAX} characters or less`)
+    .required('Required'),
```
hwong0305: Should we allow users to have symbols in the password? Currently, we don't.

helpers/formValidation.tsx
Wolfy64: I will update it with `string.lowercase()`

helpers/formValidation.tsx
Wolfy64: I should add `string.matches(regex: Regex, message?: string | function): Schema)`
Do you any recommendation ?

helpers/formValidation.tsx
songz: one thing that really got me in c0d3.com codebase is.... we must decide early on.... `userName` as key or `username`. our old codebase has the cases different everywhere, became really painful.

Wolfy64: I would say we should use the camelCase convention ‘user name’ it’s 2 words then ‘userName’ make more sens in my opinion no ?

Wolfy64: I would say we should use the camelCase convention ‘user name’ it’s 2 words then ‘userName’ make more sens in my opinion no ?

Wolfy64: @hwong0305 PR updated feel free to merge it  if you are ok
About `username` I prefer to set everything in lower case instead of just the first char. _(we can always change it later if wanted)_

songz: @Wolfy64 officially its `username`, not `userName`. `userName` implies `user's name`, which is firstname + lastname usually.

proof: https://english.stackexchange.com/questions/43436/username-user-name-or-user-name

helpers/formValidation.tsx
```diff
@@ -0,0 +1,35 @@
...
+const TEXT_MIN = 2
+const TEXT_MAX = 20
+const PASSWORD_MIN = 6
+const REGEX_ALPHANUMERICS = /^[a-zA-Z0-9_]*$/
+
+const signupValidation = Yup.object({
+  email: Yup.string()
+    .email('Invalid email address')
+    .required('Required'),
+  userName: Yup.string()
```
songz: change to `username`  😈

helpers/formValidation.tsx
```diff
@@ -0,0 +1,35 @@
...
+    .matches(REGEX_ALPHANUMERICS, 'Must be alphanumerics characters')
+    .min(TEXT_MIN, `Must be at least ${TEXT_MIN} characters`)
+    .max(TEXT_MAX, `Must be ${TEXT_MAX} characters or less`)
+    .required('Required'),
+  password: Yup.string()
+    .min(PASSWORD_MIN, `Must be at least ${PASSWORD_MIN} characters`)
+    .max(TEXT_MAX, `Must be ${TEXT_MAX} characters or less`)
+    .required('Required'),
+  firstName: Yup.string()
+    .matches(REGEX_ALPHANUMERICS, 'Must be alphanumerics characters')
```
songz: my wife's first name has a space.

helpers/formValidation.tsx
Wolfy64: I would not have believed it thank for the link

helpers/formValidation.tsx
Wolfy64: @song Hum because we can't ask her name to change her name, I updated the regex for the sake of your wife. 😁
I could not find a way to accept only one space between words. Is anyone would have an idea?

PR #46, opened 2020-02-23
 add form validation to login page

rkalra247:


pages/login.tsx
```diff
@@ -1,35 +1,56 @@
 import * as React from 'react'
+import { Formik, Form, Field } from 'formik'
+import Input from '../components/Input'
+import { signupValidation } from '../helpers/formValidation'
```
Wolfy64: Don't you think that would be better to use its own validation like `loginValidation` or else to rename it by `formValidation` to be more generic

PR #55, opened 2020-02-26
Refactored by creating LessonCard component

aortizoj15:


pages/curriculum.tsx
```diff
@@ -26,40 +26,26 @@ const Curriculum: React.FC = () => {
...
       <Layout>
         <>
-          {sortedLessons}
-          <Card title="Progress"></Card>
-          <Card title="General Announcements"></Card>
+          <div className="row">
+            <div className="col-12 col-sm-12 col-md-8 col-lg-8 col-xl-8">
+              {sortedLessons}
+            </div>
+            <div className="col-12 col-sm-12 col-md-4 col-lg-4 col-xl-4"></div>
```
songz: what is dis for?

aortizoj15: Represents the two columns in our curriculum page, one for holding the Lesson Cards. The other to hold Progress Card, General Announcements card, and Links.

components/LessonCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+  lessonCount: number
+  hourCount: number
+  description: string
+}
+
+const LessonCard: React.FC<Props> = props => {
+  return (
+    <>
+      <div className="card shadow-sm mt-5">
+        <div className="card-body p-1">
```
Wolfy64: Could you removed the class `p-1` -> `<div className="card-body">`
I removed the card-body padding `padding: 1.25rem;`
Screen Shot 2020-02-27 at 9 20 17 AM

components/LessonCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <div className="row no-gutters">
+            <div className="cover-img col-6 col-sm-6 col-md-3 col-lg-3 col-xl-3 pl-1 pr-4 py-2 w-25 mx-auto">
```
songz: remove `col-lg-3` and `col-xl-3` because `col-md-3` takes care of >= 768px. Source: https://getbootstrap.com/docs/4.4/layout/grid/

pages/curriculum.tsx
```diff
@@ -26,40 +26,26 @@ const Curriculum: React.FC = () => {
...
-          <p>{e.description}</p>
-        </Card>
+      .map((e, idx) => (
+        <LessonCard
+          key={e.id}
+          coverImg={`js-${idx}-cover.svg`}
+          title={e.title}
+          lessonCount={10}
+          challengeCount={e.challenges.length}
+          hourCount={2}
```
songz: for now, let's just do `-` for `hourCount`. I don't want to mislead users.

components/LessonCard.tsx
aortizoj15: Removing p-1 adds more padding than needed since cards are almost to edge

scss/lessonCard.scss
```diff
@@ -0,0 +1,43 @@
+* {
+  color: #292929;
```
songz: `*` selectors shouldn't be applied here, it should be in a file that feels more global.

scss/lessonCard.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+.card {
+  max-width: 308px;
```
songz: doesn't this change ALL card classes? including the default bootstrap ones? Is that really what we want to do? I don't think this should be here.

scss/lessonCard.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+// .cover-img {
+//   margin: auto;
```
songz: remove comments

pages/curriculum.tsx
```diff
@@ -26,40 +26,26 @@ const Curriculum: React.FC = () => {
...
       ))
-
     return (
       <Layout>
         <>
-          {sortedLessons}
-          <Card title="Progress"></Card>
-          <Card title="General Announcements"></Card>
+          <div className="row">
+            <div className="cards-container col-12 col-sm-9">
```
songz: why not keep it easy and do `col-9` here and `col-3` for the other div? Is there a benefit of making this so complicated?

components/LessonCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+  title: string
+  challengeCount: number
+  lessonCount: number
+  hourCount: string | number
+  description: string
+}
+
+const LessonCard: React.FC<Props> = props => {
+  return (
+    <>
```
songz: you don't need this, you only have 1 children.

components/LessonCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+      <div className="card shadow-sm mt-5">
```
songz: `mt-5` feels like too much. `mt-3` is better, closer to the designs.

components/LessonCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <div className="card-body p-2">
+          <div className="row no-gutters">
+            <div className="cover-img col-auto mr-4">
```
songz: `col-2` and `mr-4`

components/LessonCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+              <img
+                className="card-img"
+                src={`/curriculumAssets/lessonCoversSvg/${props.coverImg}`}
+                alt={props.coverImg}
+              />
+            </div>
+            <div className="lesson-details col-12 col-md-8 col-lg-9">
```
songz: just `col-9` is enough here

scss/lessonCard.scss
hwong0305: We don't want to use `*` selector. Since, `scss` is not scoped, it's changing in colors in places we don't want. 

For example, the text inside the dark purple section in the Landing Page is `#292929` instead of `white`

scss/lessonCard.scss
```diff
@@ -0,0 +1,21 @@
+.card-title {
```
songz: according to [BEM](https://en.bem.info/methodology/quick-start/), this should be `.lesson-card__title`

pages/curriculum.tsx
```diff
@@ -26,40 +26,24 @@ const Curriculum: React.FC = () => {
...
+          lessonCount={10}
+          challengeCount={e.challenges.length}
+          hourCount="-"
+          description={e.description}
+        ></LessonCard>
       ))
-
     return (
       <Layout>
         <>
```
songz: since you have a children `row` div, you no longer need this fragment. `<>`

scss/lessonCard.scss
```diff
@@ -0,0 +1,21 @@
...
+  flex-wrap: wrap;
+  max-width: 75%;
+  justify-content: space-between;
+  white-space: nowrap;
+}
+.icon-text {
+  font-size: 0.8125rem;
+  line-height: 1.38;
+}
+.lesson-description {
```
songz: according to [BEM](https://en.bem.info/methodology/quick-start/), this should be `.lesson-card__description`

scss/lessonCard.scss
```diff
@@ -0,0 +1,21 @@
...
+  font-size: 0.9375rem;
+  line-height: 1.6;
+  margin-top: 0.7rem;
```
songz: could you give it a `mt-x` instead of using `margin-top`?

components/LessonCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+const LessonCard: React.FC<Props> = props => {
+  return (
+    <div className="card shadow-sm mt-3">
+      <div className="card-body p-2">
+        <div className="row no-gutters">
+          <div className="col-2 mr-4">
+            <img
+              className="card-img h-100"
```
songz: do you need `card-img` and `h-100` class?

components/LessonCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+              src={`/curriculumAssets/lessonCoversSvg/${props.coverImg}`}
+              alt={props.coverImg}
+            />
+          </div>
+          <div className="lesson-details col-9">
```
songz: `lesson-details` isn't used anywhere, I would remove it.

scss/lessonCard.scss
```diff
@@ -0,0 +1,21 @@
+.card-title {
+  font-size: 1.5rem;
+  font-weight: bold;
+  line-height: 1.33;
+}
+.icon-messages {
```
songz: according to BEM, this should be `.lesson-card__icon-container`

songz: do you need all the css properties? What if you just gave each a `mr-5` class name instead?

scss/lessonCard.scss
```diff
@@ -0,0 +1,21 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  max-width: 75%;
+  justify-content: space-between;
+  white-space: nowrap;
+}
+.icon-text {
```
songz: according to BEM, this should be `.lesson-card__icon-text`

scss/lessonCard.scss
aortizoj15: Corrected 👍

pages/curriculum.tsx
aortizoj15: Corrected 👍

scss/lessonCard.scss
aortizoj15: Corrected 👍

scss/lessonCard.scss
aortizoj15: Changed to mt-2 👍

components/LessonCard.tsx
aortizoj15: Removed and added mw-100 to container instead. This was put in to avoid columns overlapping in small screen sizes. Now columns stack

scss/lessonCard.scss
aortizoj15: Fixed using comment in chat 👍

scss/lessonCard.scss
aortizoj15: Fixed 👍

stories/components/LessonCard.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const Basic: React.FC = () => (
+  <LessonCard
+    coverImg="Name of Image File"
+    title="Title of Lesson"
+    challengeCount={numberOfChallenges}
+    lessonCount={numberOfLessons}
+    hourCount={numberOfHours}
+    description="Lesson Description"
+  ></LessonCard>
```
songz: minor, could you hardcode one of the lessons so the story looks better? https://deploy-preview-55--youthful-thompson-448332.netlify.com/?path=/story/components-lessoncard--basic

aortizoj15: Fixed 👍

PR #59, opened 2020-03-04
Lesson Card In Progess State

aortizoj15:


components/LessonCard.tsx
```diff
@@ -9,11 +9,16 @@ type Props = {
   lessonCount: number
   hourCount: string | number
   description: string
+  complete?: boolean
+  inProgress?: boolean
```
songz: lessons cannot be `complete` and `inProgress` at the same time. Doing this could be confusing later on. I think `currentState?: string` is a good one.

This current combination gives us 4 states: `true/true`, `false/false`, `true/false`, `false/true`

We really need 3 states, so having a `string` to represent state is better imo.
`inProgress`, `completed`, or default.

scss/lessonCard.scss
```diff
@@ -1,3 +1,6 @@
+.lesson-card__container {
```
songz: this is a state, for in progress. 

```
.lesson-card__container_inprogress
```

Look at `Modifier` -> boolean section.
https://en.bem.info/methodology/quick-start/#modifier

components/LessonCard.tsx
songz: This will make your `containerClass` logic easier to manager down the road.

PR #61, opened 2020-03-06
First Draft of LessonTitleCard for Challenges Page

aortizoj15:


components/LessonTitleCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <div className="card shadow-sm mt-3 col-12 px-0 pt-3">
+      <div className="card-body p-0">
+        <div className="d-flex mb-3 px-3">
+          <img
+            className="lessonTitleCard__lesson-cover mr-3"
+            src={`/curriculumAssets/lessonCoversSvg/${props.lessonCoverUrl}`}
+            alt="lesson-cover"
+          />
+          <div>
+            <p className="m-0 font-weight-bold">
```
songz: can we remove `font-weight-bold`? Looks like ink bleed

components/LessonTitleCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+              <a href={`${props.lessonUrl}`}>Go Back</a>
+            </p>
+            <h1 className="lessonTitleCard__lesson-title font-weight-bold">
```
songz: can we remove `font-weight-bold` from here? Looks like ink bleed

components/LessonTitleCard.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import '../scss/lessonTitleCard.scss'
```
Wolfy64: I did not experienced it myself at work but could you take a look at @use instead of import that will be deprecated soon?
[DOC](https://sass-lang.com/documentation/at-rules/import)
>The Sass team discourages the continued use of the @import rule. Sass will gradually phase it out over the next few years, and eventually remove it from the language entirely. Prefer the @use rule instead.

components/LessonTitleCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+            <h1 className="lessonTitleCard__lesson-title font-weight-bold">
+              {props.lessonTitle}
+            </h1>
+          </div>
+        </div>
+        <div className="card-footer bg-white p-0">
+          <a
+            className="btn border-right rounded-0 px-4 py-3"
+            href={`${props.lessonUrl}`}
+          >
```
Wolfy64: For intern link it's [recommended](https://nextjs.org/docs/api-reference/next/link) to use `<Link />`

songz: this is not internal link, it is notion link.

pages/challenges.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react'
+import Layout from '../components/Layout'
+import LessonTitleCard from '../components/LessonTitleCard'
+
+const setupAlert = (
```
songz: `setupAlert` is not a good name for 2 reasons:

1. `verb`. Components should be nouns, putting a verb here is misleading
2. Components should be Capitalized.

A better name: `Alert`

pages/challenges.tsx
```diff
@@ -0,0 +1,32 @@
...
+  <div className="alert bg-primary text-white col-12 mt-4" role="alert">
```
songz: why not use the alert provided by bootstrap? https://getbootstrap.com/docs/4.4/components/alerts/

hwong0305:
Change from from `pages/challenges.tsx` to `pages/curriculum/[lesson].tsx`

(Source: https://nextjs.org/docs/routing/dynamic-routes)

PR #65, opened 2020-03-09
Fixing Card Borders

aortizoj15:


components/LessonCard.tsx
```diff
@@ -16,8 +16,8 @@ type Props = {
 const LessonCard: React.FC<Props> = props => {
   const containerClass =
     props.currentState === 'inProgress'
-      ? 'lesson-card__container_inprogress card shadow-sm mt-3 border-primary'
-      : 'card shadow-sm mt-3'
+      ? 'lesson-card__container_inprogress card shadow-sm mt-3 border-primary border-0'
+      : 'card shadow-sm mt-3 border-0'
   return (
     <div className={containerClass}>
```
songz:
```
className={ `border-0 ${containerClass}` }
```

components/LessonCard.tsx
```diff
@@ -16,8 +16,8 @@ type Props = {
 const LessonCard: React.FC<Props> = props => {
   const containerClass =
     props.currentState === 'inProgress'
-      ? 'lesson-card__container_inprogress card shadow-sm mt-3 border-primary'
-      : 'card shadow-sm mt-3'
+      ? 'lesson-card__container_inprogress card shadow-sm mt-3 border-primary border-0'
+      : 'card shadow-sm mt-3 border-0'
```
songz: since you are adding `border-0` to both, just add it once to the place you are using it at.

PR #66, opened 2020-03-10
Added login handler, form validation

aortizoj15:


pages/login.tsx
```diff
@@ -12,15 +13,38 @@ const initialValues = {
...
 }
 
-const Login: React.FC<Props> = props => {
+type Values = {
+  username: string
+  password: string
+}
+
+async function handleSubmit(values: Values) {
+  const res = await fetch('https://c0d3.com/signin', {
```
songz: this needs to be a global environmental variable.

next.config.js
```diff
@@ -1,2 +1,6 @@
 const withSass = require('@zeit/next-sass')
-module.exports = withSass({})
+module.exports = withSass({
+  env: {
+    SERVER_URL: 'https://c0d3.com/signin'
```
hwong0305: It's not good practice to expose your `SERVER_URL` in the config file. It should be loaded from an environmental variable like `process.env.SERVER_URL`

.env.build
```diff
@@ -0,0 +1,2 @@
+SERVER_URL="https://c0d3.com/"
+URI="https://song.c0d3.com/"
```
songz: we don't need 2. just 1 is enough.

```
SERVER_URL="https://www.c0d3.com"
```

helpers/withApollo.ts
```diff
@@ -1,14 +1,15 @@
 import withApollo from 'next-with-apollo'
 import ApolloClient, { InMemoryCache } from 'apollo-boost'
 
-/* This URL is only used for development. 
+/* This URL is only used for development.
```
hwong0305: The comment is no longer needed.

.env.build
```diff
@@ -0,0 +1 @@
+SERVER_URL="https://c0d3.com"
```
hwong0305: This should be removed from the repo since it's already been gitignored. 

Try running `git rm .env.build --cached`. That should remove it from the git repository and keep it on your computer.

Issue #33, opened 2020-02-16
Form Validation for Login


rkalra247:

songz:
Fixed - https://github.com/garageScript/c0d3-app/pull/67

PR #67, opened 2020-03-14
Login Validation

aortizoj15:


pages/login.tsx
```diff
@@ -20,26 +21,27 @@ type Values = {
...
-  const res = await fetch(`${SERVER_URL}/signin`, {
-    method: 'POST',
-    headers: {
-      'content-type': 'application/json'
-    },
-    body: JSON.stringify({
-      username: values.username,
-      password: values.password
+const Login: React.FC<Props> = () => {
+  const [invalidCredentials, setInvalidCredentials] = useState(false)
```
Wolfy64: Your state logic is to display or not `<Alert />` I would rather name the state:
 `[isAlertVisible setIsAlertVisible] = useState(false)`

pages/login.tsx
```diff
@@ -20,26 +21,27 @@ type Values = {
...
+      body: JSON.stringify({
+        username: values.username,
+        password: values.password
+      })
     })
-  })
-  const data = await res.json()
-  if (data.success) {
+    if (res.redirected) {
+      setInvalidCredentials(true)
```
Wolfy64: if I read `setInvalidCredentials(true) ` I'm expecting you are doing some kind of logic but not to display a message.

helpers/loginUser.ts
```diff
@@ -0,0 +1,18 @@
+const SERVER_URL = process.env.SERVER_URL
+
+export const loginUser = async (username: string, password: string) => {
```
songz: tests for this?

__tests__/loginUserTest.js
```diff
@@ -0,0 +1,19 @@
+import { loginUser } from '../helpers/loginUser'
+describe('loginUser helper', () => {
+  test('Should return object', async () => {
+    window.fetch = jest.fn().mockReturnValue(
+      Promise.resolve({
+        redirected: false
+      })
+    )
+    expect(Promise.resolve(loginUser('user123', 'pass123'))).resolves.toBe({})
```
songz: you can also do this:

```
const user = await loginUser(....)
expect(user).toEqual({})
```

songz:
Closes #33 - Login form validation

PR #73, opened 2020-03-24
Added documentation for ApolloClient Cache

hwong0305:


helpers/withApollo.ts
```diff
@@ -13,6 +13,7 @@ const link = createHttpLink({
 export default withApollo(
   ({ initialState }) =>
     new ApolloClient({
+      // Cache is used to rehydrating the store for Server Side Rendering.
```
songz: could you add a link to the source you got this from

PR #79, opened 2020-03-27
Step 1: Setup StoryBoard for SubmissionCards

SahilKalra98:


scss/submissionCard.scss
```diff
@@ -0,0 +1,3 @@
+.submissionTime {
+  font-size: 12px;
```
songz: looked at storyboard, looks better without this css.

PR #98, opened 2020-04-07
Pending submissions

aortizoj15:


__tests__/components/LessonCard.js
```diff
@@ -0,0 +1,89 @@
+jest.mock('@apollo/react-hooks')
+//jest.mock('../../components/LessonCard')
```
songz, line 2: remove unnecessary comments.

aortizoj15: Always catching my comments, thanks!

Issue #107, opened 2020-04-08
[Lesson Page] - Challenge Card State


aortizoj15:
AC
* On the left pane, when user has completed the challenge, it should indicate so.
* Other status (untouched, pending review, rejected)
* https://app.zeplin.io/project/5e19606fd539a701e62f9d70/screen/5e44ed57cf17278c82cb74ef

PR #113, opened 2020-04-10
WIP #107 styling for ChallengeTitleCard to show what state it is in

coltonehrman:
Added the image SVGs from zeplin, and styling to make ChallengeTitleCard appear with 3 different states -- Basic, Complete, Pending


coltonehrman:
Added styling for Current state

PR #114, opened 2020-04-11
WIP #107 styling for ChallengeTitleCard to show what state it is in

coltonehrman:
Added styling for different states a ChallengeTitleCard can be in. Basic, Complete, Pending, Current.


components/ChallengeMaterial.tsx
```diff
@@ -17,10 +18,29 @@ type ChallengeMaterialProps = {
 }
 
 export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = props => {
+  const state = props.currentState
+  let cardStyles = 'shadow-sm border-0'
+  if (state && state === 'complete') cardStyles = 'challenge-title-card--done'
```
songz, line 23: `state && ` is unnecessary because you literally set it in line 21.

coltonehrman: Fixed :)

scss/challengeTitleCard.scss
```diff
@@ -0,0 +1,8 @@
+.challenge-title-card--done {
+    color: #6d7278;
+    background-color: rgba(255, 255, 255, 0.5);
+}
+
+.challenge-title-card--current {
+    border-left: 3px solid #5438dc;
```
songz, line 7: for colors, could we use variables?

__tests__/components/__snapshots__/LessonCard.js.snap
```diff
@@ -51,7 +51,11 @@ exports[`Lesson Card Complete State Should render lessonCard with loading... 1`]
         >
           <h4
             class="lesson-card__title font-weight-bold mt-3"
-          />
+          >
+            <a
+              href="/curriculum/undefined"
```
SahilKalra98, line 56: This is not supposed to be undefined right? /curriculum/undefined

coltonehrman: Correct, I believe this is one of the test lessons that got added and has no lessonId. But, maybe it would be a good idea to check for undefined id and do something else?

scss/challengeTitleCard.scss
coltonehrman: Ah yes, been ages since I used scss. Just saw the index.scss file with variables :)

songz: https://github.com/garageScript/c0d3-app/blob/master/scss/index.scss

components/LessonCard.tsx
```diff
@@ -104,7 +104,7 @@ const LessonCard: React.FC<Props> = props => {
           </div>
           <div className="col-9">
             <h4 className="lesson-card__title font-weight-bold mt-3">
-              {props.title}
+              <a href={`/curriculum/${props.lessonId}`}>{props.title}</a>
```
hwong0305, line 107: If this is internal. It should be a `<Link>`

coltonehrman: Fixed :0

scss/challengeTitleCard.scss
coltonehrman: Fixed :)

components/ChallengeMaterial.tsx
```diff
@@ -17,10 +18,29 @@ type ChallengeMaterialProps = {
 }
 
 export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = props => {
+  const { currentState: state, active } = props
+  let cardStyles = []
```
songz, line 22: `const`

components/ChallengeMaterial.tsx
```diff
@@ -17,10 +18,29 @@ type ChallengeMaterialProps = {
...
+  if (state === 'complete') cardStyles.push('challenge-title-card--done')
```
songz, line 23: could you use `{}` please? Its easier to read / less error prone in the future. Especially since there are quite a few piece of logic.

```
if (...) {
  ...
} else {
  ...
}
if (...) {
  ...
}
```

coltonehrman: Fixed!

components/ChallengeMaterial.tsx
coltonehrman: Fixed!

PR #118, opened 2020-04-13
Clickable challenge Title with rebase

michaelbayday:


components/ChallengeMaterial.tsx
```diff
@@ -19,17 +23,21 @@ type ChallengeMaterialProps = {
...
+  const cardStyles = ['challenge-title-card']
   if (state === 'complete') {
-    cardStyles.push('challenge-title-card--done')
+    cardStyles[0] += '--done'
   } else {
     cardStyles.push('shadow-sm', 'border-0')
   }
   if (active) {
-    cardStyles.push('challenge-title-card--active')
+    cardStyles[0] += '--active'
```
coltonehrman, line 33: I just realized that this logic would incorrectly render a Challenge Title Card that is both Complete & Current. Maybe we don't want this to be possible, but personally I would like the idea of being able to click on my previously completed Challenges just to look at them again, thus making them look Complete & Current.

components/ChallengeMaterial.tsx
```diff
@@ -19,17 +23,21 @@ type ChallengeMaterialProps = {
...
   }
   return (
-    <div className={`card mb-2 ${cardStyles.join(' ')}`}>
+    <div
+      data-testid="challenge-title"
```
coltonehrman, line 37: What is data-testid?

components/ChallengeMaterial.tsx
```diff
@@ -1,16 +1,20 @@
...
   challengeNum: number
   currentState?: string
   active?: boolean
+  setCurrentChallenge: React.Dispatch<CurrentChallengeID>
 }
 
 type ChallengeQuestionCardProps = {
-  title: string
-  question: string
+  currentChallenge: Challenge | undefined
```
coltonehrman, line 17: Why `Challenge | undefined`?

components/ChallengeMaterial.tsx
```diff
@@ -51,18 +59,26 @@ export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = props => {
...
-        <p className="challenge-question-card__question bg-light p-3 mt-3">
-          {props.question}
-        </p>
+export const ChallengeQuestionCard: React.FC<ChallengeQuestionCardProps> = ({
+  currentChallenge
+}) => {
+  return (
+    <div className="card shadow-sm border-0">
+      <div className="card-body challenge-question">
+        <h1 data-testid="challenge-question-title" className="card-title">
```
coltonehrman, line 68: data-testid?

components/ChallengeMaterial.tsx
```diff
@@ -51,18 +59,26 @@ export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = props => {
...
+          {currentChallenge!.title}
+        </h1>
+        <div>
+          <p
+            data-testid="challenge-question-description"
+            className="card-question bg-light p-3 mt-3"
+          >
+            {currentChallenge!.description}
```
coltonehrman, line 76: Not sure if I understand the usage of `currentChallenge!` is this the same as forcefully unwrapping an optional when you know there is a value there?

components/ChallengeMaterial.tsx
Wolfy64: It's for tests --> [React Testing Library](https://testing-library.com/docs/dom-testing-library/api-queries#bytestid)

components/ChallengeMaterial.tsx
songz: to make this selectable during test: https://github.com/garageScript/c0d3-app/pull/118/files?file-filters%5B%5D=.js&file-filters%5B%5D=.scss&file-filters%5B%5D=.tsx#diff-7e495cc52c6d8a8ca0f0836a2c6a1f5dR42

components/ChallengeMaterial.tsx
michaelbayday: ill update this to assume that currentChallenge will be the first challenge unless there is a more recent active one.

components/ChallengeMaterial.tsx
michaelbayday: updated the types so we good!

components/ChallengeMaterial.tsx
songz: yeah they seem like 2 different independent states, not correlated.

components/ChallengeMaterial.tsx
```diff
@@ -72,25 +88,35 @@ const ChallengeMaterial: React.FC<ChallengeMaterialProps> = props => {
...
+  //hasn't been completed to set as default challengeID
+  const [currentChallengeID, setCurrentChallenge] = useState<
+    CurrentChallengeID
+  >()
+  //assign currentChallenge to first challenge if no match found
+  const currentChallenge =
+    sortedChallenges.find(
+      (challenge: Challenge) => challenge.id === currentChallengeID
+    ) || sortedChallenges[0]
+  console.log(sortedChallenges)
```
songz, line 101: remove `console.log`

PR #122, opened 2020-04-14
[NO MERGE - PENDING CLOSE] Auth Context Provider

coltonehrman:
Added AuthContext to keep track of the state of the User being logged in or not
Updated internal links to use next/link
Made a home page that way even when logged in you can visit the LandingPage if you want


pages/curriculum/[lesson].tsx
```diff
@@ -22,14 +22,17 @@ const Challenges: React.FC = () => {
   const currentLesson = sortedLessons.find(
     lesson => router.query.lesson === lesson.id.toString()
   )
+  console.log(currentLesson)
```
Comment (michaelbayday) [inline pages/curriculum/[lesson].tsx:25]: console log  :)

Comment (coltonehrman) [reply pages/curriculum/[lesson].tsx:25]: lol good catch

components/AppNav.tsx
```diff
@@ -84,31 +79,64 @@ const UnAuthLink = () => (
...
+                <a className="navbar-brand text-primary font-weight-bold">
+                  C0D3
+                </a>
+              </Link>
+              <div id="navbarNav">
+                <div className="navbar-nav collapse navbar-collapse">
+                  {user ? (
+                    <AuthLinks active={router && router.route} />
+                  ) : (
+                    <UnAuthLinks active={router && router.route} />
```
michaelbayday, line 126: when would router && router.route not be true?

michaelbayday: nvm, it seems like you need to mock router.route to get coverage for this

coltonehrman:
I had to shift around a few things from my original code but was able to get most of it in with 100% code coverage and storybook still working.
Some things to point out with this PR.

1. UserContext is not available everywhere since it's being set only in IndexPage
2. Using ugly workaround inside AppNav to prevent storybook from crashing when no router is returned from `useRouter` -- need to figure out a way to mock this inside the storyboard standalone version (which doesn't have access to jest context as far as I know)
3. Storybook looks like it is working, but if you look at AppNav versions LoggedIn & LoggedOut are the same because no next router context is gone

michaelbayday:
I worked with colton a bit and we tried using jest mock functions for mocking next router in testing/storybook but ran into issues, but am interested to see what the rest of the team would propose to address the concerns raised above

pages/curriculum/[lesson].tsx
```diff
@@ -29,7 +29,9 @@ const Challenges: React.FC = () => {
           {currentLesson && (
             <div className="challenges-container">
               <LessonTitleCard
-                lessonCoverUrl={`js-${currentLesson.order}-cover.svg`}
+                lessonCoverUrl={`js-${sortedLessons.indexOf(
+                  currentLesson
```
Comment (songz) [inline pages/curriculum/[lesson].tsx:33]: original is more correct.

pages/home.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react'
+import AppNav from '../components/AppNav'
+import LandingPage from '../components/LandingPage'
+import Footer from '../components/Footer'
+
+const Home: React.FC = () => (
+  <>
+    <AppNav />
```
songz, line 8: move landing page nav out, and put here directly. Landing page Nav shouldn't show up anywhere else imo.

pages/index.tsx
```diff
@@ -1,30 +1,32 @@
...
+  let Component = Home
+
   // while loading, don't show anything to user
   if (!data && !error) {
-    return null
+    return <h1>Loading</h1>
   }
+
+  if (data && data.userInfo) {
+    Component = Curriculum
```
songz, line 23: `if (!data || !data.userInfo) return <Landing /> `

pages/index.tsx
```diff
@@ -1,30 +1,32 @@
...
+  }
+
   return (
-    <>
-      <AppNav loggedIn={false} />
-      <LandingPage />
-      <Footer footerType="py-5 bg-white text-muted" />
-    </>
+    <AuthUserContext.Provider value={data && data.userInfo}>
+      <Component />
```
songz, line 28: should just always render `curriculum`

pages/curriculum/[lesson].tsx
Comment (coltonehrman) [reply pages/curriculum/[lesson].tsx:33]: Fixed!

components/AppNav.tsx
```diff
@@ -1,69 +1,64 @@
...
-    <Link href="/signup">
-      <a className="btn btn-secondary border mr-3">Signup</a>
-    </Link>
-  </div>
-)
-
-const UnAuthLink = () => (
+export const UnAuthLinks: React.FC<LinkProps> = ({ active }) => (
   <div className="navbar-nav collapse navbar-collapse">
     <Link href="/">
```
songz, line 49: should create a new component instead of sprinkling logic: `<NavLink path="/" activePath={activePath} />`

components/AppNav.tsx
```diff
@@ -1,69 +1,64 @@
...
-      <a className="btn btn-secondary border mr-3">Login</a>
-    </Link>
-    <Link href="/signup">
-      <a className="btn btn-secondary border mr-3">Signup</a>
-    </Link>
-  </div>
-)
-
-const UnAuthLink = () => (
+export const UnAuthLinks: React.FC<LinkProps> = ({ active }) => (
```
songz, line 47: `activePath` is a better name. Otherwise it seems like it would be a bool.

components/AppNav.tsx
```diff
@@ -84,31 +79,64 @@ const UnAuthLink = () => (
...
-  username = '',
-  firstName,
-  lastName
-}) => {
-  const initial = firstName && lastName ? firstName[0] + lastName[0] : ''
+export const AuthButtons: React.FC<AuthButtonProps> = ({
+  initial,
+  username
+}) => (
+  <div>
```
songz, line 86: why does the button need to be wrapped around a `div` tag?

songz: Use react fragments when possible: `<>....</>`

pages/home.tsx
songz: nvm, saw the logic.

pages/index.tsx
coltonehrman: Fixed

pages/index.tsx
coltonehrman: Fixed

components/AppNav.tsx
coltonehrman: Fixed

songz:
Closing, all features are implemented.

PR #125, opened 2020-04-15
[SMALL PR] - useSession Hook

coltonehrman:
Added custom useSession hook to be used in components, still need to figure out where I can use this hook to create AuthContext for all components in-app. This is the first step...


michaelbayday:
Looks good! excited to start using it!!!

PR #128, opened 2020-04-15
[NEW COMPONENT] NavLink

coltonehrman:
added NavLink component + converted <a> tags to NavLink

Added NavLink component to make it easier and cleaner for <a> tags in components
Was going to make the component throw an error when passed `internal` && `blank` but couldn't figure out how to test it properly


__tests__/components/NavLink.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import NavLink from '../../components/NavLink'
+
+describe('NavLink Component', () => {
+  test('Should crash if given internal && blank props', () => {
+    // expect(() => render(<NavLink internal blank />)).toThrow()
```
songz, line 7: why do we need an empty test?

coltonehrman: In my comments I mentioned how I wasn't sure how to test this properly lol, I will either update it when I figure it out or remove it completely if we think it's not necessary

songz: storybook comment. Just render the different props is good enough.

coltonehrman: removed test

PR #137, opened 2020-04-16
[NEW COMPONENT] NavLink

coltonehrman:


__tests__/components/__snapshots__/LessonCard.js.snap
```diff
@@ -47,6 +47,7 @@ exports[`Lesson Card Complete State Should render lessonCard with loading... 1`]
             class="lesson-card__title font-weight-bold mt-3"
           >
             <a
+              class=" "
```
songz, line 50: why is there a `class=" "`? Shouldn't it have no class?

coltonehrman: Right, i was debated whether or not to fix this. This is because of the way I am rendering the NavLinks, would it be best to add more conditionals for when no className is passed in?

components/NavLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+  text,
+  path,
+  activePath,
+  as,
+  internal,
+  blank,
+  className
+}) => {
+  const active = path === activePath
+  className = className ? className : ''
```
songz, line 24: what happens if you passed in className directly?

components/NavLink.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import Link from 'next/link'
+
+type NavLinkProps = {
+  text: string
+  path: string
+  activePath?: string
+  as?: string
+  internal?: true
```
songz, line 9: better to use `external` as a default.

components/NavLink.tsx
coltonehrman: I was getting className="undefined" without this line

components/NavLink.tsx
songz: we want to default to internal link, then if someone wants to link to external they should explicitly say so.

coltonehrman: Yes, this makes more sense

components/NavLink.tsx
songz: then `className = className || ''` is clearer than a tertiary

components/AdditionalResources.tsx
```diff
@@ -8,36 +9,31 @@ const AdditionalResources: React.FC = () => {
...
+        path="https://stackoverflow.com/"
         className="mx-3"
-        href="https://stackoverflow.com/"
-        target="_blank"
-        rel="noopener noreferrer"
-      >
-        Stack Overflow
-      </a>
+        external
+        blank
```
songz, line 27: is it safe to say that its always gonna be `blank` when link is external? I can't think of a scenario where we want users to leave our page....

components/NavLink.tsx
```diff
@@ -0,0 +1,54 @@
...
+  path: string
+  activePath?: string
+  external?: true
+  as?: string
+  blank?: true
+  className?: string
+}
+
+const NavLink: React.FC<NavLinkProps> = ({
+  text,
```
Wolfy64, line 15: I don't see the point of the `text` props . You could use `children` .
With `children` we have more freedom you can do:
```js
<NavLink>
  {children} // could be a text, img or a node
</NavLink>

<NavLink children="You text here if you want a self-closing tag" />
```

pages/admin.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+import { useQuery } from '@apollo/react-hooks'
+import { GET_LESSONS } from '../graphql/queries'
+
+const SERVER_URL = process.env.SERVER_URL
+
+export const fetcher = (url: string) =>
+  fetch(url, { credentials: 'include' }).then(r => r.json())
+
+const AdminPage: any = () => {
```
Wolfy64, line 14: What is the link with this code and your PR [NEW COMPONENT] NavLink?

components/NavLink.tsx
songz: didn't know this.... haha thanks @Wolfy64

coltonehrman: Makes sense :)

pages/admin.tsx
coltonehrman: oh shoot. lol i didn't realize this page code got pushed in as well. thanks for pointing that out

components/NavLink.tsx
coltonehrman: fixed!

components/AdditionalResources.tsx
coltonehrman: fixed!

components/NavLink.tsx
coltonehrman: fixed!

PR #140, opened 2020-04-18
[NEW COMPONENT] NavLink

coltonehrman:


components/NavLink.tsx
```diff
@@ -0,0 +1,61 @@
...
+  className?: string
+}
+
+const NavLink: React.FC<NavLinkProps> = ({
+  children,
+  path,
+  activePath,
+  as,
+  external,
+  className
```
songz, line 18: `className=''`

components/NavLink.tsx
```diff
@@ -0,0 +1,61 @@
...
+}) => {
+  const active = path === activePath
+  const hasClass = typeof className !== 'undefined'
```
songz, line 21:
```
if (path === activePath) {
  className += ' active'
}
```

coltonehrman: I need the active variable for another line as well.

coltonehrman:
```
 <a className={`${className} ${active ? 'active' : ''}`}>
   {active && <span className="sr-only">(current)</span>}
   {children}
</a>
```

Issue #121, opened 2020-04-14
[Lesson Page] - Challenges Completed 


SahilKalra98:
https://app.zeplin.io/project/5e19606fd539a701e62f9d70/screen/5e44ed56127ab694c3bd4349

PR #141, opened 2020-04-18
Fix bug where page would not render if user passed all challenges

michaelbayday:
TODO by @SahilKalra98 : #121


songz:
closes https://github.com/garageScript/c0d3-app/issues/142

PR #149, opened 2020-04-20
[TYPE UPDATE] - Session Context

coltonehrman:
We should not be returning null because data should be an object.
These changes makes it so our default session context is always an object instead of null.

Rather than make it an empty object, I just added success key

### DIFF
--- a/helpers/contexts/session.ts
+++ b/helpers/contexts/session.ts
@@ -2,8 +2,7 @@ import { createContext } from 'react'
 import { SessionData } from '../useSession'
 
 const SessionContext = createContext<SessionData>({
-  data: null,
-  error: null
+  data: { success: false }
 })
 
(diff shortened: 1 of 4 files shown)


helpers/contexts/session.ts
```diff
@@ -2,8 +2,7 @@ import { createContext } from 'react'
 import { SessionData } from '../useSession'
 
 const SessionContext = createContext<SessionData>({
-  data: null,
-  error: null
+  data: { success: false }
```
songz, line 5: this may make sense for you now, but I think in the future it may confuse other devs because you are never setting it to `success: true`

coltonehrman: yeah it would be ideal to have it match what is returned from the `/session` endpoint in the future :)

PR #156, opened 2020-04-21
[Lesson Page] Render Challenge Diff

### LINKED ISSUE #111: [Lesson Page] - Render submission code diff
AC
* Submission code diff should render the diff UI correctly

coltonehrman:
Use `react-diff-viewer` to render the submission diff on each challenge

Closes #111


components/ChallengeMaterial.tsx
```diff
@@ -200,12 +221,11 @@ const ChallengeMaterial: React.FC<ChallengeMaterialProps> = ({
     status: 'passed'
   }
   //find first challenge that is not passed on initial render after clicks will render clicked challenge
-  const currentChallenge: ChallengeSubmissionData =
+  let currentChallenge =
```
michaelbayday, line 224: Why let?

coltonehrman: good catch! this must've got changed during the rebase and  i didn't catch it :)

coltonehrman: fixed :)

components/ChallengeMaterial.tsx
```diff
@@ -96,20 +98,39 @@ export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = ({
...
+            data-testid="challenge-question-description"
+            className="bg-light p-3 mt-3"
+          >
+            <Markdown>{currentChallenge.description}</Markdown>
+          </div>
         </div>
       </div>
-    </div>
+
+      {currentChallenge.submission?.diff && (
```
songz, line 119: I know `?` is cool, but it is error prone, especially if used in multiple places, so it is a practice we discourage. In long nested objects, you must put `?` behind every `.`, which potentially leads to more errors.

Please create variables instead.
```
const diff = _.get(currentChallenge, 'submission.diff')
const username = _.get(data, 'userInfo.username')
```

coltonehrman: will do! thanks :)

Wolfy64: cool I didn't know about `_.get()`

coltonehrman:
> Why there is no Storybook with `<ReactDiffViewer />`?
> Did you forget to include it?

Do we have to add StoryBook for 3rd-party components?

Wolfy64:
It's not about 3rd-party components but to have an idea on how it's look. 🙂
You know your code and it's look & feel. As reviewer sometimes it's difficult to have an idea just with the code.
That why we have StoryBook with page folder or even a screenshot of your work would be fine for me.

PR #159, opened 2020-04-22
[GRAPHQL] Query Session

coltonehrman:
Added graphql resolver for session data.

PS - Removed unnecessary User request inside lessons query resolver (do not need to go through every user to look for lesson) *thanks to song!*

Fun Fact: Brought request time down from 12-13s to 25-50ms

### DIFF
--- a/graphql/resolvers.ts
+++ b/graphql/resolvers.ts
@@ -1,5 +1,6 @@
-import { login, logout, signup } from '../helpers/controllers/authController'
+import { Request } from 'express'
 import db from '../helpers/dbload'
+import { login, logout, signup } from '../helpers/controllers/authController'
 import _ from 'lodash'
 
 const { User, Submission, UserLesson, Lesson } = db
@@ -12,26 +13,29 @@ export default {
         order: [['order', 'ASC']]
       })
     },
-    async session(_parent: any, _args: any, context: any) {
+    async session(_parent: void, _args: void, context: { req: Request }) {
       const userId = _.get(context, 'req.session.userId', false)
 
       if (!userId) {
         return null
       }
 
-      const user = await User.findOne({
-        where: { id: userId },
-        include: [ Submission, UserLesson ]
-      })
+      // https://github.com/garageScript/c0d3-app/wiki/Sequelize-Query-Performance
+
+      const [user, submissions, lessonStatus] = await Promise.all([
+        User.findOne({ where: { id: userId } }),
+        Submission.findAll({ where: { userId } }),
+        UserLesson.findAll({ where: { userId } })
+      ])
 
       if (!user) {
         return null
       }
 
-      return { 
+      return {
         user,
-        submissions: user.Submissions,
-        lessonStatus: user.UserLessons
+        submissions,
+        lessonStatus
       }
     }
   },
--- a/graphql/resolvers.ts
+++ b/graphql/resolvers.ts
@@ -1,7 +1,8 @@
 import { login, logout, signup } from '../helpers/controllers/authController'
 import db from '../helpers/dbload'
+import _ from 'lodash'
 
-const { Lesson } = db
+const { User, Submission, UserLesson, Lesson } = db
 
 export default {
   Query: {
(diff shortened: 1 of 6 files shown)


pages/api/graphql.ts
```diff
@@ -16,7 +17,25 @@ const handler: any = nextConnect() // For session middleware. TODO: Need to defi
 const apolloServer = new ApolloServer({
   typeDefs,
   resolvers,
-  context: ({ req }) => ({ req }) // This lets GraphQL have access to sessions
+  context: async ({ req }) => {
```
hwong0305, line 20: This should not be done in context. We don't want this query to run everytime a request is made to GraphQL. In addition, having `Promise.all` iseems really inefficient when we could just include all `Submission` and `UserLesson` in one query.

pages/api/graphql.ts
```diff
@@ -31,7 +50,6 @@ handler
       resave: false,
       saveUninitialized: false,
       cookie: {
-        domain: process.env.DB_HOST,
```
hwong0305, line 34: Why is domain being removed?

songz: isn't the default `domain` the some host?

graphql/resolvers.ts
```diff
@@ -1,23 +1,24 @@
...
-        include: [
-          'challenges',
-          {
-            model: User,
-            through: {
-              attributes: ['isPassed', 'isTeaching', 'isEnrolled']
-            }
-          }
-        ],
+        include: ['challenges'],
```
hwong0305, line 10: When querying Lessons, we should get all the information about whether the User has enrolled / passed / teaching the lesson.  This way the information can be rendered in the UI.

Otherwise the UI has no way to tell whether the User is done with the lesson or not.

pages/api/graphql.ts
hwong0305: When I was testing session, it didn't have any `domain` in the cookie if I don't put that.

coltonehrman:
Did some testing on this and seem to be getting some interesting results with the two different pieces of code. Only looked at front-end network performance on this testing, but below are the results with the following code.

```
const user = await User.findOne({
  where: { id: userId },
  include: [Submission, UserLesson]
})

const submissions = user.Submissions
const lessonStatus = user.UserLessons
```

<a href="https://ibb.co/WzJ0gzM">Screen-Shot-2020-04-24-at-9-28-17-AM</a>

```
const user = await User.findOne({ where: { id: userId }})
const submissions = await Submission.findAll({ where: { userId: userId }})
const lessonStatus = await UserLesson.findAll({ where: { userId: userId }})
```

<a href="https://ibb.co/QFm19Kc">Screen-Shot-2020-04-24-at-9-28-44-AM</a>

```
const [ user, submissions, lessonStatus ] = await Promise.all([
  User.findOne({ where: { id: userId }}),
  Submission.findAll({ where: { userId: userId }}),
  UserLesson.findAll({ where: { userId: userId }})
])
```

<a href="https://ibb.co/jfYh1NY">Screen-Shot-2020-04-24-at-9-32-33-AM</a>

graphql/resolvers.ts
```diff
@@ -1,23 +1,39 @@
...
+    async session(_parent: void, _args: void, context: { req: Request }) {
+      const userId = _.get(context, 'req.session.userId', false)
+
+      if (!userId) {
+        return null
+      }
+
+      const user = await User.findOne({
+        where: { id: userId },
+        include: [Submission, UserLesson]
```
hwong0305, line 25: include: `Lesson` not `UserLesson`

coltonehrman: But that wouldn't get me the lesson status information for that user, would it?

helpers/dbload.ts
```diff
@@ -66,6 +66,14 @@ User.belongsToMany(Lesson, {
   through: { model: UserLesson }
 })
 
+User.hasMany(Submission, {
+  foreignKey: 'userId'
+})
+
+User.hasMany(UserLesson, {
```
hwong0305, line 73: This is incorrect. UserLesson is a join table that is used to manage many to many relationships

hwong0305: We can resolve this.

Reference 
https://sequelize.org/master/manual/advanced-many-to-many.html

graphql/resolvers.ts
```diff
@@ -1,23 +1,40 @@
...
         order: [['order', 'ASC']]
       })
+    },
+    async session(_parent: void, _args: void, context: { req: Request }) {
+      const userId = _.get(context, 'req.session.userId', false)
+
+      if (!userId) {
+        return null
+      }
+
```
songz, line 22: would be nice to add a comment here to link to the wiki page: https://github.com/garageScript/c0d3-app/wiki/Sequelize-Query-Performance

songz: `// FYI: The reason we are querying with parallelized promises: https://github.com/garageScript/c0d3-app/wiki/Sequelize-Query-Performance`

coltonehrman: thanks! :)

PR #160, opened 2020-04-22
Close #157 - Server tests

### LINKED ISSUE #157: [backend] - Add server tests

aortizoj15:


__tests__/pages/api/graphql.test.js
```diff
@@ -0,0 +1,72 @@
+/*
+ * @jest-environment node
```
Wolfy64, line 2: What does it do? I can't found on the doc...

songz: I think `jest` tests are by default running in windows environment (to provide browser functionalities). This causes issues for some modules (like sequelize) that relies on core nodejs modules. Setting the env will make this test run in node env. @aortizoj15 could you add a comment?

PR #170, opened 2020-04-24
Backend Signup Functionality

hwong0305:
This PR should be implemented after #169 . This will make the changes easier to read.

### DIFF
--- a/graphql/typeDefs.ts
+++ b/graphql/typeDefs.ts
@@ -8,6 +8,13 @@ export default gql`
   type Mutation {
     login(username: String!, password: String!): AuthResponse
     logout: AuthResponse
+    signup(
+      firstName: String!
+      lastName: String!
+      email: String!
+      username: String!
+      password: String
+    ): AuthResponse
   }
 
   type AuthResponse {
--- a/helpers/controllers/authController.test.js
+++ b/helpers/controllers/authController.test.js
@@ -6,6 +6,8 @@ import db from '../dbload'
 import { login, logout, signup } from './authController'
 import { chatSignUp } from '../mattermost'
 
+import { ApolloError } from 'apollo-server-micro'
+
 describe('auth controller', () => {
   let userArgs
   beforeEach(() => {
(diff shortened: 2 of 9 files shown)


helpers/controllers/authController.ts
```diff
@@ -68,3 +84,95 @@ export const logout = async (_parent: void, _: void, ctx: { req: Request }) => {
...
+  const name = `${firstName} ${lastName}`
+  const hash = await bcrypt.hash(password, 10)
+
+  const userRecord = User.create({
+    name,
+    username,
+    password: hash,
+    email,
+    emailVerificationToken: randomToken
+  })
```
songz, line 143: create this after `chatSignUp` is successful.

helpers/controllers/authController.test.js
```diff
@@ -1,18 +1,26 @@
...
+      password: 'c0d3reallyhard',
+      firstName: 'testuser1',
+      lastName: 'testuser1',
+      email: 'testuser@c0d3.com'
+    }
   })
 
   test('Login - should throw error when session is null', async () => {
-    expect(
+    return expect(
```
aortizoj15, line 23: Why is return needed here now? Just asking for my own knowledge. Thanks!

hwong0305: If you're testing an async function and you're not using `await` or `return` then the test won't fail if it resolves instead of rejecting.

graphql/typeDefs.ts
```diff
@@ -8,6 +8,13 @@ export default gql`
   type Mutation {
     login(username: String!, password: String!): AuthResponse
     logout: AuthResponse
+    signup(
+      firstName: String!
```
coltonehrman, line 12: What does the `!` mean vs without it?

helpers/controllers/authController.test.js
coltonehrman: This probably explains why I couldn't test something awhile back to see if it was throwing an error! lol

helpers/controllers/authController.test.js
```diff
@@ -81,4 +89,75 @@ describe('auth controller', () => {
...
+  test('Signup - should resolve with success false if userinfo is incomplete', async () => {
+    const result = await signup({}, {}, { req: { session: {} } })
+    expect(result).toEqual({
+      success: false,
+      error: 'Invalid registration information'
+    })
+  })
+
+  test('Signup - should resolve with success false if user already exists', async () => {
+    db.User.findOne = jest.fn().mockReturnValue({ username: 'c0d3user' })
```
coltonehrman, line 102: Should this not return `username: 'testuser'` to match `userArgs`?

coltonehrman: I guess since you mock it the return value in itself doesn't matter as long as something is returned?

helpers/controllers/authController.test.js
```diff
@@ -81,4 +89,75 @@ describe('auth controller', () => {
...
+    const result = await signup({}, userArgs, { req: { session: {} } })
+    expect(result).toEqual({
+      success: false,
+      error: 'User already exists'
+    })
+  })
+
+  test('Signup - should resolve with success false if email already exists', async () => {
+    db.User.findOne = jest
```
coltonehrman, line 111: This test structure seems odd to me (this could be since I'm still new to all this testing stuff lol)
But like, this test case, for example, seems to depend on the one before it, and it's also mocking a function return value twice which at first glance throws me off until I look at the `signup` code.

And also, I feel like the mocked return value's don't really make sense to me. Like if I testing a function that should tell me if the email already exists shouldn't I mock something like `{ username: '...', email: '[expected email]' }`

helpers/controllers/authController.test.js
```diff
@@ -81,4 +89,75 @@ describe('auth controller', () => {
...
+  test('Signup - should not create user if chat signup response with 401 or 403', async () => {
+    db.User.findOne = jest.fn().mockReturnValue(null)
+    db.User.create = jest.fn()
+
+    chatSignUp.mockResolvedValueOnce({
+      success: false,
+      error: 'Mattermost signup error'
+    })
+
+    chatSignUp.mockRejectedValueOnce({
```
coltonehrman, line 131: Why do we have to mock both the return and rejected value?

coltonehrman: Shouldn't it either resolve or reject, not both?

helpers/mattermost.test.js
```diff
@@ -0,0 +1,66 @@
...
+      email: 'testuser@c0d3.com'
+    }
+  })
+  test('ChatSignUp - should resolve if response is 201', async () => {
+    fetch.mockResolvedValue({ status: 201 })
+    return expect(
+      chatSignUp(userArgs.username, userArgs.password, userArgs.email)
+    ).resolves.toEqual({ success: true })
+  })
+  test('ChatSignUp - should reject if response is 401', async () => {
```
coltonehrman, line 21: If this is testing for reject, should it not be `expect().rejects`?

helpers/mattermost.test.js
```diff
@@ -0,0 +1,66 @@
...
+  test('ChatSignUp - should reject if response status is invalid', async () => {
+    fetch.mockResolvedValue({ status: 418 }) // MatterMost only returns 201, 401 and 403
+    return expect(
+      chatSignUp(userArgs.username, userArgs.password, userArgs.email)
+    ).resolves.toEqual({
+      success: false,
+      error: 'Unknown error (Default Case)'
+    })
+  })
+  test('ChatSignUp - should reject if internal server error', async () => {
```
coltonehrman, line 48: Is this a dup of previous test?

helpers/mattermost.test.js
```diff
@@ -0,0 +1,66 @@
...
+    fetch.mockRejectedValue({})
```
coltonehrman, line 58: Should there not be a `status` code?

coltonehrman: Ah nevermind, I see you are mocking the `rejected` value. So if I do this, it will automatically make the mock function reject instead of resolve?

graphql/typeDefs.ts
Wolfy64: `!` mean the data can not be null

helpers/controllers/authController.test.js
songz: It doesn't matter, it just have to return some user. A nitpick would be to return the same username though, it'll make prevent questions like this in the future.

helpers/controllers/authController.test.js
songz: for this, I looked at the code. The first check was for username (which returned null), the second check (which is for email) returned an object (meaning user found). What you are saying would make sense if this function was a `getByEmail` function or something, so maybe the description could be better:

`Should resolve with false If second user lookup (which is email lookup) returns a user object, meaning a user was found by email.`

helpers/controllers/authController.test.js
songz: reject I think may be caused by error (like request network failed), and success false means there may be an validation error that mattermost sent back.

helpers/controllers/authController.test.js
hwong0305: `userArgs` is setup to be a valid entry for signup. In this case, I wanted to make sure it would error out so all the fields aren't filled out.

Issue #164, opened 2020-04-24
[Lesson Page] - Submission diff is off and reviewer comments missing


songz:
# Issue 1
Submission diff is really off from the original, I feel like this is a regression. V2 should be better!
![Screen Shot 2020-04-23 at 5 53 32 PM](https://user-images.githubusercontent.com/686933/80164320-7fd88280-858d-11ea-9c4f-fbb0c6f373b9.png)

# Issue 2
1. Change submitter name (this is obvious) to submitted at instead.
![Screen Shot 2020-04-23 at 6 04 35 PM](https://user-images.githubusercontent.com/686933/80164327-8404a000-858d-11ea-8fee-c67054777605.png)
2. Add Review section (render markdown)

PR #171, opened 2020-04-24
[Challenge Diff] - Parse Diff + Styling

coltonehrman:
Add functionality to parse diff from user submission and display appropriately.
Added syntax highlighting with PrismJS

Meat of code changes:
https://github.com/garageScript/c0d3-app/pull/171/files?file-filters%5B%5D=.json&file-filters%5B%5D=.scss&file-filters%5B%5D=.tsx#diff-788b6d187ee49d6ab0239c3b537e6555

Storybook changes:
https://deploy-preview-171--youthful-thompson-448332.netlify.app/?path=/story/components-challengematerial--with-diff

Takes care of some issues in #164  

NOTE:
*Will no longer render something that is not a valid diff*


components/ChallengeMaterial.tsx
```diff
@@ -102,6 +104,47 @@ export const ChallengeQuestionCard: React.FC<ChallengeQuestionCardProps> = ({
...
+        str,
+        Prism.languages.javascript,
+        'javascript'
+      )
+      return <span dangerouslySetInnerHTML={{ __html: language }} />
+    }
+
+    return (
+      <ReactDiffViewer
+        key={i}
```
Wolfy64, line 139: It is not a best recommended to use `index` as `key` with React.
Since we use lodash you might use `_.uniqueId()` [link](https://lodash.com/docs/4.17.15#uniqueId)

coltonehrman: updated :+1:

components/ChallengeMaterial.tsx
```diff
@@ -102,6 +104,47 @@ export const ChallengeQuestionCard: React.FC<ChallengeQuestionCardProps> = ({
...
+        if (change.isDelete) oldValue.push(change.content)
+        else if (change.isInsert) newValue.push(change.content)
+        else {
+          oldValue.push(change.content)
+          newValue.push(change.content)
+        }
+      })
+    })
+
+    const syntaxHighlight = (str: string): any => {
```
hwong0305, line 126: Try setting return type to JSX.Element

coltonehrman: It won't allow me because of my first return statement when `string` doesn't exist. Also, I tried `JSX.Element | null` and it conflicted with the prop that I am passing this function into

components/ChallengeMaterial.tsx
```diff
@@ -102,6 +104,47 @@ export const ChallengeQuestionCard: React.FC<ChallengeQuestionCardProps> = ({
   const { data } = React.useContext(SessionContext)
   const username = _.get(data, 'userInfo.username', '')
   const diff = _.get(currentChallenge, 'submission.diff', false)
+  let files = null
+
+  if (diff) files = gitDiffParser.parse(diff)
```
michaelbayday, line 109: can this be done with the ternary operator? 
`const files = (diff) ? gitDiffParser.parse(diff) : null`

PR #173, opened 2020-04-25
Refactoring Login to Match Backend Engineering Standards

hwong0305:


helpers/controllers/authController.test.js
```diff
@@ -27,23 +27,19 @@ describe('auth controller', () => {
...
-    expect(result).toEqual({
-      success: false,
-      error: 'user does not exist'
-    })
+    return expect(
+      login({}, userArgs, { req: { session: {} } })
+    ).rejects.toThrowError('User does not exist')
   })
 
   test('Login - should return success false if password is invalid', async () => {
```
coltonehrman, line 37: Description wasn't updated with test

helpers/controllers/authController.ts
```diff
@@ -26,35 +26,33 @@ export const login = async (
   arg: Login,
   ctx: { req: Request }
 ) => {
-  const {
-    req: { session }
-  } = ctx
-  const { username, password } = arg
+  try {
```
coltonehrman, line 29: Why do you wrap this in a `try...catch` when the catch statement just re-throws the error?

hwong0305: If there's any special case errors from `Sequelize` or `bcrypt` won't propagate properly to `ApolloServer` if it doesn't have a catch.

songz: @coltonehrman not catching will cause nodeJS to `console.warn` that we are not handling error when there are rejections. We probably don't want that if we can help it.

PR #176, opened 2020-04-28
updated#106 display UI for profile lessons

rkalra247:


components/ProfileLessons.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const displayLessons = lessons.map((lesson, i) => {
+    if (!lesson.progress) {
+      lesson.progress = 0
+    }
+    return (
+      <LessonImage
+        key={i}
+        order={lesson.order}
+        progress={lesson.progress + '%'}
+      ></LessonImage>
```
michaelbayday, line 68: Self closing tag probably could be used here

components/ProfileLessons.tsx
```diff
@@ -0,0 +1,82 @@
...
+const LessonImage: React.FC<lesson> = ({ progress, order }) => {
+  let color = 'primary'
+  if (progress === '100%') {
+    color = 'success'
+  }
+
+  let opacity = 1
+  let badge = (
+    <p
+      style={{
```
michaelbayday, line 26: Can we put the styling our scss files?

components/ProfileLessons.tsx
```diff
@@ -0,0 +1,82 @@
...
+  lessons: Lesson[]
+}
+
+type lesson = {
+  order: number
+  progress: string
+}
+
+const LessonImage: React.FC<lesson> = ({ progress, order }) => {
+  let color = 'primary'
```
michaelbayday, line 18: Maybe a ternary operator here?

`Const color = (lesson.progress === 100%) ? ‘success’ : ‘primary’`

components/ProfileLessons.tsx
```diff
@@ -0,0 +1,82 @@
...
+type Lesson = {
+  order: number
+  progress?: number
+}
+
+type LessonsProp = {
+  lessons: Lesson[]
+}
+
+type lesson = {
```
michaelbayday, line 12: Do you think we need another type here or do you think that we can just check it lesson.progress doesn’t exist that we render the opacity, etc accordingly?

rkalra247: > Do you think we need another type here or do you think that we can just check it lesson.progress doesn’t exist that we render the opacity, etc accordingly?

It will always have lessons.progress, i added that condition before it renders.

michaelbayday: Spoke with Rahul. To be discussed later, but I’m not sure if we should make another type if only one prop is different and the absence of the prop can be handled by JS. Should we maybe just assume if the lesson.progress key does not exist that we should just render accordingly?

stories/profile/ProfileLessons.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+  { order: 4, progress: 0 },
+  { order: 5 },
+  { order: 6, progress: 80 },
+  { order: 7 },
+  { order: 8, progress: 100 },
+  { order: 9, progress: 0 }
+]
+
+export const _ProfileLessons: React.FC = () => (
+  <ProfileLessons lessons={lessons}></ProfileLessons>
```
aortizoj15, line 23: Self close tag here since no children

rkalra247: thanks :)

scss/profileLessons.scss
```diff
@@ -0,0 +1,16 @@
+.lessonImage_container{
```
coltonehrman, line 1: not sure how i feel about the camelcase class names? lol i know its a little nitpick, but maybe we should match the current class name styles?

components/ProfileLessons.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import '../scss/profileLessons.scss'
+
+type Lesson = {
```
coltonehrman, line 4: maybe rename to LessonImageProps

components/ProfileLessons.tsx
```diff
@@ -0,0 +1,49 @@
...
+  order: number
+  progress?: number
+}
+
+type LessonsProp = {
```
coltonehrman, line 9: maybe rename to ProfileLessonsProps

scss/profileLessons.scss
rkalra247: \

> not sure how i feel about the camelcase class names? lol i know its a little nitpick, but maybe we should match the current class name styles?

I am following current classNames structure, i followed LessonCard template in name convention for classes

rkalra247: > not sure how i feel about the camelcase class names? lol i know its a little nitpick, but maybe we should match the current class name styles?

https://share.getcloudapp.com/GGuk2mXN

components/ProfileLessons.tsx
```diff
@@ -0,0 +1,49 @@
...
+      )}
+      <img src={`/curriculumAssets/lessonCoversSvg/js-${order}-cover.svg`} />
+    </div>
+  )
+}
+
+const ProfileLessons: React.FC<LessonsProp> = ({ lessons }) => {
+  const displayLessons = lessons.map((lesson, i) => {
+    return (
+      <LessonImage key={i} order={lesson.order} progress={lesson.progress} />
```
coltonehrman, line 35: use _.uniqueId() for keys. referring to @Wolfy64 comment

> It is not recommended to use `index` as `key` with React.
> Since we use lodash you might use `_.uniqueId()` [link](https://lodash.com/docs/4.17.15#uniqueId)

scss/profileLessons.scss
coltonehrman: we don't have any camelCase class name though? it should be more like `lesson-image__container`

rkalra247: hmmm got it

components/ProfileLessons.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import '../scss/profileLessons.scss'
```
michaelbayday, line 2: @coltonehrman thoughts on importing this scss file into our main scss file?

songz: this is specific to this file tho.

michaelbayday: We are doing this for other styles that are specific to the component too though ie challenge material (granted I worked on this)

coltonehrman: i mean unless we decide to stick with importing everything into main or not it doesnt really matter at this point. but i do think we should decide to do one or the other, not mixing it. so nobody gets confused

hwong0305: Let's not import everything into main. We should import styles as we need it.

components/ProfileLessons.tsx
hwong0305: Let's not use uniqueId. We don't need it. Let's use `lesson.order` since each lesson is unique.

components/ProfileLessons.tsx
coltonehrman: but wouldnt it be fair to say we need every style? otherwise, that would mean we have useless scss files.plus in the end it'll all be bundled together the same basically right?

hwong0305: It's fair to say that, but you don't need every style in every page / every component. This means you lose scoped scss functionality.

coltonehrman: so i believe the way we are importing the scss files currently has no scope `import './style.scss`... i think you have to do `import styles from './style.scss` then reference it in the component code

coltonehrman: https://nextjs.org/docs/basic-features/built-in-css-support

songz: @coltonehrman `but wouldnt it be fair to say we need every style? otherwise, that would mean we have useless scss files.plus in the end it'll all be bundled together the same basically right?` The way I see it, keeping everything together makes things easier to track. So one example is... if I remove this component, I may forget to remove from `main.scss` and code reviewers wouldn't know. If you put the `scss` together in the component, then our css could be cleaner.

Issue #183, opened 2020-04-29
[Bug] - User cannot login


songz:
![Screen Shot 2020-04-28 at 9 27 45 PM](https://user-images.githubusercontent.com/686933/80560780-27432400-8997-11ea-95f4-a0ab3e8417e7.png)

coltonehrman:
I believe this one is ready to be closed?

PR #185, opened 2020-04-29
[Bug Fix] Fix login + add logout feature

### LINKED ISSUE #181: Logout button does not work

coltonehrman:
- ~~Removed max character check inside signup and login forms~~
- Increased max character check to 64 for signup & login forms
- Added logoutUser helper function
- Added tests for logoutUser helper function
- Modified Button component to have onClick prop

Closes #181 #183

### DIFF
--- a/__tests__/components/Button.test.js
+++ b/__tests__/components/Button.test.js
@@ -1,0 +1,18 @@
+import React from 'react'
+import { render, fireEvent } from '@testing-library/react'
+import Button, { noop } from '../../components/Button'
+
+describe('Button Component', () => {
+  test('noop func', () => {
+    expect(noop()).toEqual(undefined)
+  })
+
+  test('Should call onClick prop when clicked', () => {
+    const onClick = jest.fn()
+    const { getByRole } = render(<Button onClick={onClick} />)
+    const leftClick = { button: 0 }
+
+    fireEvent.click(getByRole('button'), leftClick)
+    expect(onClick).toHaveBeenCalledTimes(1)
+  })
+})
--- a/helpers/formValidation.tsx
+++ b/helpers/formValidation.tsx
@@ -1,6 +1,7 @@
 import * as Yup from 'yup'
 
 const TEXT_MIN = 2
+const TEXT_MAX = 64
 const PASSWORD_MIN = 6
 const REGEX_ALPHANUMERICS_AND_SPACE = /^[a-zA-Z0-9_\s]*$/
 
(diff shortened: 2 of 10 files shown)


__tests__/components/Button.test.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { render, fireEvent } from '@testing-library/react'
+import Button, { noop } from '../../components/Button'
+
+describe('Button Component', () => {
+  test('noop func', () => {
```
coltonehrman, line 6: Had to test `noop` function for 100% coverage, we may want to look into putting this into a global helper function to be used and tested individually

helpers/formValidation.tsx
```diff
@@ -1,7 +1,6 @@
 import * as Yup from 'yup'
 
 const TEXT_MIN = 2
-const TEXT_MAX = 20
```
Wolfy64, line 4: Why remove `TEXT_MAX` instead of just increase the value? We should have a limit.

Wolfy64:
A good practice would be have to split this in 2 PR:
- Fix login #183
- add logout feature #181
To pick an issue please assign yourself first @aortizoj15 was already working on it no?

helpers/formValidation.tsx
rkalra247: > Why remove `TEXT_MAX` instead of just increase the value? We should have a limit.

is that an industry practice to define max characters? I have seen limitations on passwords but never with usernames.

rkalra247: > > Why remove `TEXT_MAX` instead of just increase the value? We should have a limit.
> 
> is that an industry practice to define max characters? I have seen limitations on passwords but never with usernames.

What are consequences for not having max characters?

__tests__/components/Button.test.js
Wolfy64: We might also rename it. (Maybe noOperationFunction, dummyFunc ...etc)
`noop` naming was not straightforward for me I had to google it.
Code should be easy to read.

helpers/formValidation.tsx
Wolfy64: Maybe it won't do anything.
There is a rule to never trust users input... in general
What will happen if user choose to input a lorem ipsum or whathever? maybe it will work maybe it will break something?
You can keep this control with rules.
But I agree 20 is a bit to low 🙂

__tests__/components/Button.test.js
coltonehrman: `noop` is a standard name for this type of function though

Wolfy64: Hoo cool I didn't know 🙂

songz: @coltonehrman is right, `noop` is standard name for useless function. Next time, the information comment should live with the code (add a comment in the codebase)

helpers/formValidation.tsx
songz: valid point @Wolfy64 , didn't think of that.

hwong0305: I don't think this matters. Since we follow good practice, the password is hashed to a certain length. Even if you hash lorem ipsum, it is still valid.

songz: https://github.com/garageScript/c0d3-app/issues/187

coltonehrman:
> A good practice would be have to split this in 2 PR:
> 
> * Fix login #183
> * add logout feature #181
> 
> To pick an issue please assign yourself first @aortizoj15 was already working on it no?

Yeah, I just figured since it was a live bug that should be taken care of ASAP I do it all together.
I informed @aortizoj15 to in case he wanted to take over :)

PR #190, opened 2020-04-30
Migrate to New Backend

SahilKalra98:


helpers/loginUser.ts
```diff
@@ -1,15 +1,20 @@
 const SERVER_URL = process.env.SERVER_URL
 
 export const loginUser = async (username: string, password: string) => {
-  const res = await fetch(`${SERVER_URL}/signin`, {
+  const res = await fetch(`${SERVER_URL}`, {
```
coltonehrman, line 4: Should we be using the `react/apollo` hooks `useMutation` and `useQuery` for the new backend?

pages/curriculum.tsx
```diff
@@ -16,23 +17,30 @@ export const Curriculum: React.FC<WithQueryProps> = ({ queryData }) => {
...
+  const { lessons, session }: { lessons: Lesson[]; session: any } = queryData
+  const sortedLessonsById: Lesson[] = lessons.sort((a, b) => a.id - b.id)
+  const sortedLessonStatus = session.lessonStatus.sort(
+    (a: any, b: any) => a.id - b.id
   )
-  const lessonInProgressIdx = sortedLessons.findIndex(
-    lesson => !lesson.currentUser.userLesson.isPassed
-  )
-  const lessonsToRender: React.ReactElement[] = sortedLessons.map(
+  const sortedLessonsMap = sortedLessonsById
```
hwong0305, line 25: Let's not sort lesson here. Use Sequelize to sort the lessons instead.

helpers/useSession.tsx
```diff
@@ -16,7 +16,7 @@ export type SessionData = {
 }
 
 const useSession = (): SessionData => {
-  const { data, error } = useSWR(`${SERVER_URL}/session`, fetcher)
+  const { data, error } = useSWR(`${SERVER_URL}`, fetcher)
```
hwong0305, line 19: We don't need `useSWR` anymore. Use `apolloClient`

graphql/queries.ts
```diff
@@ -1,79 +1,43 @@
...
-    $lessonInfo: LessonId
-    $lessonUserInfo: LessonUserId
-    $lessonStatus: LessonId
+export const SIGNUP_USER = gql`
+  mutation signup(
+    $firstName: String!
+    $lastName: String!
+    $email: String!
+    $username: String!
+    $password: String
```
hwong0305, line 29: should be `String!`

pages/login.tsx
```diff
@@ -20,14 +21,17 @@ type Values = {
...
-    const data = await loginUser(values.username, values.password)
-    if (data) {
-      return (window.location.pathname = '/curriculum')
+    try {
+      const { data } = await loginUser({ variables: values })
+      if (data) {
+        return (window.location.pathname = '/curriculum')
+      }
+    } catch (error) {
+      setIsAlertVisible(true)
```
hwong0305, line 33: We need to show two different error messages:

- Invalid password
- User not found

components/AppNav.tsx
```diff
@@ -42,23 +44,31 @@ const AuthLink = () => (
...
-  </div>
-)
+const AuthButton: React.FC<AuthButtonProps> = ({ initial, username }) => {
+  const [logoutUser, { data }] = useMutation(LOGOUT_USER)
+  useEffect(() => {
+    const { success } = _.get(data, 'logout', false)
+    if (success) {
+      window.location.pathname = '/'
+    }
+  })
```
coltonehrman, line 54: Make sure to add `[data]` as dependencies to `useEffect`

helpers/useSession.tsx
```diff
@@ -1,23 +1,37 @@
 import useSWR from 'swr'
 import { User } from './models/User'
+import _ from 'lodash'
 
 const SERVER_URL = process.env.SERVER_URL
 
 export const fetcher = (url: string) =>
-  fetch(url, { credentials: 'include' }).then(r => r.json())
+  fetch(url, {
```
hwong0305, line 8: Try using this instead: 

```
import { request } from 'graphql-request'

const API = 'https://api.graph.cool/simple/v1/movies'
const fetcher = query => request(API, query)

function App () {
  const { data, error } = useSWR(
    `{
      Movie(title: "Inception") {
        releaseDate
        actors {
          name
        }
      }
    }`,
    fetcher
  )
  // ...
}
```

source: https://github.com/zeit/swr#data-fetching

@types/lesson.d.ts
```diff
@@ -2,6 +2,7 @@ export type LessonStatus = {
   isEnrolled: string | null
   isPassed?: string
   isTeaching: string | null
+  lessonId: string
```
Wolfy64, line 5: I know we did not do on other types but should we not use `userId: ID!` instead of `string`

@types/signup.d.ts
```diff
@@ -16,6 +16,6 @@ export type ErrorDisplayProps = {
 }
 
 export type SignupErrors = {
-  userName?: string[]
-  confirmEmail?: string[]
+  userName?: boolean
+  confirmEmail?: boolean
```
Wolfy64, line 20: When we use boolean for a variable it's a good practice to start their name with **is/has**:
- `hasName` | `hasEmail`
- `isUserNameValid` | `isEmailValid`

helpers/withApollo.ts
```diff
@@ -6,7 +6,7 @@ import { InMemoryCache } from 'apollo-boost'
 const SERVER_URL = process.env.SERVER_URL
 
 const link = createHttpLink({
-  uri: `${SERVER_URL}/graphql`,
+  uri: `${SERVER_URL}`,
```
Wolfy64, line 9: Not a big deal but `SERVER_URL` it's already a string you do not need to use string interpolation

helpers/useSession.tsx
```diff
@@ -1,23 +1,37 @@
...
-    success: boolean
-  }
-  error?: boolean
+  session: User | null
 }
 
 const useSession = (): SessionData => {
-  const { data, error } = useSWR(`${SERVER_URL}/session`, fetcher)
-  return { data, error }
+  const { data } = useSWR(`${SERVER_URL}`, fetcher)
```
Wolfy64, line 32: Not a big deal but `SERVER_URL` it's already a string you do not need to use string interpolation

@types/lesson.d.ts
hwong0305: `Int`is more appropiate

hwong0305: No. Because this is `TypeScript`not `GraphQL`

pages/signup.tsx
```diff
@@ -129,13 +131,26 @@ const SignupForm: React.FC<SignupFormProps> = ({
...
+      }
+    } catch (error) {
+      const newSignupErrors = { ...signupErrors }
+      const message = _.get(error, 'message', '')
+      if (message) {
+        message.indexOf('Email') !== -1
+          ? (newSignupErrors.confirmEmail = true)
+          : (newSignupErrors.userName = true)
+      }
+      setSignupErrors(Object.assign({}, newSignupErrors))
```
Wolfy64, line 153: This code confuse me...
we pass an object `error` to `_.get(error, 'message', '')` that will check if we have the key `Email`
if `false` that will return an empty `string`
if `true` that will return an `array`
then we mutate the `newSignupErrors` object and we update the state with a new object again
Do I have correct?
I feel it really confusing...

pages/curriculum/[lesson].tsx
```diff
@@ -1,22 +1,34 @@
...
 
 const Challenges: React.FC<WithQueryProps> = ({ queryData }) => {
   const {
-    lessonInfo: currentLesson,
-    userSubmissions,
-    lessonStatus: currentLessonStatus
-  } = queryData
+    lessons,
+    session
+  }: { lessons: Lesson[]; session: Session } = queryData
```
Comment (michaelbayday) [inline pages/curriculum/[lesson].tsx:20]: is typing here necessary if we're already typing the component props withqueryprops? I can see the incentive of it being more explicit for new people to the codebase, but seems kind of unnecessary

helpers/useSession.tsx
coltonehrman, line 32: We should just be replacing `useSWR` with the GET_APP query right?

coltonehrman: Unless you only want to query for the username, you can create a new query string for that

SahilKalra98: I get an error If I call useQuery() within useSession

pages/signup.tsx
SahilKalra98: yeah so when you catch the error it is a ApolloClient error object and the error is thrown here as a UserInputError:

https://github.com/garageScript/c0d3-app/blob/master/helpers/controllers/authController.ts#L110-L122 

So then when I do ```_.get(error, 'message', '')``` It will give me the error message string or empty string. Then I check if message exists and if it does then I see if the message has the substring ```Email``` in it meaning that the error message is about the email otherwise if ```message.indexOf('Email')``` returns -1 then it would be the ```User already Exists!``` error message.

Then based on that I update the signupErrors and update the state, where the signup errors are passed into the SignupForm component right below the above code 
```<SignupForm handleSubmit={handleSubmit} signupErrors={signupErrors} />```

If there is still any confusion or if you if a better implementation idea let me know!

File: helpers/useSession.tsx (hunk 11 above)
Comment (coltonehrman) [reply helpers/useSession.tsx:32]: In that case I would probably replace `useSession` everywhere with the `useQuery` hook. Makes more sense :) Tbh, when I added `useSession` it wasn't very useful in the beginning anyways lol

Comment (songz) [reply helpers/useSession.tsx:32]: @coltonehrman this was tried, but failed. I think that could be a step 2 migration, otherwise this PR may never finish.

File: pages/signup.tsx (hunk 12 above)
Comment (michaelbayday) [reply pages/signup.tsx:153]: If message just a string maybe ```.contains``` would work better for understanding the code.

Is it not possible to have multiple Errors? Seems like the backend is working to only throw one error right now, not sure if we want to aggregate errors and display multiple.

File: pages/curriculum/[lesson].tsx (hunk 13 above)
Comment (hwong0305) [reply pages/curriculum/[lesson].tsx:20]: We need typing if it's not implicitly provided to `TypeScript`. 

The idea of types is so someone doesn't accidentally assigned something of a different type to variable eliminating the possibility of some errors.

File: pages/signup.tsx (hunk 12 above)
Comment (hwong0305) [reply pages/signup.tsx:153]: Instead of doing this, why don't we just `useEffect` to check if error exists. 

It seems way simpler than using this logic. There's no need to write our own logic where `ApolloClient` provides this.

File: pages/signup.tsx [hunk 14]
```diff
@@ -127,26 +134,63 @@ const SignupForm: React.FC<SignupFormProps> = ({
...
+    setIsSubmitting(true)
+    try {
+      const { data } = await signupUser({ variables: values })
+      if (data.signup.success) {
+        return setSignupSuccess(true)
+      }
+      const err = new Error(
+        'Server cannot be reached. Please try again. If this problem persists, please send an email to support@c0d3.com'
+      )
+      throw err
```
hwong0305, line 152: You can just throw the error since the err isn't being used for anything.

components/AppNav.tsx
```diff
@@ -42,23 +44,31 @@ const AuthLink = () => (
...
-      }}
-    />
-  </div>
-)
+const AuthButton: React.FC<AuthButtonProps> = ({ initial, username }) => {
+  const [logoutUser, { data }] = useMutation(LOGOUT_USER)
+  useEffect(() => {
+    const { success } = _.get(data, 'logout', false)
+    if (success) {
+      window.location.pathname = '/'
```
hwong0305, line 52: `router.push('/')`

```
...
const router = useRouter()
...

```

pages/_app.tsx
```diff
@@ -14,10 +14,6 @@ interface IProps extends AppProps {
 function MyApp({ Component, pageProps, apollo }: IProps) {
   const session = useSession()
 
-  if (!session.data && !session.error) {
-    return null
```
hwong0305, line 18: You probably need this for testing purposes?

components/AppNav.tsx
hwong0305: We're no longer going to refactor to router.push since router doesn't trigger reload. There isn't a nice way for router to reload the page.

helpers/useSession.tsx
hwong0305: We are going to refactor `context` out. So this can be resolved.

helpers/useSession.tsx
hwong0305: STEP 2 migration should take out context since we don't need it. This can be resolved.

__tests__/components/__snapshots__/Layout.js.snap
```diff
@@ -237,7 +237,7 @@ exports[`Layout Component Should render correctly when user session found 1`] =
         <button
           class="btn border btn-secondary overflow-hidden p-2 text-truncate"
         >
-          tester
```
hwong0305, line 240: lol

components/AppNav.tsx
```diff
@@ -48,23 +50,31 @@ const AuthLink = () => (
...
+    <div>
+      <Button
+        btnType="border btn-secondary overflow-hidden p-2 text-truncate"
+        initial={initial}
+        text={username}
+      />
+      <Button
+        text="Logout"
+        btnType="border btn-secondary ml-2"
+        onClick={async () => {
```
songz, line 71: `onClick = { logoutUser }`

__tests__/pages/signup.js
```diff
@@ -39,107 +43,90 @@ describe('Signup Page', () => {
...
   })
 
-  test('Should submit signup form values and display error component upon failure of both username and email', async () => {
-    signupHelper.signupUser = jest.fn().mockReturnValue(
-      Promise.resolve({
-        success: false,
-        errorMessage: {
-          confirmEmail: ['This email has been taken'],
-          userName: ['Username not available']
+  test('Should submit signup form values and display error component upon failure of email', () => {
```
songz, line 108: when user does not exist

PR #191, opened 2020-04-30
[FEATURE] Logging -  1 / 3 

hwong0305:
1. **Current** Add logic for logging using winston.
2. Add support for Sentry and others for logging
3. Add logging functionality to GraphQL

### DIFF
--- a/helpers/logger.ts
+++ b/helpers/logger.ts
@@ -1,19 +1,35 @@
+import util from 'util'
 import { nanoid } from 'nanoid'
 import winston from 'winston'
 import { TransformableInfo } from 'logform' // Types for Winston
 import { LoggedRequest } from '../@types/helpers'
 import { NextApiResponse } from 'next'
 
+export const processArgs = (inputArr: any) => {
+  if (Array.isArray(inputArr)) {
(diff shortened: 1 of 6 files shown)


helpers/logger.ts
```diff
@@ -0,0 +1,29 @@
...
+      winston.format.colorize(),
+      winston.format.printf(printFunc)
+    ),
+    transports: [new winston.transports.Console()]
+  })
+}
+
+export default (req: any, res: any, next: any) => {
+  const uid = nanoid()
+  req.logger = winstonLogger(__filename)
```
songz, line 25: `uid` -> I think we want the request session as the label right? What is the best practice for labeling? If there isn't any, then I prefer to use the request id.

helpers/logger.ts
```diff
@@ -0,0 +1,29 @@
...
+  req.id = uid
```
songz, line 26:
```
req.info = 
req.warn = 
req.error = 
```

We want the 3 types of log directly in the request object, makes it easier to use. 

Also, winston's generic logger will only log strings. We should allow objects and error objects by stringifying it.

PR #194, opened 2020-05-02
update #161 show UI for Lessons challenges

rkalra247:


components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+    if (eachChallenge.challengeStatus === 'wrong answer') {
+      challengeStatus = 'unapproved_challenge_status'
+    }
+    if (eachChallenge.challengeStatus === 'pending') {
+      challengeStatus = 'pending_challenge_status'
+    }
+    return (
+      <span
+        key={_.uniqueId()}
```
michaelbayday, line 40: think @hwong0305 mentioned we should use something else for key

rkalra247: > think @hwong0305 mentioned we should use something else for key

thats how I did it with my last PR, not sure about what he recommended. Let me know :)

__tests__/components/__snapshots__/ChallengeMaterial.test.js.snap
```diff
@@ -335,7 +335,7 @@ exports[`Curriculum challenge page Should render first challenge that is not pas
           class="card-header bg-white"
         >
           Submitted 
-          16 days ago
+          18 days ago
```
hwong0305, line 338: lol

components/LessonsChallenges.tsx
hwong0305: I don't see a reason why we need to use `uniqueId`. React recommends that we don't use the `index` from map for our key. But we can use either the `id` or `order` from `Challenge`.

songz: using `id` or `order` is like the same thing as using `index` no? using `uniqueId` forces this to rerender. @Wolfy64 brought this up last time.

stories/profile/LessonsChallenges.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+  { challengeNumber: 5, challengeStatus: 'pending' },
+  { challengeNumber: 6, challengeStatus: 'pending' },
+  { challengeNumber: 7, challengeStatus: 'pending' },
+  { challengeNumber: 8, challengeStatus: 'pending' },
+  { challengeNumber: 9, challengeStatus: 'pending' },
+  { challengeNumber: 10, challengeStatus: 'pending' },
+  { challengeNumber: 11, challengeStatus: 'pending' }
+]
+
+const unapprovedChallenges = [
```
songz, line 53: `rejectedChallenges`

stories/profile/LessonsChallenges.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+  { challengeNumber: 2, challengeStatus: 'default' },
+  { challengeNumber: 3, challengeStatus: 'default' },
+  { challengeNumber: 4, challengeStatus: 'default' },
+  { challengeNumber: 5, challengeStatus: 'default' }
+]
+
+const lessons = [
+  {
+    correctAnswers: 10,
+    totalChallenges: 10,
```
songz, line 97: `totalChallenges` does not have to exist. The component could display this value by `challenges.length`

stories/profile/LessonsChallenges.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+  { challengeNumber: 1, challengeStatus: 'default' },
+  { challengeNumber: 2, challengeStatus: 'default' },
+  { challengeNumber: 3, challengeStatus: 'default' },
+  { challengeNumber: 4, challengeStatus: 'default' },
+  { challengeNumber: 5, challengeStatus: 'default' }
+]
+
+const lessons = [
+  {
+    correctAnswers: 10,
```
songz, line 96: `correctAnswers` does not have to exist. The component should get this value by adding up `challengeStatus: 'passed'` for all the challenges.

stories/profile/LessonsChallenges.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+  { challengeNumber: 2, challengeStatus: 'default' },
+  { challengeNumber: 3, challengeStatus: 'default' },
+  { challengeNumber: 4, challengeStatus: 'default' },
+  { challengeNumber: 5, challengeStatus: 'default' },
+  { challengeNumber: 6, challengeStatus: 'default' },
+  { challengeNumber: 7, challengeStatus: 'default' },
+  { challengeNumber: 8, challengeStatus: 'default' }
+]
+
+const challengesLesson5 = [
```
songz, line 74: `const challenges5=` <- this could be removed, it is extra case that is not necessary.

stories/profile/LessonsChallenges.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+  { challengeNumber: 10, challengeStatus: 'passed' }
+]
+
+const defaultChallenges = [
+  { challengeNumber: 1, challengeStatus: 'default' },
+  { challengeNumber: 2, challengeStatus: 'default' },
+  { challengeNumber: 3, challengeStatus: 'default' },
+  { challengeNumber: 4, challengeStatus: 'default' },
+  { challengeNumber: 5, challengeStatus: 'default' },
+  { challengeNumber: 6, challengeStatus: 'default' },
```
songz, line 30: could you make some of these challengeStatus null? or some does not even have this key?

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,75 @@
...
+  challengeStatus: string
+}
+
+export const ChallengeStatus: React.FC<ChallengeStatusProps> = props => {
+  const displayChallengeStatus = props.challengesData.map(eachChallenge => {
+    let challengeStatus = 'default_challenge_status'
+    if (eachChallenge.challengeStatus === 'passed') {
+      challengeStatus = 'passed_challenge_status'
+    }
+    if (eachChallenge.challengeStatus === 'wrong answer') {
```
songz, line 32: we should stick with using one word, `wrong`

rkalra247: understood

stories/profile/LessonsChallenges.stories.tsx
rkalra247: > `const challenges5=` <- this could be removed, it is extra case that is not necessary.

this lesson has uniqe number of challenges, i have to hard code it. How do i go about doing that without using this? Like add an if condition on front end and look for lesson number 5 and then only show that many questions by using the same data?

stories/profile/LessonsChallenges.stories.tsx
rkalra247: understood

stories/profile/LessonsChallenges.stories.tsx
rkalra247: understood

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const ChallengeStatus: React.FC<ChallengeStatusProps> = props => {
+  const displayChallengeStatus = props.challengesData.map(
+    (eachChallenge, challengeId) => {
+      let challengeStatus = 'default_challenge_status'
+      if (eachChallenge.challengeStatus === 'passed') {
+        challengeStatus = 'passed_challenge_status'
+      }
+      if (eachChallenge.challengeStatus === 'wrong') {
+        challengeStatus = 'unapproved_challenge_status'
```
songz, line 31: keep it consistent, `fail_challenge_status`

`unapproved` could cause confusion because `unapproved` means 2 things:
1. Haven't been reviewed yet.
2. fail

songz: `wrong_challenge_status` or `fail_challenge_status`. If you change it to `fail`, please update your `if` check to do: `eachChallenge.challengeStatus === 'fail'` so that it is consistent.

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <h6 className="challenges_stats">
+            {`${filterPassedChallenges.length} of ${lesson.challenges.length}  Challenges completed`}
+          </h6>
+          <ChallengeStatus challengesData={lesson.challenges} />
+        </div>
+      </div>
+    )
+  })
+  return (
+    <div className="card shadow-sm">
```
songz, line 69: could you give this a class `profile`?

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,78 @@
...
+          className={`challenge_status ${challengeStatus}`}
+        ></span>
+      )
+    }
+  )
+  return <div className="challenges_container">{displayChallengeStatus}</div>
+}
+
+const LessonsChallenges: React.FC<LessonChallengeProps> = props => {
+  let filterPassedChallenges = []
```
michaelbayday, line 48: Can you just declare this variable as a const on line 50? I don’t think you need to reassign it multiple times and closure will allow for this variable to be persistent per lesson

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const displayLessons = props.lessons.map((lesson, lessonId) => {
```
michaelbayday, line 49: I think it would be good practice to type the parameters of the callback.

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type Challenge = {
+  challengeNumber: number
+  challengeStatus?: string
+}
+
+export const ChallengeStatus: React.FC<ChallengeStatusProps> = props => {
+  const displayChallengeStatus = props.challengesData.map(
+    (eachChallenge, challengeId) => {
```
michaelbayday, line 25: Can you type the parameters of the callback?

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,78 @@
...
+        challengeStatus = 'unapproved_challenge_status'
+      }
+      if (eachChallenge.challengeStatus === 'pending') {
+        challengeStatus = 'pending_challenge_status'
+      }
+      return (
+        <span
+          key={challengeId}
+          className={`challenge_status ${challengeStatus}`}
+        ></span>
```
michaelbayday, line 40: Self closing tag

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,64 @@
...
+  background-color: #00cf92;
+}
+
+.unapproved_challenge_status{
+  background-color: red;
+}
+
+.pending_challenge_status{
+ background-color: orange;
+}
```
songz, line 15: should be using bootstrap defaults instead of specifying your own colors.

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.lesson_challenges{
+  display: flex;
+  margin-top: 15px
+}
+
+.lessons_challenges_title{
```
songz, line 22: there are a few of these, perhaps making these `.profile_section_title` would be best, then you don't have to repeat `padding: 8px` css over and over again.

components/LessonsChallenges.tsx
rkalra247: > Can you just declare this variable as a const on line 50? I don’t think you need to reassign it multiple times and closure will allow for this variable to be persistent per lesson

that makes sense, thanks for the explanation.

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,47 @@
...
+    margin-top: 40px;
+  }
+  .lesson_title_container{
+    display: block;
+  }
+
+  .challenge_status{
+    height: 7px;
+    width: 40px;
+    padding: 10px;
```
songz, line 22: `padding` is not necessary here, just adjust the `height`

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,47 @@
...
+    margin-left: 2px;
```
songz, line 23: `margin-right` is better. Otherwise your block will have a 2px offset from the words

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,47 @@
...
+    display: inline-block;
+  }
+
+  .challenges_container{
+    margin-left: 50px;
```
songz, line 28: remove

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  .challenges_stats{
+    margin-left: 50px;
```
songz, line 32: remove

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,47 @@
...
+    margin-top: 15px;
+  }
+
+
+  .lesson_titles{
+    margin-left: 50px;
```
songz, line 38: remove

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,47 @@
...
+    margin-top: 10px;
+    color: grey;
+
+  }
+
+  .profile_section_title{
+    padding: 8px;
```
songz, line 45: looks better without it.

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,79 @@
...
+      )
+    }
+  )
+  return <div className="challenges_container">{displayChallengeStatus}</div>
+}
+
+const LessonsChallenges: React.FC<LessonChallengeProps> = ({ lessons }) => {
+  const displayLessons = lessons.map((lesson, lessonId) => {
+    const filterPassedChallenges = lesson.challenges.filter(
+      e => e.challengeStatus === 'passed'
```
Wolfy64, line 52: Quick tip: It's always better to name the element `e ==> lesson`

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,47 @@
+.profile-submissions{
+
+  .lesson_challenges{
+    display: flex;
+    margin-top: 15px
+  }
+
+  .lessons_challenges_card_body{
+    margin-left: 30px;
```
songz, line 9: remove. Use the default, otherwise this will look different from your other component (ProfileLessons)

components/LessonsChallenges.tsx
michaelbayday: @Wolfy64  do you mean rename e to challenge in this case?

Wolfy64: > @Wolfy64 do you mean rename e to challenge in this case?

What are you filter? I would say `lesson` what do you think? 
Anyway but it's always better than just `e` 🙂

scss/lessonsChallenges.scss
```diff
@@ -0,0 +1,34 @@
...
+    display: flex;
+    margin-top: 15px
+  }
+
+ .display_lessons{
+    margin-top: 40px;
+  }
+  .lesson_title_container{
+    margin-left: 50px;
+    display: block;
```
songz, line 13: remove, by default the `div` is already `display block`

components/LessonsChallenges.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const filterPassedChallenges = lesson.challenges.filter(
+      eachChallenge => eachChallenge.challengeStatus === 'passed'
+    )
+    return (
+      <div key={lessonId} className="lesson_challenges">
+        <img
+          src={`/curriculumAssets/lessonCoversSvg/js-${lesson.order}-cover.svg`}
+        />
+        <div className="lesson_title_container">
+          <h6 className="lesson_titles">{lesson.title}</h6>
```
songz, line 60: `lesson_title`

PR #204, opened 2020-05-06
update #161 UI for Profile Image and User Info

rkalra247:


@types/user.d.ts
```diff
@@ -0,0 +1,5 @@
+export type User = {
+  userName: string
```
songz, line 2: `username` is used everywhere, not `userName`

components/ProfileImageInfo.tsx
```diff
@@ -0,0 +1,26 @@
...
+import '../scss/profileImageInfo.scss'
+
+type ProfileImageInfoProps = {
+  user: User
+}
+
+const ProfileImageInfo: React.FC<ProfileImageInfoProps> = props => {
+  return (
+    <div className="card shadow-sm">
+      <div className="profile_image_placeholder text-center">
```
songz, line 13: `profile_image_container`

components/ProfileImageInfo.tsx
```diff
@@ -0,0 +1,26 @@
...
+const ProfileImageInfo: React.FC<ProfileImageInfoProps> = props => {
+  return (
+    <div className="card shadow-sm">
+      <div className="profile_image_container text-center">
+        <UserInfoImage user={props.user} />
+        <h2 className="profile_image_full_name text-center">
+          {`${props.user.firstName} ${props.user.lastName}`}
+        </h2>
+        <h4 className="profile_image_username text-center text-muted">
+          {'@' + props.user.username}
```
Wolfy64, line 19: You use string interpolation above, why not use here and stay consistent
``` `@  ${props.user.username}` ```

File: components/SubmissionCard.tsx [hunk 4]
```diff
@@ -1,23 +1,26 @@
 import React from 'react'
 import Button from './Button'
+import UserInfoImage from './UserInfoImage'
+import { User } from '../@types/user'
+import '../scss/submissionCard.scss'
 
 type Props = {
   userId: string
```
Wolfy64, line 8: I know we did not do on other types but should we not use `userId: ID!` instead of `string`

songz: what is `ID` used for?

Wolfy64: Nothing, I mixed up Type and GraphQL^^

components/ProfileImageInfo.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import UserInfoImage from './UserInfoImage'
+import { User } from '../@types/user'
+import '../scss/profileImageInfo.scss'
+
+type ProfileImageInfoProps = {
+  user: User
+}
+
+const ProfileImageInfo: React.FC<ProfileImageInfoProps> = props => {
```
coltonehrman, line 10: `({ user })` instead of `props`

components/SubmissionCard.tsx
```diff
@@ -1,23 +1,26 @@
...
-  initial: string
   time: string
   challengeTitle: string
   reviewUrl: string
+  user: User
 }
 
 const SubmissionCard: React.FC<Props> = props => {
```
coltonehrman, line 15: `({ user, ... })` instead of `props`

PR #208, opened 2020-05-14
fixes #203 adds functionality to navbar links

### LINKED ISSUE #203: [BUG] - Navbar links when logged in doensn't work

rkalra247:


pages/book.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { useEffect } from 'react'
+
+const Book: React.FC = () => {
+  useEffect(() => {
+    window.location.href =
+      'https://www.notion.so/garagescript/Table-of-Contents-a83980f81560429faca3821a9af8a5e2'
+  }, [])
+  return <></>
```
Wolfy64, line 8: Return `null` instead of `<></>`.
What is the point of this component do you plan to add content for `/book` or it is just a redirection?

Wolfy64:
Could link your PR with the issue like this --> closes #203

pages/book.tsx
rkalra247: > Return `null` instead of `<></>`.
> What is the point of this component do you plan to add content for `/book` or it is just a redirection?

For now just redirection. Which is better approach? Returning null or <></> ?

songz: It don't even matter @Wolfy64 

This route will eventually be deprecated.

hwong0305: @Wolfy64  Does it even matter if he returns `<></>` or `null` ?

rkalra247: but still, what is a better approach? When to use null vs <></> ?

Wolfy64: It do not matter we could even return `false`, `true`, `undefined`, it's just a [common practice](https://stackoverflow.com/a/52223388/8279060).

rkalra247: understood.

PR #214, opened 2020-05-20
fix #211 add Submission Query resolver

### LINKED ISSUE #211: [Backend] - Missing Submission query resolver
Submission Query is not working at the moment.

rkalra247:


graphql/resolvers.test.js
```diff
@@ -3,10 +3,15 @@ import db from '../helpers/dbload'
 
 describe('GraphQL resolvers', () => {
   const { Lesson } = db
+  const { Submission } = db
```
SahilKalra98, line 6: you can do const { Lesson, Submission } = db

graphql/resolvers.ts
```diff
@@ -13,6 +13,9 @@ export default {
         order: [['order', 'ASC']]
       })
     },
+    submissions() {
+      return Submission.findAll()
```
songz, line 17: this will be insanely slow. Our database has over 10000 submissions so far, some of them are really bad (includes huge 20 file changes). Could you accomplish what you want to do, without pulling every submission from the database?

graphql/resolvers.ts
```diff
@@ -13,6 +13,13 @@ export default {
         order: [['order', 'ASC']]
       })
     },
+    submissions() {
+      return Submission.findAll({
+        where: {
+          status: 'open'
```
songz, line 19: you will get a lessonId query.

PR #217, opened 2020-05-22
Refactor PR

hwong0305:
moved tests from `__tests__` to helpers. 
update index to route to curriculum instead of rendering curriculum


pages/index.tsx
```diff
@@ -1,24 +1,32 @@
...
   }
 
-  return (
-    <>
-      <AppNav loggedIn={false} />
-      <LandingPage />
-      <Footer footerType="py-5 bg-white text-muted" />
-    </>
-  )
+  if (session)
```
songz, line 19: do you need this `if` condition? Why not just return directly?

hwong0305: Yes. Otherwise React Testing Library will throw an error for the case where there's no session.

PR #223, opened 2020-05-28
[1 / 3] - Backend Functionality for Reset Password

hwong0305:
This is part of multiple PRs to add reset Password functionality 

- Adds Helper Functions for sending emails
- Adds resolver functions for resetting password, and changing password
- Add changing password functionality on Mattermost

The second PR will add the resolvers to GraphQL server, update Signup and make changes to Front End

Screenshot for Reset Password Email Template
image


helpers/controllers/passwordController.ts
```diff
@@ -0,0 +1,108 @@
...
+) => {
+  const { req } = ctx
+  try {
+    const userOrEmail = _.get(arg, 'userOrEmail', null)
+    if (!userOrEmail) {
+      throw new UserInputError('Please provide username or email')
+    }
+
+    let user: any
+    if (userOrEmail.indexOf('@') !== -1) {
```
songz, line 28: this could be a really useful util function, like `const user = getUserByEmailOrUsername( userOrEmail )`

Util function could do the throwing and input validation, making this `reqPwReset` function much smaller.

songz: If you make this a util function, we could potentially use this in login as well!

helpers/controllers/passwordController.ts
```diff
@@ -0,0 +1,108 @@
...
+      throw new AuthenticationError('User does not exist')
+    }
+    if (token !== user.forgotToken || Date.now() > user.expiration) {
+      throw new AuthenticationError('Invalid Token')
+    }
+    const hash = await bcrypt.hash(password, 10)
+    user.expiration = null
+    user.forgotToken = null
+    user.password = hash
+    await user.save()
```
songz, line 91: It's better to do this after `changeChatPassword`. This prevents our pw from being out of sync with chat.

helpers/formValidation.tsx
```diff
@@ -17,10 +17,6 @@ const signupValidation = Yup.object({
     .max(TEXT_MAX, `Must be ${TEXT_MAX} characters or less`)
     .trim('Leading and trailing value must be alphanumeric character')
     .required('Required'),
-  password: Yup.string()
```
songz, line 20: How does this affect the signup password validation?

songz: ooooo there's no password inputs for signup anymore :)

hwong0305: I meant to move this to a different PR. I don't want to break production on this one.

PR #227, opened 2020-05-31
Closes #226 - Implementation of Contributors Page

### LINKED ISSUE #226: [Launch] - No Current Contributor Page

aortizoj15:
This PR adds a contributors page displaying the contributors to c0d3.com.
Also adds a link to contributors page when no user is signed in.
image
image


scss/contributorCard.scss
```diff
@@ -0,0 +1,4 @@
+.contributorCard_avatarContainer{
```
michaelbayday, line 1: we're doing BEM so use dashes instead

PR #232, opened 2020-06-01
[FEATURE] - View Source Code in Storybook

hwong0305:
Also included small  bug fix with AppNav without external flag


components/AppNav.tsx
```diff
@@ -31,12 +31,14 @@ const AuthLink = () => (
     <NavLink
       path="https://github.com/garageScript/c0d3-app"
       className="nav-item nav-link"
+      external
```
michaelbayday, line 34: nice lol this error was bugging me

hwong0305: Lol. I noticed that it only affects dev but not prod.

PR #233, opened 2020-06-01
Posthog implementation

michaelbayday:
Basic posthog implementation for monitoring web traffic

//To do: look into blocking dev environment api calls, look into monitoring other activities on website


pages/_app.tsx
```diff
@@ -20,6 +20,7 @@ function MyApp({ Component, pageProps, apollo }: IProps) {
         <Head>
           <title>C0D3.com</title>
           <link rel="shortcut icon" href="/favicon.ico" />
+          <script type="text/javascript" src="/posthog.js" />
```
songz, line 23: could we just add it into `script` tag here? You are creating another file / another request by doing this.

PR #238, opened 2020-06-02
Posthog API Key environment variable load

michaelbayday:
New environment variable to store Posthog API Key

POSTHOG_API_KEY


pages/_app.tsx
```diff
@@ -19,7 +19,7 @@ function MyApp({ Component, pageProps, apollo }: IProps) {
   const session = useSession()
   useEffect(() => {
     if (process.env.NODE_ENV === 'production') {
```
songz, line 21: should be.... `if (process.env.NODE_ENV === 'production && process.env.POSTHOG....)`

PR #240, opened 2020-06-02
Review Page boilerplate

michaelbayday:
Review Page boilerplate

Issue: gitdiffparser is not liking some of our diff data

image


components/ChallengeMaterial.tsx
```diff
@@ -139,7 +139,11 @@ export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = ({
 export const ChallengeQuestionCard: React.FC<ChallengeQuestionCardProps> = ({
   currentChallenge
 }) => {
-  const diff = _.get(currentChallenge, 'submission.diff', false)
+  //temp fix while CLI tool is getting fixed
```
songz, line 142: could use a comment on what the regex is filtering out for. In the industry, regexes should always be accompanied by a comment.

graphql/queries.ts
```diff
@@ -99,10 +110,7 @@ export const GET_APP = gql`
         order
         challengeId
         lessonId
-        reviewer {
-          id
-          username
```
songz, line 104: don't you want to view who reviewed your submission? I think reviewer id and username is important.

graphql/queries.ts
```diff
@@ -61,7 +61,18 @@ export const SIGNUP_USER = gql`
...
+      id
       status
+      diff
+      comment
+      challengeId
+      user {
+        id
+        username
+      }
+      reviewerId
```
songz, line 73: same should this be reviewer instead?

helpers/controllers/submissionController.ts
```diff
@@ -12,7 +12,7 @@ type ArgsCreateSubmission = {
   challengeId: string
 }
 
-type Submission = {
+type ArgsGetSubmissions = {
```
songz, line 15: I think a better name: `GetSubmissionsWhere`

helpers/controllers/submissionController.ts
```diff
@@ -23,12 +23,14 @@ export const createSubmission = async (
   try {
     if (!args) throw new Error('Invalid args')
     const { challengeId, cliToken, diff, lessonId } = args
+    //to revert once CLI is fixed
+    const fixedDiff = diff.replace(/(.?\[\d*m)/g, '')
```
songz, line 27: if you do this here, why do it on the front end?

helpers/controllers/submissionController.ts
```diff
@@ -43,12 +45,53 @@ export const createSubmission = async (
...
+  const subs = await Submission.findAll({
     where: {
       status: 'open',
       lessonId
     }
   })
+  if (subs.length === 0) {
+    return subs
+  }
+  const userIds = subs.reduce((acc: Set<number>, sub: any) => {
```
songz, line 63: none of these should be necessary, you just use an `include` field in your submission query.

helpers/controllers/submissionController.ts
michaelbayday: This is for handling new submissions for the interim until CLi is fixed

graphql/queries.ts
michaelbayday: good point haha

songz:
HOT MERGE

PR #243, opened 2020-06-03
[Review Submissions] - Accept/Reject submissions with comments

coltonehrman:


components/ReviewCard.tsx
```diff
@@ -1,18 +1,30 @@
...
   submissionData: SubmissionData
 }
 
-export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
-  const diff = _.get(submissionData, 'diff', '')
+export const ReviewCard: React.FC<ReviewCardProps> = ({
+  session,
+  submissionData
+}) => {
+  const diff = _.get(submissionData, 'diff', '').replace(
```
michaelbayday, line 20: @songz  and I spoke and we should remove the regex and let users fix the submissions themselves. The CLI is fixed so we'll let them reconcile the data themselves

components/ReviewCard.tsx
```diff
@@ -53,6 +65,7 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
       />
     )
   }
+  console.log(session)
```
michaelbayday, line 68: console.log

components/ReviewCard.tsx
```diff
@@ -68,6 +81,44 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
...
+              onClick={async () => {
+                const textField = commentTextField.current as any
+                await accept({
+                  variables: {
+                    reviewer: Number(session.id),
+                    submissionId: submissionData.id,
+                    comment: textField.value
+                  }
+                })
+                window.location.reload()
```
michaelbayday, line 100: do you want to maybe just refetch the data for comments instead of reloading? maybe smoother UI

coltonehrman: well after you either accept/reject a submission it will disappear from the reviews page

hwong0305: Instead of doing this, look at updating the cache with mutation.

graphql/resolvers.ts
```diff
@@ -69,6 +71,22 @@ export default {
...
+      }
+    },
+    async rejectSubmission(_parent: void, args: ArgsUpdateSubmission) {
+      try {
+        if (!args) throw new Error('Invalid args')
+        return await updateSubmission({ ...args, status: 'needMoreWork' })
+      } catch (error) {
+        throw new Error(error)
+      }
+    },
```
hwong0305, line 89: Move to `submissionController`

pages/review/[lesson].tsx
```diff
@@ -14,21 +14,27 @@ import { SubmissionData } from '../../@types/submission'
 import _ from 'lodash'
 
 type SubmissionDisplayProps = {
+  session: any
```
Comment (hwong0305) [inline pages/review/[lesson].tsx:17]: Let's define the type since we already have it instead of using any.

components/ReviewCard.tsx
```diff
@@ -1,18 +1,30 @@
...
+export const ReviewCard: React.FC<ReviewCardProps> = ({
+  session,
+  submissionData
+}) => {
+  const diff = _.get(submissionData, 'diff', '').replace(
+    /(\\u\d*b)?(.?\[\d*m)/g,
+    ''
+  )
   const comment = _.get(submissionData, 'comment', '')
+  const commentTextField = useRef(null)
```
hwong0305, line 25: Why do we need to use ref? I think we can use `useState` and assign an `onChange`

hwong0305: `const [comment, setComment] = useState('')`

components/ReviewCard.tsx
```diff
@@ -68,6 +81,44 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
           </div>
           <div className="card-footer bg-white">
             {comment && <Markdown>{comment}</Markdown>}
+            <textarea
+              ref={commentTextField}
```
hwong0305, line 85: `value={comment} onChange={e => setComment(e.target.value)}`

graphql/resolvers.ts
hwong0305: You should able to if you change the way this is written. Instead of calling a mutation from another mutation. Make a helper function to `updateSubmission`. That way you can just do a unit test for that function itself and then mock that function for the mutations.

components/ReviewCard.tsx
songz: @hwong0305 updating the cache with mutation. @coltonehrman could you figure it out? It'll be the first time its done in our codebase lol.

Also, could we move this function to outside the component? Its nice to keep the components to just components, instead of being cluttered with JS logic.

components/ReviewCard.tsx
```diff
@@ -68,6 +81,44 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
...
+              onClick={async () => {
+                const textField = commentTextField.current as any
+                await reject({
+                  variables: {
+                    reviewer: Number(session.id),
+                    submissionId: submissionData.id,
+                    comment: textField.value
+                  }
+                })
+                window.location.reload()
```
songz, line 117: same as above, could we move this function to outside the component? Its nice to keep the components to just components, instead of being cluttered with JS logic.

graphql/queries.ts
```diff
@@ -76,6 +76,42 @@ export const GET_SUBMISSIONS = gql`
...
+      reviewer: $reviewer
+    ) {
+      id
+      comment
+      status
+    }
+  }
+`
+
+export const REJECT_SUBMISSION = gql`
```
songz, line 97: I think its better to have 1 mutation, `REVIEW_SUBMISSION` with a status `approve / reject`.

coltonehrman: I felt that it made more sense to have one for each specific status that way we don't have to rely on knowing the particular status string for each one.

components/ReviewCard.tsx
coltonehrman: It appears that the cache is actually being updated automagically! lol I must not have noticed

graphql/queries.ts
songz: 👍

graphql/queries.ts
```diff
@@ -76,6 +76,42 @@ export const GET_SUBMISSIONS = gql`
   }
 `
 
+export const ACCEPT_SUBMISSION = gql`
+  mutation acceptSubmission(
+    $submissionId: String!
+    $comment: String!
+    $reviewer: Int!
```
songz, line 83: could `reviewer` just be taken from the user object in the request? This way we don't introduce a new security where a user can pretend to be anyone in their reviews.

pages/review/[lesson].tsx
```diff
@@ -11,24 +11,31 @@ import LoadingSpinner from '../../components/LoadingSpinner'
...
 const SubmissionDisplay: React.FC<SubmissionDisplayProps> = ({
+  user,
   submissions
 }) => (
   <div className="submissions-container">
     {submissions.map((submission: SubmissionData) => (
-      <ReviewCard key={submission.id} submissionData={submission} />
+      <ReviewCard
+        key={submission.id}
+        reviewerId={Number(user.id)}
```
Comment (songz) [inline pages/review/[lesson].tsx:30]: why add this new prop?

graphql/queries.ts
coltonehrman: What do you mean by the user object in the request?

coltonehrman: I'm grabbing it from the resolver ctx now 👍

helpers/controllers/submissionController.ts
```diff
@@ -45,7 +47,37 @@ export const createSubmission = async (
   }
 }
 
-export const submissions = (
+export const acceptSubmission = async (
+  _parent: void,
+  args: ArgsUpdateSubmission,
+  ctx: any
```
hwong0305, line 53: `Context` from `@types/helper`

helpers/controllers/submissionController.ts
```diff
@@ -45,7 +47,37 @@ export const createSubmission = async (
...
+    return updateSubmission({ ...args, reviewerId, status: 'passed' })
+  } catch (error) {
+    throw new Error(error)
+  }
+}
+
+export const rejectSubmission = async (
+  _parent: void,
+  args: ArgsUpdateSubmission,
+  ctx: any
```
hwong0305, line 68: `Context` from `@types/helper`

components/ReviewCard.tsx
```diff
@@ -68,6 +82,25 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
...
             {comment && <Markdown>{comment}</Markdown>}
+            <textarea
+              value={commentValue}
+              onChange={e => setCommentValue(e.target.value)}
+              placeholder="Type something..."
+              style={{ width: '100%', padding: '1rem' }}
+            ></textarea>
+            <button
+              className="btn bg-success m-1 text-white"
+              onClick={reviewSubmission(accept)}
```
songz, line 93: beautiful closure!

PR #261, opened 2020-06-08
fixed incorrectly typed HTML entity (&lsquo;) on forgot password page

anthonykhoa:
HTML entity for left single quote needs a semi-colon at the end, updated code to reflect this need.

### DIFF
--- a/pages/forgotpassword.tsx
+++ b/pages/forgotpassword.tsx
@@ -42,7 +42,7 @@ export const ResetPassword: React.FC = () => {
   return (
     <Card title="Reset your password">
       <p className="mb-5">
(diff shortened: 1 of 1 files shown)


pages/forgotpassword.tsx
```diff
@@ -42,7 +42,7 @@ export const ResetPassword: React.FC = () => {
   return (
     <Card title="Reset your password">
       <p className="mb-5">
-        Type in your email or username below and we`&lsquo`ll send you an email
+        Type in your email or username below and we`&lsquo;`ll send you an email
```
hwong0305, line 45: `we&lsquo;ll`  then run `yarn test -u`

PR #270, opened 2020-06-09
[Theming] - Started theme components

coltonehrman:
- Added Text theme component
- Replaced some elements in ReviewCard with Text
- Added storybook for theme components


components/ReviewCard.tsx
```diff
@@ -67,19 +83,26 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
...
+
   return (
     <>
       {diff && (
-        <div className="card shadow-sm border-0 mt-3 p-3">
+        <div className="card shadow-sm border-0 mt-3">
           <div className="card-header bg-white">
-            <strong>{submissionData.user.username}</strong> submitted Submission
-            ID: {submissionData.id}
+            <Text color="darkgrey" bold>
```
hwong0305, line 92: I know we didn't have it previously, but this seems to be the title of the review card so it should be wrapped in a heading tag.

components/ReviewCard.tsx
```diff
@@ -1,20 +1,35 @@
 import React, { useState } from 'react'
+
```
hwong0305, line 2: ?

components/ReviewCard.tsx
coltonehrman: Right, I didn't want to add too many components off the start. So, for now, I just have the Text component. We can always add additional ones later as needed :)

components/ReviewCard.tsx
coltonehrman: I was trying to group the imports to appear nicer? lol

PR #273, opened 2020-06-09
update #234 Show individual profile page for each user. 

rkalra247:
Can see profile page for signed in user. 

Image 2020-06-10 at 4 52 29 PM


pages/userprofile.tsx
```diff
@@ -0,0 +1,98 @@
...
+          </div>
+        </div>
+      </>
+    </Layout>
+  )
+}
+
+export default withQueryLoader(
+  {
+    query: GET_APP
```
hwong0305, line 95: You can't use session. Otherwise it will only show the user that's signed in.

pages/userprofile.tsx
```diff
@@ -0,0 +1,98 @@
...
+  },
+  UserProfile
+)
```
hwong0305, line 98: filepath should be `/pages/profile/[username].tsx`


then you can get the user by `router.query.username`

pages/userprofile.tsx
```diff
@@ -0,0 +1,98 @@
...
+  status: string
+  lessonId: number
+  challengeId: number
+}
+
+const UserProfile: React.FC<AppQuery> = ({ queryData }) => {
+  const userInfo = {
+    username: queryData.session.user.username,
+    firstName: queryData.session.user.name.split(' ')[0],
+    lastName: queryData.session.user.name.split(' ')[1]
```
hwong0305, line 40: Why do you need to do this if you're just displaying the user's name?

pages/userprofile.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const progress = Math.floor(lessonProgress * 100)
+    const order = lesson.order
+    return {
+      progress,
+      order
+    }
+  })
+
+  const lessons = queryData.lessons.map(lesson => {
+    console.log('lesson:', lesson)
```
hwong0305, line 57: remove `console.log`

pages/userprofile.tsx
```diff
@@ -0,0 +1,98 @@
...
+    username: queryData.session.user.username,
+    firstName: queryData.session.user.name.split(' ')[0],
+    lastName: queryData.session.user.name.split(' ')[1]
+  }
+
+  const lessonInfo = queryData.lessons.map(lesson => {
+    const lessonProgress =
+      queryData.session.submissions.filter(
+        r => r.status === 'passed' && r.lessonId === lesson.id
+      ).length / lesson.challenges.length
```
hwong0305, line 47: Is there anything that's preventing the student from making another submission when it's already approved? If they can, then this logic breaks.

hwong0305: This seems to be logic that is breaking by existing user. From older uses, some challenges may have been deleted causing the challengeId to be `null` breaking this logic.

pages/userprofile.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const order = lesson.order
+    const title = lesson.title
+    const lessonChallenges = queryData.session.submissions.filter(
+      r => r.lessonId === lesson.id
+    )
+    const status = lessonChallenges.map((eachChallenge, challengeOrder) => {
+      return {
+        challengeNumber: challengeOrder,
+        challengeStatus: eachChallenge.status
+      }
```
hwong0305, line 67: You're not going to get a return value for something that doesn't have a submission.

You're going to need to go through this array and find out what is missing. If it's missing you gotta add another item with `{ challengeNumber: missingNumber, challengeStatus: 'open' }`

pages/userprofile.tsx
```diff
@@ -0,0 +1,113 @@
...
+    return {
+      progress,
+      order
+    }
+  })
+
+  const lessons = queryData.lessons.map(lesson => {
+    const order = lesson.order
+    const title = lesson.title
+    const challengesInLesson = lesson.challenges
```
coltonehrman, line 65: const { order, title, challenges: challengesInLesson } = lesson

pages/userprofile.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+type Submission = {
+  diff: string
+  status: string
+  lessonId: number
+  challengeId: number
+}
+
+const UserProfile: React.FC<AppQuery> = ({ queryData }) => {
```
coltonehrman, line 36: ({ queryData: { lessons, session } })

pages/userprofile.tsx
```diff
@@ -0,0 +1,113 @@
...
+  user: SessionUser
+  submissions: Submission[]
+}
+
+type Submission = {
+  diff: string
+  status: string
+  lessonId: number
+  challengeId: number
+}
```
coltonehrman, line 34: I believe some of these types should be defined already under `@types` folder

pages/userprofile.tsx
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  const lessonInfo = queryData.lessons.map(lesson => {
+    const challengesInLesson = lesson.challenges
+    const submissionsInLesson = queryData.session.submissions.filter(
+      eachSubmission =>
+        eachSubmission.status === 'passed' &&
+        eachSubmission.lessonId === lesson.id
+    )
+    const updateSubmissions = submissionsInLesson.filter(eachSubmission => {
```
coltonehrman, line 50: What is this filter doing?

pages/userprofile.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+const UserProfile: React.FC<AppQuery> = ({ queryData }) => {
+  const userInfo = {
+    username: queryData.session.user.username,
+    firstName: queryData.session.user.name.split(' ')[0],
+    lastName: queryData.session.user.name.split(' ')[1]
+  }
+
+  const lessonInfo = queryData.lessons.map(lesson => {
```
coltonehrman, line 43: Just a little revising, can use if you like :)

```
const lessonInfo = lessons.map((lesson: any) => {
  const { challenges, order } = lesson
  const { submissions } = session
  const passedLessonSubmissions = submissions.filter(
    ({ status, lessonId }) =>
      status === 'passed' &&
      lessonId === lesson.id
  )
  const updateSubmissions = passedLessonSubmissions.filter(({ challengeId }) => challengeId)
  const lessonProgress = updateSubmissions.length / challenges.length
  const progress = Math.floor(lessonProgress * 100)

  return { progress, order }
})
```

pages/userprofile.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const lessonProgress = updateSubmissions.length / challengesInLesson.length
+    const progress = Math.floor(lessonProgress * 100)
+    const order = lesson.order
+    return {
+      progress,
+      order
+    }
+  })
+
+  const lessons = queryData.lessons.map(lesson => {
```
coltonehrman, line 62: Just some revisions if you want to use them :)

```
const lessons = lessons.map(({ order, title, challenges }) => {
  const { submissions } = session

  const challengesStatus = challenges.map((c, order) => {
    const challengeSubmission = submissions.find(s => c.id === s.challengeId)

    return (challengeSubmission) ? ({
      challengeNumber: order,
      challengeStatus: 'open'
    }) : ({
      challengeNumber: order,
      challengeStatus: challengeSubmission.status
    })
  })

  return {
    order,
    title,
    challenges: challengesStatus
  }
})
```

pages/userprofile.tsx
hwong0305: It filters out null in `challengeId`

pages/userprofile.tsx
hwong0305: The path should be `/profile/[username].tsx` then. So to avoid moving the file in a future PR.

pages/userprofile.tsx
coltonehrman: Could also do `challengeSubmission?.status || 'open'`

pages/profile/[Username].tsx
```diff
@@ -0,0 +1,87 @@
...
+    username,
+    firstName: fullname.split(' ')[0],
+    lastName: fullname.split(' ')[1]
+  }
+
+  const userSubmissions: UserSubmission[] = _.get(session, 'submissions', [])
+  const lessonInfo = lessons.map((lesson: Lesson) => {
+    const { challenges, order } = lesson
+    const passedLessonSubmissions = userSubmissions.filter(
+      ({ status, lessonId }) => {
```
Comment (songz) [inline pages/profile/[Username].tsx:28]: could you add a comment here:

```
// TODO: Fix lesson.id and lessonId types
```

pages/profile/[Username].tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from 'react'
```
Comment (songz) [inline pages/profile/[Username].tsx:1]: rename file to `[username].tsx`

pages/profile/[Username].tsx
```diff
@@ -0,0 +1,87 @@
...
+    return {
+      order,
+      title,
+      challenges: challengesStatus
+    }
+  })
+
+  return (
+    <Layout>
+      <>
```
Comment (songz) [inline pages/profile/[Username].tsx:67]: `<>` is not necessary.

PR #274, opened 2020-06-10
[FIX BUG] - UserLesson data never getting updated

### LINKED ISSUE #222: [Bug] - Submission Approval backend not completing student lessons
When a user's submission for the last challenge has been approved, user's lesson status does not updated to passed.

1. ~~[Data remediation](https://github.com/garageScript/c0d3-app/wiki/Code-Snippets#data-redemiation-bug) - There are users who have technically passed a lesson, but does not have isPassed set to true in lesson status table. Need to find out who they are and fix them.~~
2. New users - last submission approval should pass their lesson.

When student completes a lesson, mattermost notifies the lesson channel more than 1 time. It should only notify the channel one time. 

For example, If a student just finished lesson 2 Array, it should only notify the JS2 channel once that a student has completed lesson 2 successfully. 

mattermost bug

coltonehrman:
During updateSubmission, make sure to check to see if the user has completed all Challenges within Lesson and update UserLesson data appropriately.

Notify MatterMost channels when user passes the lesson!

Closes #222

### DIFF
--- a/helpers/updateSubmission.ts
+++ b/helpers/updateSubmission.ts
@@ -1,6 +1,7 @@
 import db from './dbload'
+import { getUserByEmail, publicChannelMessage } from './mattermost'
 
-const { Submission, Challenge, UserLesson } = db
+const { Submission, Challenge, User, UserLesson, Lesson } = db
 
 export type ArgsUpdateSubmission = {
   id: number
@@ -36,12 +37,36 @@ export const updateSubmission = async (
     )
 
     if (challengeCount === passedSubmissions) {
-      const lesson = await UserLesson.findOne({
+      const userLesson = await UserLesson.findOne({
         where: { lessonId: submission.lessonId }
       })
-      lesson.set('isPassed', true)
-      lesson.set('isTeaching', true)
-      await lesson.save()
+
+      if (!userLesson.isPassed) {
+        const { email } = await User.findByPk(submission.userId)
+        const { order, title, chatUrl } = await Lesson.findByPk(
+          userLesson.lessonId
+        )
+        const chatUsername = await getUserByEmail(email)
+        const nextLesson = await Lesson.findOne({
+          where: { order: order + 1 }
+        })
+
+        const lessonChannel = chatUrl.split('/').pop()
+        const nextLessonChannel = nextLesson.chatUrl.split('/').pop()
+
+        await publicChannelMessage(
+          lessonChannel,
+          `Congratulations to @${chatUsername} for passing and completing ${title}! @${chatUsername} is now a guardian angel for the students in this channel.`
+        )
+        await publicChannelMessage(
+          nextLessonChannel,
+          `We have a new student joining us! @${chatUsername} just completed ${title}.`
+        )
+      }
+
+      userLesson.set('isPassed', true)
+      userLesson.set('isTeaching', true)
+      await userLesson.save()
     }
 
     return submission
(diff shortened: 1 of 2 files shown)


helpers/updateSubmission.ts
```diff
@@ -14,12 +15,60 @@ export const updateSubmission = async (
...
+      })
+
+      if (!userLesson.isPassed) {
+        const { email } = await User.findByPk(submission.userId)
+        const { order, title, chatUrl } = await Lesson.findByPk(
+          userLesson.lessonId
+        )
+        const chatUsername = await getUserByEmail(email)
+        const nextLesson = await Lesson.findOne({
+          where: { order: order + 1 }
```
songz, line 51: should be `previousLesson` :)

If I'm in JS0, my submissions goes to JS1. so... when my last JS0 submission is approved, current channel (JS1) will be updated that I'm joining. And then previous lesson (JS0) should be notified that I passed. 

Add a comment for this too!

coltonehrman: But the submission that's being updated belongs to the `currentLesson` and the `nextLesson` would be the channel that the user is joining right?

If I'm currently in JS0 and my last submission for JS0 get's approved, my currentLesson is still JS0 at that point in time, and my next lesson would be JS1.

songz: `currentLesson.chatUrl` is actually the chatroom for the next lesson.

helpers/updateSubmission.ts
```diff
@@ -14,12 +15,60 @@ export const updateSubmission = async (
...
+      }
+    })
+
+    const passedSubmissions = submissions.reduce(
+      (sum: number, s: any) => sum + (s.status === 'passed' ? 1 : 0),
+      0
+    )
+
+    if (challengeCount === passedSubmissions) {
+      const userLesson = await UserLesson.findOne({
```
songz, line 40: this doesn't always exist, may throw an error.

helpers/updateSubmission.ts
```diff
@@ -14,12 +15,60 @@ export const updateSubmission = async (
...
+        })
+
+        const lessonChannel = chatUrl.split('/').pop()
+        const nextLessonChannel = nextLesson.chatUrl.split('/').pop()
+
+        await publicChannelMessage(
+          lessonChannel,
+          `Congratulations to @${chatUsername} for passing and completing ${title}! @${chatUsername} is now a guardian angel for the students in this channel.`
+        )
+        await publicChannelMessage(
```
songz, line 61: would it make sense to `Promise.all` these 2 functions?

helpers/updateSubmission.ts
```diff
@@ -13,13 +14,87 @@ export const updateSubmission = async (
...
+        userId: submission.userId
+      }
+    })
+
+    // count how many submissions user passed in total
+    const passedLessonSubmissions = lessonSubmissions.reduce(
+      (sum: number, s: any) => sum + (s.status === 'passed' ? 1 : 0),
+      0
+    )
+
```
songz, line 43:
```
// Done if user did not pass the lesson.
if (lessonChallengeCount !== passedLessonSubmissions) {
  return submisson
}

const [userLesson] = await UserLesson.findOrCreate(....)

// User is resubmitting a challenge for a lesson he has already passed. Do not proceed
if (userLesson.isPassed) {
  return submisson
}

const [{email}, currentLesson] = await Promise.all(....)
...
```

Issue #275, opened 2020-06-11
[Refactor] Decouple UI Components from Data


danielthalhuber:
Remove data queries from UI components in Storybook. Consider `AppNav` and `LessonCard`/`ReviewCount`. Each relies on a GraphQL query and uses a different approach to dealing with this situation in Storybook:

- `AppNav` comments out the story that would require a query
- `LessonCard`/`ReviewCount` introduces a prop to avoid using a mock for the client provider

For UI component stories, we should either decouple the components from their queries, or decide on a strategy for mocking/using the Apollo client Provider. I recommend that we adopt the decoupling strategy for the sake of simplicity.

More detail in [this wiki article](https://github.com/garageScript/c0d3-app/wiki/UI-Components-with-GraphQL-Queries)

songz:
not an issue anymore. will create another ticket if this is not the case

PR #276, opened 2020-06-11
Resolves #242: add space between words in button text

### LINKED ISSUE #242: Bug Fix: Review Submissions Button needs a space when there are no submissions to review for a lesson
Bug Fix: Review Submissions needs a space when there are no submissions to review for a lesson

image

danielthalhuber:
This PR addresses issue #242 where the text rendered in the button on a lesson card would be displayed without spacing between words.

##### Issue

During the process of researching a fix for issue #242 , it became apparent that the prop `shouldNotGetCount` was only being used to allow `LessonCard` component stories to render. The `LessonCard` component has a sub component `ReviewCount` that queries data in order to render. The `shouldNotGetCount` prop is used as a flag in the `ReviewCount` to skip the query and render static content for Storybook (the `ApolloProvider` is not currently made available in Storybook).

##### Resolution

Since the prop had no other purpose in the application, and since it's purpose was undocumented and not intuitive, this PR replaces it with a mock provider. With the `shouldNotGetCount` back-door removed, the `LessonCard` components are a little easier to read and understand.

##### Issue

The `ReviewCount` component returns

- A `span` containing a number surrounded by spaces when the query returns data
- `null` (without a `span` or spacing) when the `shouldNotGetCount` prop is `true` or the data returned by the query is falsey
- A spinner when the query is loading

The `ReviewButton` renders a button with the value returned by `ReviewCount` sandwiched between two words, with no spacing. This means that the the `ReviewButton` renders a button with two words that have no space between them when `ReviewCount` returns `null` (as described in issue #242).

##### Resolution

This PR removes the variability in spacing rendered by `ReviewCount`

- `ReviewCount` return values are always:

  - JSX with inner values that contain no surrounding spaces
  - either a `span` containing a number, or a div with a spinner (loading)

- Spacing is added around the rendered `ReviewCount` component in `ReviewButton`

#### Unresolved Issues

This PR does not address the fact that the application has data queries in UI/presentational components (see issue #275 for more detail).


components/LessonCard.tsx
```diff
@@ -26,7 +22,7 @@ const ReviewCount: React.FC<ReviewCountProps> = props => {
   }
 
   if (!data) {
-    return null
+    return <span> </span>
```
coltonehrman, line 25: What's the purpose of returning `<span></span>` vs `null`?

stories/components/LessonCard.stories.tsx
```diff
@@ -6,6 +6,13 @@ export default {
   title: 'Components/LessonCard'
 }
 
+export const Dummy: React.FC = () => <div>Dummy Lesson Card</div>
```
coltonehrman, line 9: I didn't dig too much into the dependencies of LessonCard, but have you explored trying to mock `useQuery`?

coltonehrman: At least for the time being? I understand the end goal is to decouple them, but for now, we might as well try to mock the dependencies

components/LessonCard.tsx
danielthalhuber: Since the `ReviewCount` component represents a text value (rendered in a `span`), and its rendered in the parent component `ReviewButton` between two words (with no additional, surrounding spaces), when it returns `null` the two words have no space between them (leading to the condition cited in the original issue).

There are other ways to accomplish this. I might write this very differently if I was starting from scratch. An HOC that handles the query and passes relevant props down to `LessonCard` might be the way to go.

coltonehrman: I see. Maybe we can just add a comment to summarize why it's a span, just to make sure someone doesn't accidentally change it back to null lol

coltonehrman: But it appears that the query might need to be modified to return something other than null so we have a clearer representation of what's happening.

danielthalhuber: I appreciate your feedback. We have some bigger changes to make at some point in the future, but in the mean-time returning an empty `span` isn't much of an improvement over `null`.

This component is called `ReviewCount`, so aside from the spinner (let's not go there for now 😀), this component should return a review count. I'll update it to return `0` in the odd/rare case where the query data is `undefined`. That should bring a sliver of consistency to the component.

Please take a look after the next commit is applied and let me know if you think a clarifying comment is still in order.

stories/components/LessonCard.stories.tsx
danielthalhuber: After unraveling the mystery of the `shouldNotGetCount` prop, I wasn't feeling particularly motivated to do more work to add a space between the two words on the button.

But you make a good point here. I have a working mock implementation. I'll add it to the PR in the next commit.

There's at least one rough edge that I'm aware of. The mock data is a little distracting in the story file. I feel like it should be hidden somewhere in a distant sub-directory.

Please take another look after the next commit and let me know what you think.

danielthalhuber:
> How does this PR relate to the name `add space between words in button text`?

I was asking myself the same thing yesterday 🤣

PR #290, opened 2020-06-18
Part 1: update #234 - Add typedef and query skeleton

songz:
Add typedef and resolver


__tests__/pages/index.test.js
```diff
@@ -16,7 +16,7 @@ describe('<IndexPage />', () => {
     jest.clearAllMocks()
   })
 
-  const mocksWithSession = (session) => [
```
hwong0305, line 19: That's weird... why didn't eslint catch this?

PR #293, opened 2020-06-21
Add typedef and resolver for userInfo

rkalra247:


graphql/resolvers.test.js
```diff
@@ -129,3 +129,57 @@ describe('Alerts resolvers', () => {
     ])
   })
 })
+
+describe('userInfo resolver', () => {
+  test('should return null if no username is passed in', async () => {
```
coltonehrman, line 134: We may want this to throw an error, I believe that is the behavior of other resolvers when the args does not match with what is expected

graphql/resolvers.test.js
```diff
@@ -1,7 +1,7 @@
-jest.mock('node-fetch')
-jest.mock('mailgun-js')
 import resolvers from '../graphql/resolvers'
 import db from '../helpers/dbload'
+jest.mock('node-fetch')
+jest.mock('mailgun-js')
```
coltonehrman, line 5: `jest.mock('../helpers/dbload')`

This will auto-mock every import from this folder, save a few lines :)

graphql/resolvers.test.js
```diff
@@ -129,3 +129,57 @@ describe('Alerts resolvers', () => {
...
+
+describe('userInfo resolver', () => {
+  test('should return null if no username is passed in', async () => {
+    const returnValue = await resolvers.Query.userInfo({}, {})
+    expect(returnValue).toEqual(null)
+  })
+  test('should return null if User.findOne returns null', async () => {
+    User.findOne = jest.fn()
+    const returnValue = await resolvers.Query.userInfo(
+      {},
```
coltonehrman, line 141: null

graphql/resolvers.test.js
coltonehrman: Just remember to use `jest.resetAllMocks` in `beforeEach` that way specific mocks don't carry over to other tests

graphql/resolvers.test.js
```diff
@@ -129,3 +129,57 @@ describe('Alerts resolvers', () => {
...
+      .fn()
+      .mockReturnValue([{ id: '1', status: 'passed' }])
+    Submission.findAll = jest.fn().mockReturnValue([
+      {
+        id: '1',
+        diff: 'testing2020'
+      }
+    ])
+    const returnValue = await resolvers.Query.userInfo(
+      {},
```
coltonehrman, line 162: null

graphql/resolvers.ts
```diff
@@ -65,6 +65,37 @@ export default {
     },
     alerts() {
       return Alert.findAll()
+    },
+    async userInfo(_parent: void, args: any) {
```
coltonehrman, line 69: Can we type the `args`?

helpers/controllers/userInfoController.test.js
```diff
@@ -0,0 +1,57 @@
+jest.mock('../dbload')
+jest.mock('node-fetch')
+import db from '../dbload'
+import fetch from 'node-fetch'
+import resolvers from '../../graphql/resolvers'
+
+const { User, UserLesson, Submission } = db
+describe('userInfo resolver', () => {
+  test('should return null if no username is passed in', async () => {
```
coltonehrman, line 9: update test description

graphql/resolvers.test.js
coltonehrman: sorry, this looks like the wrong file for this lol
and i think you need to add back the 
```
jest.mock('node-fetch')	
jest.mock('mailgun-js')
```

coltonehrman: the above comments are for the tests that you created for this PR

PR #298, opened 2020-06-22
update #234 add query for userInfo

rkalra247:


graphql/queries/userInfo.ts
```diff
@@ -0,0 +1,54 @@
+import { gql } from 'apollo-boost'
+
+const USER_INFO = gql`
+  query USER_INFO($username: String!) {
```
coltonehrman, line 4: let's try to stick with camelCase for the names. this name will be used by the generator when creating the types and helpers

`query userInfo { ... }`

coltonehrman:
Did you run `yarn generate` to create the new types?

pages/profile/[username].tsx
```diff
@@ -1,35 +1,37 @@
 import * as React from 'react'
+import { useQuery } from '@apollo/react-hooks'
 import _ from 'lodash'
 import Layout from '../../components/Layout'
+import { useRouter } from 'next/router'
 import { Lesson } from '../../@types/lesson'
-import { AppData } from '../../@types/app'
 import { UserSubmission, Challenge } from '../../@types/challenge'
-import GET_APP from '../../graphql/queries/getApp'
+import USER_INFO from '../../graphql/queries/userInfo'
```
Comment (coltonehrman) [inline pages/profile/[username].tsx:8]: We should be importing from `graphql/index.tsx` only now, and using the generated types.

pages/profile/[username].tsx
```diff
@@ -1,35 +1,37 @@
...
 
-const UserProfile: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
-  const { lessons, session } = queryData
-  const fullname = _.get(session, 'user.name', '')
-  const username = _.get(session, 'user.username', '')
-
+const UserProfile: React.FC = () => {
+  const router = useRouter()
+  const username = router.query.username as string
+  const { loading, error, data } = useQuery(USER_INFO, {
```
Comment (coltonehrman) [inline pages/profile/[username].tsx:22]: use the custom `use` query that is generated by code-gen

pages/profile/[username].tsx
```diff
@@ -86,9 +87,4 @@ const UserProfile: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
   )
 }
 
-export default withQueryLoader<AppData>(
-  {
-    query: GET_APP
-  },
-  UserProfile
-)
+export default UserProfile
```
Comment (coltonehrman) [inline pages/profile/[username].tsx:90]: You could also use the custom `with` HOC from the generator, instead of the `use` Query one.

Comment (coltonehrman) [reply pages/profile/[username].tsx:90]: My rule of thumb is if the component depends on a single data set, and that data set is tightly integrated with the component, go with HOC, otherwise use hooks

pages/profile/[username].tsx
Comment (coltonehrman) [reply pages/profile/[username].tsx:8]: Reference https://github.com/garageScript/c0d3-app/blob/4c3f4c0c6d89010fe708021d24f66b5c507f2df4/pages/index.tsx#L6

pages/profile/[username].tsx
Comment (coltonehrman) [reply pages/profile/[username].tsx:90]: Reference:
https://github.com/garageScript/c0d3-app/blob/4c3f4c0c6d89010fe708021d24f66b5c507f2df4/pages/index.tsx#L28

pages/profile/[username].tsx
Comment (coltonehrman) [reply pages/profile/[username].tsx:22]: References:
https://github.com/garageScript/c0d3-app/blob/4c3f4c0c6d89010fe708021d24f66b5c507f2df4/components/AppNav.tsx#L4
https://github.com/garageScript/c0d3-app/blob/4c3f4c0c6d89010fe708021d24f66b5c507f2df4/components/AppNav.tsx#L54

pages/profile/[username].tsx
Comment (rkalra247) [reply pages/profile/[username].tsx:90]: > My rule of thumb is if the component depends on a single data set, and that data set is tightly integrated with the component, go with HOC, otherwise use hooks

I have multiple components being powered through one query on the user profile page, so you recommend i should use hooks?

Comment (coltonehrman) [reply pages/profile/[username].tsx:90]: Don't look at how the components interact with one another, that only concerns the usage of `props` and `contexts`, in this case, we are only concerned with a single component `UserProfile`. The question is, what data does this component depend on, and how tightly coupled is the component with that data. In the case `UserProfile` depends on the single dataset `userInfo` aka `USER_INFO` right, and that data is very tightly coupled with the component itself, it wouldn't make sense to have a UserProfile without userInfo data. So, I would use an HOC

Comment (coltonehrman) [reply pages/profile/[username].tsx:90]: I just realized you will have to use a hook for this case in particular because you need to pass a variable to the query, but the variable will not be known until the component is being rendered, so... you will have to use hooks.

Comment (coltonehrman) [reply pages/profile/[username].tsx:90]: References:

https://github.com/garageScript/c0d3-app/blob/4c3f4c0c6d89010fe708021d24f66b5c507f2df4/components/AppNav.tsx#L4
https://github.com/garageScript/c0d3-app/blob/4c3f4c0c6d89010fe708021d24f66b5c507f2df4/components/AppNav.tsx#L54

```
const [action, { data }] = useTypedMutation({
  variables: {
    ...
  }
})
```

Comment (coltonehrman) [reply pages/profile/[username].tsx:90]: Actually, I just had another thought, you might be able to double wrap with 2 HOCs to make it work, a `withRouter` HOC from next.js to provide the variable you need, and then the `withTypedQuery` HOC from graphql/index.tsx. Up to you though, these are only suggestions :)

pages/profile/[username].tsx
```diff
@@ -1,37 +1,45 @@
...
+  const router = useRouter()
+  const username = router.query.username as string
+  const { loading, error, data } = useUserInfoQuery({
+    variables: { username }
+  })
+  if (loading) return <h1>Loading</h1>
+  if (error) return <h1>Error</h1>
+  const { lessons } = data || {}
+  const lessonsList = lessons || []
+  console.log('data:', data)
```
Comment (coltonehrman) [inline pages/profile/[username].tsx:27]: console.log

pages/profile/[username].tsx
```diff
@@ -1,37 +1,45 @@
...
-  const fullname = _.get(session, 'user.name', '')
-  const username = _.get(session, 'user.username', '')
-
+const UserProfile: React.FC = () => {
+  const router = useRouter()
+  const username = router.query.username as string
+  const { loading, error, data } = useUserInfoQuery({
+    variables: { username }
+  })
+  if (loading) return <h1>Loading</h1>
```
Comment (coltonehrman) [inline pages/profile/[username].tsx:23]: Should we return the `<LoadingSpinner />` here?

pages/profile/[username].tsx
```diff
@@ -1,37 +1,50 @@
...
+  }
+  const { lessons } = data || {}
+  const lessonsList = lessons || []
 
+  const fullname = _.get(data, 'userInfo.user.name', '')
   const userInfo: UserInfo = {
     username,
-    firstName: fullname.split(' ')[0],
-    lastName: fullname.split(' ')[1]
+    firstName: fullname.split(' ')[0] || 'A',
```
Comment (hwong0305) [inline pages/profile/[username].tsx:36]: I think this should default to `anon` instead

pages/profile/[username].tsx
```diff
@@ -44,29 +57,31 @@ const UserProfile: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
     const updateSubmissions = passedLessonSubmissions.filter(
       ({ challengeId }) => challengeId
     )
-    const lessonProgress = updateSubmissions.length / challenges.length
+    const lessonProgress = updateSubmissions.length / challengeList.length
```
Comment (hwong0305) [inline pages/profile/[username].tsx:60]: `updateSubmissions` is a bit misleading. `completedSubmissions` would be better

Issue #301, opened 2020-06-29
[Component] - Create a new component MdInput


songz:
As a reviewer, I want to get a preview of what my markdown looks like.

![Screen Shot 2020-06-28 at 5 10 29 PM](https://user-images.githubusercontent.com/686933/85961893-72ab8b80-b962-11ea-80c2-e1dc8a0c4129.png)

PR #303, opened 2020-07-04
open #301 creation of a MdInput component to preview comments when reviewing submissions

anthonykhoa:
This PR is for the creation of a `MdInput` component to preview comments when reviewing submissions.

### DIFF
--- a/components/MdInput.test.js
+++ b/components/MdInput.test.js
@@ -21,15 +21,11 @@ describe('write inside textarea', () => {
 describe('click Write Button', () => {
   test('switches to write mode when user clicks Write button', () => {
     const { container } = render(<MdInput />)
-    fireEvent.change(screen.getByRole('textbox'), {
-      target: { value: 'Javascript' }
-    })
     fireEvent.click(screen.getByText('Write'))
     const textbox = queryByTestId(container, 'textbox')
     const markdown = queryByTestId(container, 'markdown')
     expect(markdown).toBeNull()
     expect(textbox).toBeTruthy()
-    expect(screen.getByRole('textbox').value).toContain('Javascript')
     expect(container).toMatchSnapshot()
   })
 })
--- a/components/MdInput.test.js
+++ b/components/MdInput.test.js
@@ -1,0 +1,47 @@
+import React from 'react'
+import {
+  render,
+  screen,
+  fireEvent,
+  queryByTestId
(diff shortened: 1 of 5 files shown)


components/MdInput.test.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import {
+  render,
+  screen,
+  fireEvent,
+  queryByTestId
+} from '@testing-library/react'
+import { MdInput } from './MdInput'
+
+describe('write inside textarea', () => {
```
coltonehrman, line 10: Typical test description following the pattern of
```
describe([Component you are testing])
  test([It/Should] [not] do something when X or Y...)
```

coltonehrman: Very nice and concise test cases though :)

components/MdInput.test.js
```diff
@@ -0,0 +1,43 @@
...
+    expect(screen.getByRole('textbox').value).toContain('Javascript')
+    expect(container).toMatchSnapshot()
+  })
+})
+
+describe('click Write Button', () => {
+  test('switches to write mode when user clicks Write button', () => {
+    const { container } = render(<MdInput />)
+    fireEvent.click(screen.getByText('Write'))
+    const textbox = queryByTestId(container, 'textbox')
```
coltonehrman, line 25: I believe you can swap out the global method `queryByTestId` with a method that is returned from `render(<MdInput>)`

components/MdInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+      onChange={handleChange}
+      placeholder="Type something..."
+      style={style}
+      data-testid="textbox"
+    />
+  )
+
+  return (
+    <div style={{ backgroundColor: bgColor }}>
+      <Button onClick={() => preview && setPreview(false)} data-testid="write">
```
coltonehrman, line 53: Do you need the `preview && part`? Does it not work with just `setPreview(false)`?

components/MdInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+        Write
+      </Button>
+      <Button
+        onClick={() => !preview && setPreview(true)}
```
coltonehrman, line 57: Do you need the `!preview &&` part? Does it not work with just `setPreview(true)`?

Issue #291, opened 2020-06-18
[GraphQL HOC] replace withQueryLoader with Types from generator


coltonehrman:
Depends on #292 
Part of #288 

# Problem
Replace all occurrences of `withQueryLoader` to use `graphql` HOC instead.
*Reference - https://www.apollographql.com/docs/react/development-testing/static-typing/#typing-higher-order-components*

- [x]  Replace `withQueryLoader`
  - https://github.com/garageScript/c0d3-app/blob/077bed585af9e5124de2524da9486d476888e5e8/containers/withQueryLoader.tsx#L15-L28

- [x] Replace inside index.tsx page
  - https://github.com/garageScript/c0d3-app/blob/077bed585af9e5124de2524da9486d476888e5e8/pages/index.tsx#L29-L34

- [x] Replace inside curriculum.tsx page
  - https://github.com/garageScript/c0d3-app/blob/077bed585af9e5124de2524da9486d476888e5e8/pages/curriculum.tsx#L92-L97

- [ ] Replace inside curriculum/[lesson].tsx page
  - https://github.com/garageScript/c0d3-app/blob/077bed585af9e5124de2524da9486d476888e5e8/pages/curriculum/%5Blesson%5D.tsx#L58-L63

- [ ] Replace inside review/[lesson].tsx page
  - https://github.com/garageScript/c0d3-app/blob/077bed585af9e5124de2524da9486d476888e5e8/pages/review/%5Blesson%5D.tsx#L69-L74

- [ ] Replace inside profile/[username].tsx page
  - https://github.com/garageScript/c0d3-app/blob/077bed585af9e5124de2524da9486d476888e5e8/pages/profile/%5Busername%5D.tsx#L89-L94

- [x] Replace inside AppNav.tsx component
  - https://github.com/garageScript/c0d3-app/blob/077bed585af9e5124de2524da9486d476888e5e8/components/AppNav.tsx#L152-L157

flacial:
Closing because there isn't much info

PR #310, opened 2020-07-09
[Refactor] Migrate Curriculum page to new GraphQL types

coltonehrman:
Switch over old custom GraphQL HOC with new generated typed HOC in curriculum page component.

Works on #291

### DIFF
--- a/components/Alert.tsx
+++ b/components/Alert.tsx
@@ -1,10 +1,13 @@
 import React from 'react'
 import NavLink from './NavLink'
-import { AlertData } from '../@types/alerts'
+import { GetAppQuery } from '../graphql/'
+import _ from 'lodash'
 import '../scss/alerts.scss'
 
+const alerts: GetAppQuery['alerts'] = []
+
 type Props = {
-  alert: AlertData
+  alert?: typeof alerts[0]
   text?: string
   instructionsUrl?: string
   icon?: string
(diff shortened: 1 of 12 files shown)


pages/curriculum.tsx
```diff
@@ -4,76 +4,75 @@ import LessonCard from '../components/LessonCard'
...
+  const lessonStatusMap: { [id: string]: typeof lessonStatus[0] } = {}
+  for (const status of lessonStatus) {
+    const lessonId = _.get(status, 'lessonId', '-1') as string
+    lessonStatusMap[lessonId] = status
+  }
 
-  const lessonInProgressIdx =
-    lessonsWithStatus.findIndex(lesson => !lesson.lessonStatus.isPassed) || 0
+  const lessonInProgressIdx = _.cond([
+    [_.isEqualWith.bind(null, -1), _.constant(0)],
```
danielthalhuber, line 34: I'm learning some new `lodash` methods today :)

In this case, when we have no `customizer`, what advantage does `isEqualWith` provide over `isEqual`?

coltonehrman: Very good point! haha I'm was learning these as I went as well lol

components/Alert.tsx
```diff
@@ -1,10 +1,13 @@
 import React from 'react'
 import NavLink from './NavLink'
-import { AlertData } from '../@types/alerts'
+import { GetAppQuery } from '../graphql/'
+import _ from 'lodash'
 import '../scss/alerts.scss'
 
+const alerts: GetAppQuery['alerts'] = []
```
songz, line 7:
```
const alertsArrType = 

typeof alerts[0] // TypeScript: Even though alerts is empty array, alerts[0] is still `Alert` type
```

PR #321, opened 2020-07-15
closes #320 Update alerts to check for admin rights

### LINKED ISSUE #320: Check for admin-rights for Alert graphql mutation requests
`addAlert` and `removeAlert` currently do not check if a person is a user, before removing or adding an alert. This is a security risk, because someone with malicious intent can continuously send alerts as much as they want.

anthonykhoa:
Our resolvers for Graphql mutations `addAlert` and `removeAlert` do not check if the user is an admin or not. This is a security flaw because  non-admins can attack the site by sending out alerts continuously

This PR aims to solve this issue by checking to make sure the user is an admin before adding or removing any alerts. A helper function `isAdmin` is also created.

### DIFF
--- a/helpers/controllers/alertController.ts
+++ b/helpers/controllers/alertController.ts
@@ -1,6 +1,7 @@
 import db from '../dbload'
 import { LoggedRequest } from '../../@types/helpers'
 
+import _ from 'lodash'
 const { Alert } = db
 
 type AlertData = {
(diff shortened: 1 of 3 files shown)


helpers/controllers/alertController.ts
```diff
@@ -16,7 +16,11 @@ export const addAlert = async (
   ctx: { req: LoggedRequest }
 ) => {
   const { req } = ctx
+  const isAdmin = _.get(req, 'user.isAdmin', false)
```
songz, line 19: could you create a helper function for this? In the future we are going to change away from strings, so if you create a helper function then in the future when we change to `boolean` we only need to change in 1 place instead of everywhere.

```
if (!isAdmin(req)) {
  throw new Error('User is not admin')
}
```

And your `isAdmin` function should check for both string and boolean at the moment:

```
const isAdmin = (req) => {
  return _.get(req, 'user.isAdmin', false) === 'true'
}
```

PR #325, opened 2020-07-16
Signup Schema: Require Lowercase Alpha-Characters for Email Field (Resolves #244)

### LINKED ISSUE #244: [Bug] Users who upper case their email address will experience an error
Sentry.io alerted an error in a downsteam service.

Solution: Lowercase username / email fields during login / signup

danielthalhuber:
#### Issue

- sentry.io reports an error in a downstream service when email address and usernames contains uppercase alpha-characters

#### Resolution

- Update signup form schema to require lowercase alpha-characters for email field

#### Notes

- Resolves #244
- Schemas for signup and login already require lowercase alpha-characters

### DIFF
--- a/helpers/formValidation.tsx
+++ b/helpers/formValidation.tsx
@@ -7,6 +7,8 @@ const REGEX_ALPHANUMERICS_AND_SPACE = /^[a-zA-Z0-9_\s]*$/
 
 const signupValidation = Yup.object({
   email: Yup.string()
+    .strict(true)
+    .lowercase('Alpha characters must be lowercase')
     .email('Invalid email address')
     .required('Required'),
   username: Yup.string()


songz:
rather than validate, could we auto convert them automatically on the graphql resolver flow when username / email is received?

danielthalhuber:
> rather than validate, could we auto convert them automatically on the graphql resolver flow when username / email is received? This way the users don't have to retype, they can just type in their email / username the way they are used to doing

Definitely. The advantage would be that we reduce the risk of triggering unnecessary discussions/debates about the correctness of casing in the local part of the email. The downside would be minimal: the chance that a user's email server/service supports case sensitivity is low.

Look forward to an exciting new revision/commit!

songz:
closing. Resolution:

```
rather than validate, we auto convert them automatically on the graphql resolver flow when username / email is received. This way the users don't have to retype, they can just type in their email / username the way they are used to doing
```

PR #328, opened 2020-07-17
closes #327 - Creation of graphql resolvers needed for admin page

### LINKED ISSUE #327: GraphQL resolvers for the Admin Page
There are various graphql resolvers that need to be made for the back-end of the Admin Page. 

These are:
1) Query to retrieve all users from the database
2) Mutation to make a user an admin
3) Mutations to create/update a lesson
4) Mutations to create/update a challenge

anthonykhoa:
There are various graphql resolvers that need to be made for the back-end of the Admin Page.

These are:

1) Query to retrieve all users from the database
2) Mutation to make a user an admin
3) Mutations to create/update a lesson
4) Mutations to create/update a challenge

This PR creates all of the listed resolvers.

# Example of new lessons:
Screen Shot 2020-07-16 at 2 32 49 PM
 
# Example of new challenges:
Screen Shot 2020-07-16 at 6 54 43 PM

### DIFF
--- a/graphql/queryResolvers/allUsers.test.js
+++ b/graphql/queryResolvers/allUsers.test.js
@@ -1,0 +1,32 @@
+import { allUsers } from './allUsers'
+import db from '../../helpers/dbload'
+
+const mockUsers = [
+  {
+    username: 'lolcakes',
+    id: 2
+  },
+  {
+    username: 'superLolcakes',
+    id: 6
+  }
+]
+const ctx = {
+  req: {
+    user: { isAdmin: 'true' }
+  }
+}
+
+describe('allUsers resolver', () => {
+  const { User } = db
+
+  test('should return list of users', async () => {
+    User.findAll = jest.fn().mockReturnValue(mockUsers)
+    expect(allUsers(null, null, ctx)).toEqual(mockUsers)
+  })
+
+  test('Should throw Error when user is not an admin when querying allUsers', () => {
+    ctx.req.user.isAdmin = 'false'
+    expect(allUsers(null, null, ctx)).toBeNull
+  })
+})
--- a/graphql/queryResolvers/allUsers.test.js
+++ b/graphql/queryResolvers/allUsers.test.js
@@ -11,6 +11,7 @@ const mockUsers = [
     id: 6
   }
 ]
+
 const ctx = {
   req: {
     user: { isAdmin: 'true' }
--- a/graphql/queryResolvers/allUsers.test.js
+++ b/graphql/queryResolvers/allUsers.test.js
@@ -21,12 +21,12 @@ const ctx = {
 describe('allUsers resolver', () => {
   const { User } = db
 
-  test('should return list of users', async () => {
+  test('Should return list of users', async () => {
     User.findAll = jest.fn().mockReturnValue(mockUsers)
     expect(allUsers(null, null, ctx)).toEqual(mockUsers)
   })
 
-  test('Should throw Error when user is not an admin when querying allUsers', () => {
+  test('Should return null when user is not an admin when querying allUsers', () => {
     ctx.req.user.isAdmin = 'false'
     expect(allUsers(null, null, ctx)).toBeNull
   })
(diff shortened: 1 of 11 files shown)


graphql/typeDefs.ts
```diff
@@ -42,6 +44,40 @@ export default gql`
...
     acceptSubmission(id: String!, comment: String!): Submission
     rejectSubmission(id: String!, comment: String!): Submission
+    createLesson(
+      description: String!
+      docUrl: String
+      githubUrl: String
+      videoUrl: String
+      title: String!
+      chatUrl: String
+      id: Int!
```
songz, line 54: you can't pass in an `id` during `createLesson` because `id` is auto generated.

graphql/typeDefs.ts
```diff
@@ -42,6 +44,40 @@ export default gql`
...
+      docUrl: String
+      githubUrl: String
+      videoUrl: String
+      title: String
+      chatUrl: String
+      order: Int
+    ): SuccessResponse
+    createChallenge(
+      lessonId: Int!
+      id: Int!
```
songz, line 69: you can't pass in an `id` during `createLesson` because `id` is auto generated.

helpers/controllers/adminController.ts
```diff
@@ -0,0 +1,45 @@
+import db from '../dbload'
+import { Context } from '../../@types/helpers'
+import _ from 'lodash'
+import { isAdmin } from '../isAdmin'
+
+const { User } = db
+
+type adminData = {
+  username: string
```
songz, line 9: `id` is much faster. your resolver should take id instead.

graphql/queryResolvers/allUsers.test.js
```diff
@@ -0,0 +1,33 @@
...
+
+describe('allUsers resolver', () => {
+  const { User } = db
+
+  test('should return list of users', async () => {
+    User.findAll = jest.fn().mockReturnValue(mockUsers)
+    expect(allUsers(null, null, ctx)).toEqual(mockUsers)
+  })
+
+  test('Should throw Error when user is not an admin when querying allUsers', () => {
```
danielthalhuber, line 29: I see that the resolver returns `null` when the `isAdmin` function returns a falsey value. But where is the Error thrown?

anthonykhoa: You are right, let me update that test description. Thank you!

helpers/controllers/challengesController.ts
```diff
@@ -0,0 +1,60 @@
+import db from '../dbload'
+import { Context } from '../../@types/helpers'
+import _ from 'lodash'
+import { isAdmin } from '../isAdmin'
+
+const { Challenge } = db
```
danielthalhuber, line 6: I noticed this pattern in the controller modules and their tests where `db` is imported, and then destructured on a separate line. Is there a reason why we don't just destructure when importing?

anthonykhoa: I don't know, but I think that it looks cleaner if we destructure when importing. So.. i will look at my files again and update as necessary. Thanks!

anthonykhoa: The reason why, is because we are using `export default` in the `dbload.ts` file.
```
export default {
  Lesson,
  Challenge,
  Submission,
  User,
  UserLesson,
  Alert,
  sequelize
}
```
And so it isn't possible to destructure when importing.

danielthalhuber: Thank you for explaining this! So `import { Challenge } from '../dbload'` would fail because the object `default` is exported, but not mentioned in the import, so `Challenge` would not be found.

OK, that brings up another question. I won't delay this PR any further, but I do wonder why we're using the `export default` pattern in `dbload` instead of named exports.

A quick survey shows that we're usually, if not always, importing the default export object from `dbload` and then destructuring it on a different line.

anthonykhoa: I wonder about why we use `export default` in our `dbload` file too. If there isn't a good reason for using `export default`, then I think a refactoring would be nice.

songz: refactoring backlog.

PR #341, opened 2020-07-25
closes #340 - update return information of update/create challenge re…

### LINKED ISSUE #340: Updating return information of update/create challenge resolvers
A return response of `{ success: true }` is not meaningful when the front-end sends graphql requests to update or create a new challenge. The front-end needs to receive the updated list lessons and challenges instead of a `{ success: true }` object.

anthonykhoa:
A return response of `{ success: true `} is not meaningful when the front-end sends graphql requests to update or create a new challenge.

The front-end needs to receive the updated list lessons and challenges instead of a { success: true } object. Making this change will remove the need for the front-end to send another graphql query to retrieve updated data after sending a createChallenge/updateChallenge mutation request.

This PR will:

* Change the response of the updateChallenge and createChallenge resolvers to return updated list of lessons and challenges
* Update the createChallenge and updateChallenge queries in order to reflect the changes from their respective resolvers

### DIFF
--- a/helpers/controllers/challengesController.ts
+++ b/helpers/controllers/challengesController.ts
@@ -20,7 +20,6 @@ export const createChallenge = async (
   ctx: Context
 ) => {
   const { req } = ctx
-  console.log(req.user)
   try {
     if (!isAdmin(req)) {
       throw new Error('User is not an admin')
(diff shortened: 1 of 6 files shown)


helpers/controllers/challengesController.ts
```diff
@@ -19,6 +20,7 @@ export const createChallenge = async (
   ctx: Context
 ) => {
   const { req } = ctx
+  console.log(req.user)
```
SahilKalra98, line 23: Is console.log necessary here?

anthonykhoa: It is not, i have removed it thanks!

PR #351, opened 2020-07-25
closes #348 - FormCard

### LINKED ISSUE #348: InputCard Component
A reusable FormCard component needs to be created for the admin pages. This is because the `lessons` and `alerts` admin page need to be able to take in inputs for various fields.

anthonykhoa:
A reusable FormCard component needs to be created for the admin pages. This is because the `lessons` and `alerts` admin page need to be able to take in inputs for various fields.

## What does component does:
1) Renders a form consisting of `input fields` and their respective labels. The `values` prop passed in will provide the information for the text that the input fields and labels display. The `values` prop is an `array of options`.
2) The `array of options` will be updated every time someone alters an input field
3) Optional rendering of buttons and title, both passed in as props
4) Optional field validation. Right now there are two options:
* check for field value to be numbers only
* check if field value is required 

## How to use it to retrieve information from input fields:
* The `FormCard` component allows buttons to have callback functions. When these callback functions are called, they will take in the `updated array of options` as a parameter.

# Basic
Screen Shot 2020-07-27 at 3 49 35 PM

# With Title
Screen Shot 2020-07-27 at 3 49 41 PM

# With Button
Screen Shot 2020-07-27 at 3 21 04 PM

# With Title and Button
Screen Shot 2020-07-27 at 3 20 14 PM

# Examples of field validation 
Screen Shot 2020-07-30 at 2 46 39 PM


components/FormCard.tsx
```diff
@@ -0,0 +1,142 @@
...
+export const OptionsList: React.FC<OptionsListProps> = ({
+  handleChange,
+  capitalizeTitle,
+  options
+}) => {
+  const inputs = options.map((option: Option, i: number) => {
+    const { title, value, placeHolder } = option
+    if (option.title === 'id') return []
+    const inputType =
+      title === 'description' ? (
```
songz, line 81: instead of hardcoding, could you set `if (option.type === MDINPUT)`

this component should export a few types.

components/FormCard.tsx
```diff
@@ -0,0 +1,142 @@
...
+}
+
+type Btn = {
+  title: string
+  onClick?: Function
+}
+
+type FormCardProps = {
+  values: Option[]
+  buttons?: Btn[]
```
songz, line 22: what is this option for? could we also have a function as a prop? to be called when users hit submit button. this prop should be required.

anthonykhoa: the function is passed into the buttons prop. This component will be re-done using formik

components/FormCard.tsx
anthonykhoa: yes, i will do that. thanks!

components/FormCard.tsx
anthonykhoa: the function prop is the `button` prop. i will change the naming.

stories/components/FormCard.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const mockValues = [
+  { title: 'Nickname', value: 'ChickenFarmer00' },
+  { title: 'Description', value: 'Amazing farm of chicken' },
+  { title: 'Title', value: 'oheuhehe' },
+  { title: 'Level', value: 'noob' }
+]
+
+export const Basic: React.FC = () => (
+  <FormCard values={mockValues} onSubmit={mockBtn} />
```
songz, line 22:
```
const Admin = () => {
  const [mockValues, setMockValues] = useState( [....] )
  // all the admin check logic
  /*
{ title: 'Nickname', value: '.....', error: '....' }
  */
  return (
<FormCard values={mockValues} onChange={ (title, newVal) => {
  if (title === 'Nickname' && !newVal.length) {
    const newState = [...mockValues]
   newState[title].error = 'cannot be empty'
   setMockValue( newState) 
} onSubmit= {mockBtn} />
  )
}
```

PR #356, opened 2020-07-31
closes #355 - Add sidebar to admin lessons page

### LINKED ISSUE #355: Admin Lessons Page Sidebar
The lessons page needs a sidebar display a list of `lesson titles` as well as a `create new lesson` button. When someone clicks on a `lesson title`, a state should be updating its value.

anthonykhoa:
The lessons page needs a sidebar display a list of lesson titles as well as a create new lesson button. Each state value corresponds to an index  in the `lessons` array(`which is received from a graphql query`).  When someone clicks on a lesson title, the state should be updated with the correct index value.

This PR will:
* Create a sidebar for the admin lessons page that displays a list of lesson titles
* Display the state value upon page load as well as after each lesson title is clicked on.

## Note:
The next PRS will make changes so that the `state value` will not be displayed anymore. Instead, the information of a lesson and all of its challenges will be displayed.

Screen Shot 2020-08-01 at 4 29 40 PM


@types/adminLesson.d.ts
```diff
@@ -0,0 +1,10 @@
+export type Lesson = {
+  id: string
+  title: string
+  description: string
+  order: number
+  challenges: Challenge[]
+  docUrl: string
+  chatUrl: string
+  __typename: string
+}
```
songz, line 10: Can't you use the type from `graphql/index` because this comes from graphql?

anthonykhoa: Let me try to see if that can work

components/AdminLessonsSideBar.tsx
```diff
@@ -0,0 +1,53 @@
...
+  return lessonList
+}
+
+export const AdminLessonsSideBar: React.FC<SideBarLessonProps> = ({
+  lessons,
+  setSelectedLesson
+}) => {
+  const lessonsList = lessons
+    ? makeLessonsList(lessons, setSelectedLesson)
+    : [<h1 key={0}>Create new Lesson</h1>]
```
songz, line 50: would this create a bug where if there are no lessons, the `createNewLesson` is not clickable?

```
let lessonListData = [{title: 'Create New Lesson'}]
lessonListData = lessonListData.concat( lessons || [] )
const lessonList = lessonListData.map(.....)
```

I think `makeLessonsList` function is redundant.

anthonykhoa: Correct that would make a bug. I will fix ! thanks

components/AdminLessonsSideBar.tsx
```diff
@@ -0,0 +1,45 @@
...
+      </div>
+    </div>
+  </div>
+)
+
+export const AdminLessonsSideBar: React.FC<SideBarLessonProps> = ({
+  lessons,
+  setSelectedLesson
+}) => {
+  let lessonListData = lessons || []
```
songz, line 33:
```
const lessonListData = lessons || []
lessonListData.push({title: '....'})
const lessonList = .....
```

Better to use `const`, good stuff for refactoring the original comment.

pages/admin/lessons.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Layout>
+      {loading && <LoadingSpinner />}
+      {error && <h1>Error</h1>}
+      {session && session.user && session.user.isAdmin === 'true' && (
+        <AdminLessons
+          lessons={lessonsList || lessons}
+          setLessons={setLessons}
+        />
+      )}
```
songz, line 38: instead of doing it like this, it is better to be explicit in your return. Like... there is no scenario where loading is truthy and `session.user` information exist.

```
if (loading) {
  return (
     <Layout> <Loading Spinner /> </Layout>
  )
}
if (error) {
  return (....)
}
const isAdmin = _.get(session, 'user.isAdmin', false)
if (!isAdmin) {
  return (<Layout><h1>You must be admin to access this page</h1></Layout>)
}
return (
...
)
```

components/AdminLessonsSideBar.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
```
songz, line 1: move this file to `components/admin/lessonSideBar`

songz: hopefully there is an admin folder inside `components`

components/AdminLessonsSideBar.tsx
songz: since this is not really reusable and not too long, does it make sense to put this inside `pages/admin/lessons.tsx`?

components/AdminLessonsSideBar.tsx
songz: since this is not really reusable and not too long, does it make sense to put this inside `pages/admin/lessons.tsx`

anthonykhoa: I think it would be better to make an `lessons` folder like
`components/admin/lessons/lessonSideBar`
That way, if the other admin pages need to make un-reusable components, another folder can be created inside
`components/admin` to store the files.

For example:
`components/admin/users` <-- here, all components needed for the admin users page can go here

components/admin/lessons/AdminLessonsSideBar.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type SideBarLessonProps = {
+  lessons: Lesson[] | undefined
+  setLessons: React.Dispatch<React.SetStateAction<Lesson[] | null>>
+  setSelectedLesson: React.Dispatch<React.SetStateAction<number>>
+}
+
+type LessonProps = {
+  obj: { title: string }
+  lessonId: number
```
songz, line 12: `arrIndex`

PR #366, opened 2020-08-03
closes #360 - Add lesson challenges to admin lessons page

### LINKED ISSUE #360: Admin Lesson Challenges
A list of challenges with information need to be displayed when someone clicks on a lesson title.

anthonykhoa:
Each lesson needs to display the challenges that it has. These displayed challenges need to be able to be updated.

This PR will:
* Display lesson challenges of each lesson
* Enable users to update any current challenges
* Enable users to create new challenges for any lesson
* Change name of `makeLessonVariable` into `makeGraphqlVariable` since it is used for challenges as well.
* Added an `addProp` optional parameter to the `makeGraphqlVariable` function, in order to make it work with challenges

# Create New Challenge:
Screen Shot 2020-08-03 at 6 39 52 PM

# Edit Challenge: 
Screen Shot 2020-08-02 at 8 57 04 PM

### DIFF
--- a/components/admin/lessons/AdminLessonChallenges.tsx
+++ b/components/admin/lessons/AdminLessonChallenges.tsx
@@ -93,7 +93,7 @@ export const NewChallenge: React.FC<NewChallengeProps> = ({
   )
 }
 
-const OneChallenge: React.FC<OneChallengeProps> = ({ challenge, alter }) => {
+const LessonChallenge: React.FC<OneChallengeProps> = ({ challenge, alter }) => {
   const [challengeProperties, setChallengeProperties] = useState(
     getPropertyArr(challenge, ['lessonId'])
   )
@@ -155,7 +155,7 @@ export const AdminLessonChallenges: React.FC<LessonChallengesProps> = ({
   const allChallenges = !challenges
     ? []
     : challenges.map((challenge: Challenge, key: number) => (
-        <OneChallenge challenge={challenge} alter={alter} key={key} />
+        <LessonChallenge challenge={challenge} alter={alter} key={key} />
       ))
 
   return <>{allChallenges}</>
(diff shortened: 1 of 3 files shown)


components/admin/lessons/AdminLessonInfo.tsx
```diff
@@ -147,15 +148,46 @@ export const AdminLessonInfo: React.FC<LessonInfoProps> = ({
...
   const lesson = lessons && lessons[selectedLesson]
+
+  const lessonId = parseInt(lesson ? lesson.id + '' : '')
+
+  const newChallengeClick = () => setNewChallengeView(!newChallengeView)
+
+  const viewOption = newChallengeView ? (
+    <NewChallenge setLessons={setLessons} lessonId={lessonId} />
+  ) : (
+    <>
```
Oroul, line 168: What do empty component tags resolve to?

anthonykhoa: They are called fragments https://reactjs.org/docs/fragments.html
It is so that I don't have to use a `div` html element to wrap list of children

Oroul: Ah, ok. I hadn't seen the short syntax before.

songz: @anthonykhoa I thought they need to be self closing? `< />`

danielthalhuber: The closing tag is on line 180.

components/admin/lessons/AdminLessonChallenges.tsx
```diff
@@ -0,0 +1,164 @@
...
+      setChallengeProperties(newProperties)
+      return
+    }
+
+    try {
+      await createChallenge(
+        makeGraphqlVariable(challengeProperties, { lessonId })
+      )
+
+      window.location.reload()
```
danielthalhuber, line 65: Why do we force a refresh after creating a challenge?

anthonykhoa: There was a problem whenever I clicked the submit button to create a challenge. It would cause the page to not function correctly anymore. After trying a long to figure out without coming to a solution, song suggested for me to just reload the page for now since this `admin lessons` page will not be used very often.

songz: created an issue: https://github.com/garageScript/c0d3-app/issues/368

PR #369, opened 2020-08-04
closes #362 Adds graphql query for Stars Received/Profile Page

### LINKED ISSUE #362: Update profile query to include stars for a given lesson
As an engineer, I should be able to send a graphql query for profile and get an array of stars given out for each lesson.

SahilKalra98:
Modifies ```userInfo``` query so that the ```Profile Page``` is ready to display new UI that includes how many stars they have received for each lesson.

### DIFF
--- a/helpers/controllers/userInfoController.ts
+++ b/helpers/controllers/userInfoController.ts
@@ -5,7 +5,8 @@ type Username = {
   username: string
 }
 
-const { User, UserLesson, Submission } = db
+const { User, UserLesson, Submission, Star } = db
+
 export const userInfo = async (_parent: void, args: Username) => {
   const username = _.get(args, 'username')
   if (!username) {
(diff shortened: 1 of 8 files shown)


helpers/controllers/userInfoController.ts
```diff
@@ -29,8 +31,23 @@ export const userInfo = async (_parent: void, args: Username) => {
...
       }
+    }),
+    Star.findAll({
+      where: {
+        mentorId: user.id
+      }
     })
   ])
+  const starMap = starsReceived.reduce((map: any, star: any) => {
+    map[star.lessonId] = map[star.lessonId]?.concat(star.dataValues) || [
```
anthonykhoa, line 42: I think it would be more contextual to use `push` over `concat` in this instance because what you are doing is pushing a new element to the `map[star.lessonId]` array. It would also be more efficient since concat will make a loop through the array, while push will not

PR #373, opened 2020-08-04
closes #372 - Add allUsers query

### LINKED ISSUE #372: Create graphql query for all users
A graphql resolver to get information on all users has been made, but the query to use in the front-end has not been made yet. This query needs to be made in order for the upcoming `Admin Users` to display a list of users.

anthonykhoa:
The Admin Users page needs to be able to display a list of users and let admins `remove` or `grant` admin rights to other users. Resolvers for allUsers and changeAdminRights have been made, but queries for them have not. These queries are both needed for the Admin User page. 

There is also a problem with the `LoadingSpinner` component. It does not take up the full width and height of the browser window when it is wrapped by the `Layout` component. This needs to be addressed, since the `Admin Alerts` page will run into this problem as well if not fixed.

This PR will:
* Create a graphql query that calls the allUsers resolver to use in the front-end
* Create a graphql query that calls the changeAdminRights resolver to use in the front-end
* Create an `Admin Users` page that will display a list of users with a button to remove or grant admin rights to any user
* Fix css styling on `LoadingSpinner` component so that it takes up 100% width and height of browser screen when wrapped by the `Layout` component

# Title and beginning of Users Information Table
Screen Shot 2020-08-03 at 10 59 11 PM

# Examples of `remove/grant` buttons
Screen Shot 2020-08-03 at 10 59 45 PM

# Example of previous problem with loading spinner
Screen Shot 2020-08-05 at 10 30 59 AM

### DIFF
--- a/pages/admin/users.tsx
+++ b/pages/admin/users.tsx
@@ -1,17 +1,11 @@
 import React, { useState } from 'react'
-import { useQuery } from '@apollo/react-hooks'
-import LoadingSpinner from '../../components/LoadingSpinner'
-import Layout from '../../components/Layout'
-import { withGetApp, GetAppProps } from '../../graphql'
 import allUsers from '../../graphql/queries/allUsers'
-import _ from 'lodash'
 import { UsersTable } from '../../components/admin/users/AdminUsersTable'
-import { User } from '../../graphql/index'
-
-type AdminUsersProps = {
-  users: User[] | null
-  setUsers: React.Dispatch<React.SetStateAction<User[] | null>>
-}
+import { User, withGetApp, GetAppProps } from '../../graphql/index'
+import { AdminLayout } from '../../components/admin/AdminLayout'
+import withQueryLoader, {
+  QueryDataProps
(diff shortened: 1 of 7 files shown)


pages/admin/users.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return (
+      <Layout>
+        <h1>Error</h1>
+      </Layout>
+    )
+  }
+
+  if (!session) window.location.href = '/login'
+
+  if (admin === 'false') {
```
songz, line 75: useadmin layout

pages/admin/users.tsx
```diff
@@ -0,0 +1,86 @@
...
+    </div>
+  )
+}
+
+const LoadUsers: React.FC = () => {
+  const { loading, error, data } = useQuery(allUsers)
+  const [users, setUsers] = useState<null | User[]>(null)
+
+  if (loading) {
+    return <LoadingSpinner />
```
songz, line 39: use `useQuery`

PR #377, opened 2020-08-06
closes #376 - Implement search and admin-filter functionality into Admin Users page

### LINKED ISSUE #376: Implement search and admin-filter functionality into Admin Users page
Users should be able to see a search bar, search options, as well as filter options when on the admin users page. These functionalities should all work as well.

anthonykhoa:
Users should be able to see a search bar, search options, as well as filter options when on the admin users page. These functionalities should all work as well.

This PR will:
* Create and implement `filter/search options` and `search bar` into admin users page
* Set up a state that changes when someone clicks on another `search` or `filter` option button
* Remove some custom css styling from `pages/admin/users.tsx` to look more professional, used bootstrap classes instead where possible
* Highlight searchTerm in the search results

Screen Shot 2020-08-09 at 9 28 29 PM


components/admin/users/AdminUsersSearchBar.tsx
```diff
@@ -0,0 +1,35 @@
...
+      searchTerm: string
+    }>
+  >
+}
+
+export const SearchBar: React.FC<SearchBarProps> = ({
+  searchOption,
+  setSearchOption
+}) => {
+  const run = _.debounce(setSearchOption, 500)
```
songz, line 24: you don't need debounce, because all the data is already there. you are just searching through existing data so this should be instant.

anthonykhoa: Debounce is needed because when someone is typing in the searchbar, it becomes very laggy. I noticed that the users list was re-rendering way too often. This re-rendering issue was happening before I added in the option to `highlight` the searchTerm in the users list too.

anthonykhoa: Could this be a problem similar to what we faced with the `ReviewCard` component? 

Using `memo` will not work in this instance because in the next PR, the component that displays a list of users is going to take in a `searchOption` property, whose value will change each time someone times in the search bar.

songz: thanks for verifying, I believe you and I wouldn't have guessed that. Could you add a comment in case someone else has the question?

components/admin/users/AdminUsersTable.tsx
```diff
@@ -4,132 +4,167 @@ import _ from 'lodash'
...
+  const list = users.reduce((acc: any[], user: any, usersIndex: number) => {
+    if (searchTerm) {
+      const value = user[option]
+      if (!value.includes(searchTerm)) return acc
+    }
+
+    const { isAdmin } = user
+
+    if (admin === 'Non-Admins' && isAdmin === 'true') return acc
+    if (admin === 'Admins' && isAdmin === 'false') return acc
```
songz, line 130: I don't think you need an `isAdmin` check here because non-admins won't even be able to see the page.

components/admin/users/AdminUsersTable.tsx
```diff
@@ -4,132 +4,167 @@ import _ from 'lodash'
...
+const UsersList: React.FC<UsersListProps> = ({
+  users,
+  setUsers,
+  searchOption
+}) => {
+  const { searchTerm, admin } = searchOption
+  let { option } = searchOption
+  option = _.lowerCase(option)
+
+  const list = users.reduce((acc: any[], user: any, usersIndex: number) => {
```
songz, line 121: IMO, its better to filter -> map instead of reduce. This way you can clearly see that you are filtering through the results and looking through filtering criteria.

```
users.filter( (user) => {
  if (searchTerm) {
    return (user[option] || '').includes(searchTerm)
  }
  if (admin === 'Non-Admins') return user.isAdmin === 'false'
  if (admin === 'Admins') return user.isAdmin === 'true'
  return user
}).map( (user, userIndex) => {
  return (
    <tr ....
    </tr>
  )
})
```

components/admin/users/AdminUsersSplitSearch.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  return split(str, searchTerm, strIndex + 1, res, current)
+}
+
+export const AdminUsersSplitSearch: any = (str: string, searchTerm: string) => {
+  const splitArr = split(str, searchTerm)
+  const value = splitArr.map((word: string) => {
+    const bgColor = word !== searchTerm ? 'none' : 'rgb(84, 64, 216, .25)'
+    return (
+      <span style={{ backgroundColor: bgColor }} key={_.uniqueId()}>
```
songz, line 57: I thought `key={_.uniqueId()}` is bad practice. it forces constant refreshes even when unnecessary.

components/admin/users/AdminUsersSplitSearch.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
+import _ from 'lodash'
+
+const split: any = (
```
songz, line 4: hahahaah its so amazing you wrote your own split function haha. there is `.split` function that you should just use.

```
"hello".split('ell') // ['h', 'o']
```

components/admin/users/AdminUsersSplitSearch.tsx
```diff
@@ -0,0 +1,63 @@
...
+    !current && res.push(str[strIndex])
+  } else {
+    res.push(str[strIndex])
+  }
+
+  return split(str, searchTerm, strIndex + 1, res, current)
+}
+
+export const AdminUsersSplitSearch: any = (str: string, searchTerm: string) => {
+  const splitArr = split(str, searchTerm)
```
songz, line 53: `str.toLowerCase().split(searchTerm.toLowerCase())`

components/admin/users/AdminUsersSplitSearch.tsx
anthonykhoa: My split function will include the delimiter.
split('hello', 'ell') // ['h', 'ell', 'o']
The delimiter is needed so that `line 54` (`const value = splitArr.map((word: string) => {`)
can highlight the searchTerm

components/admin/users/AdminUsersSplitSearch.tsx
anthonykhoa: Thanks, i did not think of converting all characters to lowercase on searches. I will do that!

components/admin/users/AdminUsersSplitSearch.tsx
anthonykhoa: It is needed in this instance, or else the `highlighting of searchTerm` will not highlight the correct words.

Example: see the last row  in this column. It is highlighted incorrectly when not using `key={_.uniqueId()}`
Screen Shot 2020-08-10 at 10 52 48 AM

components/admin/users/AdminUsersSplitSearch.tsx
ggwadera: Can't this be done using regex and replace? Like `"hello".replace(/ell/g, "<span style={{ backgroundColor: bgColor }}>ell</span>")`

songz: @ggwadera using `span` tag in quotes as a string instead of react component makes it inconsistent.

@anthonykhoa I see your point, perhaps you could use the [`_.join` function from lodash](https://lodash.com/docs/4.17.15#join)

```
const splitStr = _.join(str.toLowerCase().split(....), ...)
```

anthonykhoa: I didn't think to use join. That would make my code much cleaner. Thanks!

anthonykhoa: @songz I am experiencing trouble using `_.join`. I need to use an array, but using `_.join` returns a string. I will try to think of other logic to implement.

components/admin/users/AdminUsersSplitSearch.tsx
songz: then the key should be `word+searchTerm+index`. This way, if the search term and the word stays the same, it will not rerender.

components/admin/users/AdminUsersSplitSearch.tsx
songz: @anthonykhoa 

```
'hello'.split('ell').reduce( (acc, e, i) => {
  acc.push( <span>{e} </span> )
  if (i < 'hello'.length - 1) {
    arr.push( <span className="matchStr">ell</span> )
  }
  return acc
})
```

components/admin/users/AdminUsersSplitSearch.tsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react'
+
+const split = (str: string, lowerCaseSearchTerm: string) => {
```
songz, line 3: could you add a comment? 

```
/* Sample input: 'hello', 'el'
  * Sample output: [....]
*/
```

components/admin/users/AdminUsersSplitSearch.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const splitArr = str.toLowerCase().split(lowerCaseSearchTerm)
+  let tracker = 0
+
+  const res = splitArr.reduce((acc: string[], word: string) => {
```
songz, line 7: instead of using a `tracker`, couldn't you use the index? 

```
...reduce( (acc: string[], word: string, i: number) => {
```

anthonykhoa: I tried using index like you originally suggested, but it didn't work. It kept on adding the `searchTerm` back into the array when it shoudn't. 

Can you explain the logic of using `index` instead of `tracker`? I don't understand the reasoning behind using `index`

components/admin/users/AdminUsersSplitSearch.tsx
```diff
@@ -0,0 +1,73 @@
...
+    acc.push(word)
```
songz, line 8:
```
const originalWord = str.substr( tracker, word.length )
acc.push(originalWord)
```

components/admin/users/AdminUsersSplitSearch.tsx
anthonykhoa: Can you explain to me the logic of using `index`? I don't understand how it can be used.

anthonykhoa: I tried your original suggestion before I tried using tracker, but then I used `tracker` because using `index` wasn't working for me.

components/admin/users/AdminUsersSplitSearch.tsx
```diff
@@ -0,0 +1,56 @@
...
+	convert string to array, separated by and including searchTerm
+	Ex:
+		Inputs: searchTerm='bon', str = 'bonJourbon'
+		Output: [bon, Jour, bon]
+*/
+const split = (
+  str: string,
+  lowerCaseSearchTerm: string,
+  searchTerm: string
+) => {
```
songz, line 13: `const lowerCaseSearchTerm = searchTerm.toLowerCase()` you don't need to pass this in as a variable.

PR #379, opened 2020-08-07
[BUG] Stars is undefined for Profile Page

SahilKalra98:


pages/profile/[username].tsx
```diff
@@ -100,12 +101,16 @@ const UserProfile: React.FC = () => {
       },
       {}
     )
+    let starsReceived = [] as Star[]
```
Comment (anthonykhoa) [inline pages/profile/[username].tsx:104]: i think its better to use 
`let starsReceived: Star[] | undefined`
than
` let starsReceived = [] as Star[]`
that way you don't have declare types for the same variable twice

Comment (songz) [reply pages/profile/[username].tsx:104]: I think its okay to do `let starsReceived = [] as Star[]`, because then this always guarantees that the return object's `starsReceived` key is always array.

PR #384, opened 2020-08-15
closes #383 - searchTerm not capitalized correctly when highlighted

### LINKED ISSUE #383: Admin Users page not capitalizing searchTerm correctly when highlighting searchTerm
When highlighted, the searchTerm is not being correctly capitalized to how it appeared in the original string. 
Example:
Screen Shot 2020-08-15 at 10 30 30 AM

The last two highlighted `E`'s should be lowercase, not uppercase.

anthonykhoa:
When highlighted, the searchTerm is not being correctly capitalized to how it appeared in the original string. 
## Example of problem:
Screen Shot 2020-08-15 at 10 30 30 AM

The last two highlighted `E`'s should be lowercase, not uppercase.

This PR will:
* capitalize the searchTerm to how it appears in the original string when highlighting it in the search results.
* put `splitWithSearchTerm` into a separate file so it can be re-usable 
* create tests for `splitWithSearchTerm` to ensure capitalization is correct in its return value

## Example of fix:
Screen Shot 2020-08-15 at 10 44 31 AM

### DIFF
--- a/helpers/splitWithSearchTerm.tsx
+++ b/helpers/splitWithSearchTerm.tsx
@@ -7,9 +7,10 @@
 
 export const splitWithSearchTerm = (str: string, searchTerm: string) => {
   /*
-	Using regex with split will return an array with the separator included.
-	The `i` flag is the case insensitive flag, used to ensure that the returned array
-	will include the searchTerm, regardless of the original capialization of the searchTerm
+	Using regex containing capturing parentheses `()` with split will return an array
+	with the separator included. The `i` flag is the case insensitive flag, used to
+	ensure that the returned array will include the searchTerm, regardless of the
+	original capialization of str or searchTerm.
   */
   const regex = new RegExp(`(${searchTerm})`, 'i')
   return str.split(regex).filter(s => s !== '')
--- a/helpers/splitWithSearchTerm.tsx
+++ b/helpers/splitWithSearchTerm.tsx
@@ -4,30 +4,14 @@
 		Inputs: searchTerm='bon', str = 'bonJourbon'
 		Output: [bon, Jour, bon]
 */
+
 export const splitWithSearchTerm = (str: string, searchTerm: string) => {
-  const lowerCaseSearchTerm = searchTerm.toLowerCase()
-  const splitArr = str.toLowerCase().split(lowerCaseSearchTerm)
-  // tracker is used for `str.substr` to extract characters from the original string at the correct places
-  let tracker = 0
-
-  const res = splitArr.reduce(
-    (acc: string[], word: string, splitArrIndex: number) => {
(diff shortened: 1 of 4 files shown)


helpers/splitWithSearchTerm.tsx
```diff
@@ -0,0 +1,33 @@
+/*
+	convert string to array, separated by and including searchTerm
+	Ex:
+		Inputs: searchTerm='bon', str = 'bonJourbon'
+		Output: [bon, Jour, bon]
+*/
+export const splitWithSearchTerm = (str: string, searchTerm: string) => {
```
ggwadera, line 7: This could potentially be replaced by a simpler code using split with a regex with a capturing group to include the search term and case insensitive flag (i). For example:

```js
const splitWithSearchTerm = (str, searchTerm) => {
  const regex = new RegExp(`(${searchTerm})`, 'i')
  return str.split(regex).filter(s => s !== '')
  // filter to remove the empty strings that appear if searchTerm is at the start or end
}
```

Results in:

```js
splitWithSearchTerm('bonJourbon', 'bon') = [ 'bon', 'Jour', 'bon' ]
splitWithSearchTerm('hiNgaDiNGADuurgen', 'dINga') = [ 'hiNga', 'DiNGA', 'Duurgen' ]
splitWithSearchTerm('hiNgaDiNGADuurgen', 'DINGA') = [ 'hiNga', 'DiNGA', 'Duurgen' ]
splitWithSearchTerm('hiNgaDiNGADuurgen', 'dinga') = [ 'hiNga', 'DiNGA', 'Duurgen' ]
```

helpers/splitWithSearchTerm.tsx
```diff
@@ -4,30 +4,14 @@
...
-    },
-    []
-  )
-
-  return res
+export const splitWithSearchTerm = (str: string, searchTerm: string) => {
+  /*
+	Using regex with split will return an array with the separator included.
+	The `i` flag is the case insensitive flag, used to ensure that the returned array
+	will include the searchTerm, regardless of the original capialization of the searchTerm
```
ggwadera, line 12: Just a correction in the documentation. What actually makes `searchTerm` be included in the returned array are the parenthesis. The `i` flag is only for case insensitive match.

As per the MDN example:

> If `separator` is a regular expression that contains capturing parentheses `( )`, matched results are included in the array.

anthonykhoa: Ah I see there has been a misunderstanding, because I indeed was trying to explain in my comment that the `i` flag is used for case sensitivity as you have thought I didn't. 

Thank you for bringing this to my attention, I have updated my comment in hopes that people are less confused by my `i` flag explanation.

PR #388, opened 2020-08-28
closes #387 - change functions in lessonHelpers so they can be usable for admin alerts page

### LINKED ISSUE #387: Change helpers/admin/lessonHelpers file so that admin alerts page can use it
The `lessonHelpers` has 2 functions that need to be changed in order for it to be usable for the admin alerts page.

1) Function `makeGraphqlVariable` needs to check if the `order` property exists or typeError will occur
2) Functions that check for errors for FormCard component need to be replaced with using Yup for error checking

anthonykhoa:
The lessonHelpers has several functions that need to be changed in order for it to be usable for the upcoming admin alerts page.
These functions are:
1) Function `makeGraphqlVariable` needs to check if the order property exists or typeError will occur
2) Functions `checkForErrors` and `checkForAllErrors` need to take in required and numbersOnly properties. This is because the admin alerts page have different requirements on what property fields must be required what property fields can only contain numbers only.

This PR will:
* Change `makeGraphqlVariable`, `checkForErrors`, and `checkForAllErrors` functions so that it is usable for upcoming admin alerts page
* Change name of file from `lessonHelpers` to `processingHelpers` because the file is not only used for the admin lessons page
* Change code for admin lessons page to reflect update to the functions in the `processingHelpers` file

### DIFF
--- a/components/admin/lessons/AdminLessonChallenges.tsx
+++ b/components/admin/lessons/AdminLessonChallenges.tsx
@@ -9,8 +9,8 @@ import {
   checkForErrors,
   checkForAllErrors,
   makeGraphqlVariable
-} from '../../../helpers/admin/lessonHelpers'
-import { titleStyle } from './AdminLessonInfo'
+} from '../../../helpers/admin/processingHelpers'
+import { titleStyle, errorChecks } from './AdminLessonInfo'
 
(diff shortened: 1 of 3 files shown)


songz:
closing because we are gonna use `yup`

PR #399, opened 2020-09-22
Initial UI Implementation (Give Star Feature)

Cijin:
Just the initial setup for the UI. So far the following should happen on clicking the Give Star Button:

1. A card should pop up with an input box for getting the username of the person who helped you the most.
2. You should be able to close the popup.

The next step: 
1. Create a gql query that returns only eligible users list from the search term if provided or all eligible users. I'm calling it for now GetMentors, it should take lesson id as a parameter.
2. Display the cards of the returned users by the above query.
3. Finally, give star on click + make mutation request.

Thoughts on the feature(open to suggestions)
1.Should a user be able to give star multiple times?
2. User should be able to give star to a mentor only once per lesson.


songz:
closing, out of date. You need a graphql query to get all users who have passed a lesson first.

PR #403, opened 2020-09-25
Issue#402

Cijin:
Removes the optional chaining syntax '?' from the AppNav.tsx file.

This syntax is causing fresh installs to fail on some node versions.


components/AppNav.tsx
```diff
@@ -68,7 +68,8 @@ const AuthLink: React.FC<AuthLinkProps> = ({ session }) => {
 const AuthButton: React.FC<AuthButtonProps> = ({ initial, username }) => {
   const [logoutUser, { data }] = useLogoutMutation()
   useEffect(() => {
-    if (data?.logout?.success) {
+    const success = _.get(data, 'logout', false)
```
songz, line 71: shouldn't it be `_.get(data, 'logout.success', false)`

pages/curriculum.tsx
```diff
@@ -69,7 +71,7 @@ export const Curriculum: React.FC<GetAppProps> = ({ data }) => {
         currentState={lessonState}
         reviewUrl={`/review/${id}`}
         challengesUrl={`/curriculum/${id}`}
-        docUrl={lesson?.docUrl || ''}
+        docUrl={lesson.docUrl || ''}
```
hwong0305, line 74: Shouldn't this be `_.get(lesson, 'docUrl', '')` or does the GraphQL definition already guarantee that it exists?

Cijin: I thought since it's got a fallback `|| ''`. Wasn't completely sure about it.

anthonykhoa: I believe @hwong0305 is right. If `lesson` is undefined, then `lesson.docUrl` would cause a TypeError

Cijin: Using `_.get(lesson, 'docUrl', '')` creates a type error saying `type Maybe<String> is not assignable to type string`. How can I fix that?
If lesson is undefined an error would be thrown earlier saying `<h1>Bad Data</h1>`

pages/curriculum.tsx
```diff
@@ -48,11 +49,12 @@ export const Curriculum: React.FC<GetAppProps> = ({ data }) => {
   const lessonsToRender: React.ReactElement[] = lessons.map((lesson, idx) => {
     const id = _.get(lesson, 'id', idx) as number
     const status = lessonStatusMap[id]
+    const passed = _.get(status, 'isPassed', false)
     let lessonState = ''
     if (idx === lessonInProgressIdx) {
       lessonState = 'inProgress'
     }
-    if (status?.isPassed) {
```
songz, line 55: put variables declarations close to where it is used. so move `const passed = ....` to here.

pages/curriculum.tsx
anthonykhoa: Try `_.get(lesson, 'docUrl', '') as string || ''`

songz (approved):

ggwadera (approved):

anthonykhoa (approved):

PR #404, opened 2020-09-25
Issue#398 - Give Star Feature

Cijin:
1. Created skeleton on GiveStarCard component
2. Added to stories as GiveStarCard -> Primary

Next Steps:

1. Create a card (`userCard` for example) to display a list of users that can receive the stars
2. Create a gql query to return users based on the lesson-id


ggwadera (changes requested):
Could you remove the `package-lock.json` file from the PR? We already have `yarn.lock` in the repo.

components/GiveStarCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+import Modal from 'react-bootstrap/Modal'
+import Button from 'react-bootstrap/Button'
+import InputGroup from 'react-bootstrap/InputGroup'
+import FormControl from 'react-bootstrap/FormControl'
+
+type Props = {
+  show: boolean
+}
+
+const GiveStarCard: React.FC<Props> = props => {
```
anthonykhoa, line 11: You should destructure the props to maintain codebase consistency, because the rest of the codebase destructures props.
```
const GiveStarCard: React.FC<Props> = ({ show }}=> {
```

components/GiveStarCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+          </InputGroup.Prepend>
+          <FormControl
+            placeholder="Username"
+            aria-label="Username"
+            aria-describedby="basic-addon1"
+          />
+        </InputGroup>
+      </Modal.Body>
+      <Modal.Footer>
+        <Button onClick={() => {}}>Close</Button>
```
anthonykhoa, line 37: Is it possible to use our internal Button component instead? It should be in the `components/theme` folder

graphql/queryResolvers/getLessonMentors.ts
```diff
@@ -0,0 +1,27 @@
...
+  lessonId: string
+}
+
+export const getLessonMentors = async (
+  _parent: void,
+  args: ArgsGetLessonMentors
+) => {
+  const { lessonId } = args
+  try {
+    let results = await UserLesson.findAll({
```
anthonykhoa, line 15: results should be declared `const` since its a constant memory address

graphql/queryResolvers/getLessonMentors.ts
```diff
@@ -0,0 +1,27 @@
...
+      where: { lessonId },
+      include: [{ model: User }]
+    })
+
+    return results.map((result: any) => {
+      return result.User
+    })
+  } catch (err) {
+    console.error('Unable to fetch data: ', err)
```
anthonykhoa, line 24: instead of console.log, you should `throw New Error(some message here)`
And then you can test to see if that error message displays in your tests

components/GiveStarCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </Modal.Footer>
+    </Modal>
+  )
+}
+
+export default GiveStarCard
```
anthonykhoa, line 43: Is it guaranteed that there will only be one variable exported from this file? If so, using `export default` is fine.
However, if you have plans to export more variables from this file, then you should export the normal way. This is to give more codebase consistency when someone imports variables from this file.

components/GiveStarCard.tsx
Cijin: Yes will do. :)

components/GiveStarCard.tsx
Cijin: On it! :)

components/GiveStarCard.tsx
hwong0305: I think it's either way it's fine. There's no real reason why we need `named exports`. We should use `default exports` unless we need more functionality such as exporting more than one component from a file.

anthonykhoa: Ah, I see your point -- unless the file will export more than one variable, then both default or named export is fine. Thank you for the lesson Mr. Hyper Herman

ggwadera (approved):

PR #408, opened 2020-09-25
closes #407 - Update addAlert graphql mutation resolver to return meaningful data

### LINKED ISSUE #407: Update addAlert graphql mutation resolver
The [addAlert](https://github.com/garageScript/c0d3-app/blob/master/helpers/controllers/alertController.ts#L15) graphql mutation resolver currently returns info that has no value to the front end. It simply returns `{ success: true }`

This resolver should be updated so that the front end receives meaningful data back from the mutation request.

anthonykhoa:
## The issue

The [addAlert](https://github.com/garageScript/c0d3-app/blob/master/helpers/controllers/alertController.ts#L15) graphql mutation resolver currently returns info that has no value to the front end. 
Right now it simply returns `{ success: true }`

This resolver should be updated so that the front end receives meaningful data back from the mutation request.
This will reduce the complexity of logic needed for the `Admin Alerts` page, since no additional queries need to be made to get an `updated list of alerts` after adding an alert.

## This PR will:
* Update the `addAlerts` resolver to return `updated list of Alerts`
* Update the [addAlert mutation request](https://github.com/garageScript/c0d3-app/blob/master/graphql/queries/addAlert.ts) to request for properties of Alert rather than `success`
* Update tests, typedefs, and the `graphql/index.tsx` file to reflect change in return statement

### DIFF
--- a/graphql/index.tsx
+++ b/graphql/index.tsx
@@ -1,48 +1,41 @@
-import {
-  GraphQLResolveInfo,
-  GraphQLScalarType,
-  GraphQLScalarTypeConfig
-} from 'graphql'
-import gql from 'graphql-tag'
-import * as ApolloReactCommon from '@apollo/react-common'
-import * as React from 'react'
-import * as ApolloReactComponents from '@apollo/react-components'
-import * as ApolloReactHoc from '@apollo/react-hoc'
-import * as ApolloReactHooks from '@apollo/react-hooks'
-export type Maybe<T> = T | null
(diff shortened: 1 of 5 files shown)


ggwadera (approved):

songz:
merging with tech debt: https://github.com/garageScript/c0d3-app/issues/409

songz:
oh shit I merged in the wrong PR. whoops. Thanks for reviewing @ggwadera

Issue #418, opened 2020-10-06
CI doesn't error out if eslint and prettier can be automatically fix


hwong0305:
Need to modify CI script so it doesn't automatically fix prettier and eslint issues.

PR #419, opened 2020-10-06
Lint update

hwong0305:
Fixes issue #418 

This PR will fix the issue where our CI doesn't catch any eslint errors if they can be automatically fixed. The cause of the issue is the CI didn't output any errors when eslint can auto fix issues.

image


songz (changes requested):
breaking tests

songz (approved):

anthonykhoa (approved):

PR #425, opened 2020-11-07
Fix contributor page loading screen styling

bryanjenningz:
Before, when you loaded the contributor page, you'd see a loading
screen where there was a component above the loading page which
looked awkward. This commit fixes that styling bug.

I was able to figure out how to fix the styling by opening
up Chrome devtools, going to the Elements tab, placing a break
point on subtree modifications on the body element, then I refreshed
the page and was able to add styles. It turns out that adding a simple
z-index: 1 was all it took to fix this.

How to reproduce this bug:
* Go to /contributors
* You should see the bug on the loading screen, if it's too fast, go to the Chrome devtools Network tab and simulate 3G connection speed or go to Elements tab and set a break point on subtree modification of the body element then refresh the page.

Before:
Screenshot from 2020-11-07 13-15-33

After:
Screenshot from 2020-11-07 13-16-48

### DIFF
--- a/__tests__/__snapshots__/storyshots.test.js.snap
+++ b/__tests__/__snapshots__/storyshots.test.js.snap
@@ -209,6 +209,7 @@ exports[`Storyshots Components/AppNav Logged In 1`] = `
       "minHeight": "100vh",
       "position": "fixed",
       "top": 0,
+      "zIndex": 1,
     }
   }
 >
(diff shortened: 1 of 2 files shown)


songz (approved):

songz (approved):

ggwadera:
@all-contributors please add @bryanjenningz for code and review

ggwadera:
@all-contributors please add @bryanjenningz for code

PR #459, opened 2021-01-07
closes #458 - update starGiven field to return string instead of Star type

### LINKED ISSUE #458: starGiven needs to be changed to string instead of Star type
In our graphql typedefs, the `starGiven` field of the `lessonStatus` type is defined as a `Star`, but the field needs a username `string` instead.

anthonykhoa:
In our graphql typedefs, the starGiven field of the lessonStatus type is defined as a Star, but the field needs a username string instead. This field is intended to be used for the `GiveStarCard` component as a prop value to pass in -- therefore a username is needed

## This PR will update these files to reflect the change in the `starGiven` field:
* session resolver to return a username `string` instead of a `Star` object for the `starGiven` field
* tests for session resolver
* other tests that were failing after updating the session resolver
* typescript `LessonStatus` type 
* graphql/index file since typedefs file changed
* getApp query

### DIFF
--- a/graphql/typeDefs.ts
+++ b/graphql/typeDefs.ts
@@ -134,7 +134,7 @@ export default gql`
     isTeaching: String
     isEnrolled: String
     starsReceived: [Star]
-    starGiven: Star
+    starGiven: String
   }
 
   type Lesson {
--- a/@types/lesson.d.ts
+++ b/@types/lesson.d.ts
@@ -4,7 +4,7 @@ export type LessonStatus = {
   isTeaching: string | null
   lessonId: string
   starsReceived?: Star[]
-  starGiven?: String
+  starGiven?: string
 }
 
 export type User = {
--- a/@types/lesson.d.ts
+++ b/@types/lesson.d.ts
@@ -4,7 +4,7 @@ export type LessonStatus = {
   isTeaching: string | null
   lessonId: string
   starsReceived?: Star[]
-  starGiven?: Star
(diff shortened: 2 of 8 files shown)


graphql/typeDefs.ts
```diff
@@ -134,7 +134,7 @@ export default gql`
     isTeaching: String
     isEnrolled: String
     starsReceived: [Star]
-    starGiven: Star
+    starGiven: String
```
ggwadera, line 137: Since you'll be return only the username now, wouldn't be better to name this something like `starredUser` or `starredMentor` to avoid confusion? What do you think?

anthonykhoa: I think the name `starGiven` provides more meaning because the word `give` tells us that a user has been sent a `star` to another user

anthonykhoa: Instead of just returning a string, maybe I should return an object containing the username + the rest of the fields in the Star type?

anthonykhoa: Hm.. im not sure if returning an object with all that info would be ideal. Im not able to imagine a scenario where all that info would be useful on the front end after a person has already given star to another person

ggwadera: Alright, that sounds reasonable enough. And yeah, I don't think the IDs would be useful for anything in the component.

ggwadera (approved):

PR #461, opened 2021-01-09
closes #456 - update development setup

### LINKED ISSUE #456: Lessons and challenges tables are empty after fresh install
You need to manually copy data from \__dummy__ folder to have working app, otherwise curriculum page will look like this:
cropped

Ulisseus:
Fixes issue #456.

This PR will add step by step guide on how to configure fully local copy of c0d3 app. It also automatically populates Lesson/Challenge tables with dummy data after fresh install.

### DIFF
--- a/helpers/dbload.ts
+++ b/helpers/dbload.ts
@@ -2,10 +2,13 @@ import { Sequelize } from 'sequelize'
 import { Lesson, LessonTypes } from './models/Lesson'
 import { User, UserTypes } from './models/User'
 import { UserLessonTypes, UserLesson } from './models/UserLesson'
+import { Lesson as LesssonTypescriptTypes } from '../@types/lesson'
+import { Challenge as ChallengeTypescriptTypes } from '../@types/challenge'
 import { SubmissionTypes, Submission } from './models/Submission'
 import { ChallengeTypes, Challenge } from './models/Challenge'
 import { AlertTypes, Alert } from './models/Alert'
 import { StarTypes, Star } from './models/Star'
+import lessonData from '../__dummy__/lessonData'
 
 const sequelize = new Sequelize(
   process.env.DB_NAME || 'you',
@@ -96,6 +99,34 @@ UserLesson.belongsTo(User, {
 
 sequelize.sync({ alter: !!process.env.ALTER_DB })
 
+/* istanbul ignore next */
+if (process.env.NODE_ENV === 'development') {
+  //add data after fresh install
+  Lesson.findAll({
+    limit: 1
+  }).then((lessons: Array<LesssonTypescriptTypes>) => {
+    if (!lessons[0]) {
+      Lesson.bulkCreate(lessonData)
+      console.log('Successfully inserted lesson data')
+    }
+  })
+
+  Challenge.findAll({
+    limit: 1
+  }).then((challenges: Array<ChallengeTypescriptTypes>) => {
+    if (!challenges[0]) {
+      lessonData.map(l => {
+        //need to add lessonID to comply with Challenge schema
+        let validChalleges: Array<ChallengeTypescriptTypes> = []
+        l.challenges.forEach((c: ChallengeTypescriptTypes) =>
+          validChalleges.push({ ...c, lessonId: l.id })
+        )
+        Challenge.bulkCreate(validChalleges)
+      })
+      console.log('Successfully inserted challenge data')
+    }
+  })
+}
 export default {
   Lesson,
   Challenge,
--- a/@types/challenge.d.ts
+++ b/@types/challenge.d.ts
@@ -3,6 +3,7 @@ export type Challenge = {
   id: string
   order: number
   description: string
+  lessonId?: string
 }
 
 export type ChallengeSubmissionData = {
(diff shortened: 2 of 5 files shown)


helpers/dbload.ts
```diff
@@ -96,6 +99,34 @@ UserLesson.belongsTo(User, {
 
 sequelize.sync({ alter: !!process.env.ALTER_DB })
 
+/* istanbul ignore next */
```
Ulisseus, line 102: This will make jest ignore this branch when calculating total coverage, I couldn't figure out how to force istanbul to properly cover it.
If anyone knows how to make this work properly, please comment.

helpers/dbload.ts
```diff
@@ -96,6 +99,34 @@ UserLesson.belongsTo(User, {
...
+if (process.env.NODE_ENV === 'development') {
```
Ulisseus, line 103: I choose this check over `process.env.NODE_ENV != 'production'` since it allows me to isolate dbload.test from grapql.test (which is too hard to mock properly, at least with my jest skills). For example, both Lesson and Challenge classes are extending sequelize Model, so when mocked both will share the same arguments calls. In some cases test won't even start due to circular references in files.

helpers/dbload.ts
anthonykhoa: We currently ignore files we don't want to cover by adding them into the package.json file:

https://github.com/garageScript/c0d3-app/blob/master/package.json#L120

helpers/dbload.ts
```diff
@@ -96,6 +99,34 @@ UserLesson.belongsTo(User, {
...
+  //add data after fresh install
+  Lesson.findAll({
+    limit: 1
+  }).then((lessons: Array<LesssonTypescriptTypes>) => {
+    if (!lessons[0]) {
+      Lesson.bulkCreate(lessonData)
+      console.log('Successfully inserted lesson data')
```
anthonykhoa, line 110: instead of console.log, we have an internal logger we use:

https://github.com/garageScript/c0d3-app/blob/master/helpers/middleware/logger.ts

helpers/dbload.ts
```diff
@@ -96,6 +99,34 @@ UserLesson.belongsTo(User, {
...
+    if (!challenges[0]) {
+      lessonData.map(l => {
+        //need to add lessonID to comply with Challenge schema
+        let validChalleges: Array<ChallengeTypescriptTypes> = []
+        l.challenges.forEach((c: ChallengeTypescriptTypes) =>
+          validChalleges.push({ ...c, lessonId: l.id })
+        )
+        Challenge.bulkCreate(validChalleges)
+      })
+      console.log('Successfully inserted challenge data')
```
anthonykhoa, line 126: internal logger.

helpers/dbload.ts
```diff
@@ -96,6 +99,34 @@ UserLesson.belongsTo(User, {
 
 sequelize.sync({ alter: !!process.env.ALTER_DB })
 
+/* istanbul ignore next */
+if (process.env.NODE_ENV === 'development') {
+  //add data after fresh install
+  Lesson.findAll({
+    limit: 1
+  }).then((lessons: Array<LesssonTypescriptTypes>) => {
```
anthonykhoa, line 107: This codebase uses `async/await` rather than `.then` to handle promises, can you put your promises into an async await function?

helpers/dbload.ts
```diff
@@ -96,6 +99,34 @@ UserLesson.belongsTo(User, {
...
+    }
+  })
+
+  Challenge.findAll({
+    limit: 1
+  }).then((challenges: Array<ChallengeTypescriptTypes>) => {
+    if (!challenges[0]) {
+      lessonData.map(l => {
+        //need to add lessonID to comply with Challenge schema
+        let validChalleges: Array<ChallengeTypescriptTypes> = []
```
anthonykhoa, line 120: should be `validChallenges` instead of `validChalleges`,
also needs to be declared `const` since its value does not change

ggwadera:
For the database data, I'm not so sure if this is something that should be handled through the app code. By the steps you wrote, there's still some manual work to setup the users and the rest so challenges and review works.

I think a better approach for local development is to have a `.pqsl` file that people can import to have a complete dev setup in their DB. This can be done by generating a dump from a pre-existing database. I can help with this if you want. Another way would be to have a remote dev database we could use (which I think was available for some time, idk).

And since you're already recommending docker for the Mattermost setup, the same could be done for Postgres. And it's quite easy to import a `.psql` file into docker. For example, here's the docker-compose file I use for c0d3 development:

```yaml
version: '3'
services:
  db:
    image: postgres:12-alpine
    ports:
      - 5432:5432
    environment:
      POSTGRES_DB: c0d3
      POSTGRES_USER: c0d3_admin
      POSTGRES_PASSWORD: admin
    volumes:
      - ./c0d3.psql:/docker-entrypoint-initdb.d/schema.postgresql.sql:ro
      - c0d3-db:/var/lib/postgresql/data
volumes:
  c0d3-db:
```

`- ./c0d3.psql:/docker-entrypoint-initdb.d/schema.postgresql.sql:ro` -> this line mounts the psql file that I want to be executed when the database container is started.

This is just my two cents. Feel free to comment with your opinions.

Ulisseus:
I close this PR and open new one instead of overwriting this branch.

PR #465, opened 2021-01-12
closes #456 - update development setup

### LINKED ISSUE #456: Lessons and challenges tables are empty after fresh install

Ulisseus:
Fixes issue #456.

This PR will add step by step guide on how to configure fully local copy of c0d3 app. It comes with preconfigured docker compose (lesson/challenges data, admin user which has completed lesson 0 and is able to review new submissions). 

Potential issues:
No tests for docker images, in theory I could try to start test containers, load dump and compare it to __dummy__  lesson data, but should I? These tests won't be fast. 
I explicitly added port in sequelize options, maybe I am overlooking something but sequelize refuses to work when you provide `host:port` in options. I even went as far as checking my hostfile since I use localhost, but I still couldn't figure it out.


helpers/dbload.ts
```diff
@@ -13,6 +13,7 @@ const sequelize = new Sequelize(
   process.env.DB_PW || 'this',
   {
     host: process.env.DB_HOST || 'city',
+    port: ((process.env.DB_PORT as unknown) as number) || 5432,
```
Ulisseus, line 16: I hope this won't break production.

anthonykhoa: I don't think this will break production since `5432` is the default port for a postgresql database. I'm not sure if we have changed our configurations to use a different port.

However, I think that this `port` key should only be added in when used for development. In case the port for the database we use ever changes, this will ensure that nothing breaks.


```
import { Options, Sequelize } from 'sequelize'
// showing you the import so you can see where I got the Options type

const sequelizeOptions: Options = {
  host: process.env.DB_HOST || 'city',
  logging: false,
  dialect: 'postgres',
  pool: {
    max: 5,
    min: 0,
    acquire: 30000,
    idle: 10000
  }
}
if (process.env.MODE === 'DEV') {
  sequelizeOptions.port = ((process.env.DB_PORT as unknown) as number) || 5432
}
const sequelize = new Sequelize(
  process.env.DB_NAME || 'you',
  process.env.DB_USER || 'failed',
  process.env.DB_PW || 'this',
  sequelizeOptions
)
```

docs/developmentSetup.md
```diff
@@ -53,13 +53,15 @@ sudo chown -R 2000:2000 ./volumes/app/mattermost/
...
+10. Generate a personal access token. Close system console, go to team chat (town square) and click on three horizontal bars near your name *Account settings > Security* , then `generate personal access token`. This value will be used as MATTERMOST_ACCESS_TOKEN.
 
-11. Generate a personal access token. Close system console, go to team chat (town square) and click on three horizontal bars near your name *Account settings > Security* , then `generate personal access token`. This value will be used as MATTERMOST_ACCESS_TOKEN.
+11. Keep in mind that by default mattermost requires very secure passwords: at least 10 characters, at least one uppercase letter, one number and one special symbol. Either use such passwords or edit mattermost config at *docker/docker/volumes/app/mattermost/config/config.json* PasswordSettings and restart the server.
```
ggwadera, line 60: This can be changed in the Mattermost System Console too, might be easier, it's in Authentication > Password.

Might be interesting to provide a pre-configured `config.json` too for convenience.

docs/developmentSetup.md
```diff
@@ -68,14 +70,16 @@ To submit challenges:
...
-3. You can connect to postgres `psql -U *username* -d *dbname*`. If you have psql installed you can do it the easier way `psql -U c0d3_admin -h localhost -p 7000 -d c0d3`
+To connect to postgres database inside container:
+1. Check your current active docker containers `docker ps`. Notice docker_postgres container id (first few characters are enough).
+2. Open bash console inside postgres container `docker exec -it *postgres id* /bin/bash`.
+3. Connect to postgres `psql -U c0d3_admin -d c0d3`. 
+
+If you have psql installed locally you can type `psql -U c0d3_admin -h localhost -p 7000 -d c0d3` with the same result.
 
-If you want to pass further lessons as admin you need to manually edit database (since no one can review your submissions). Connect to postgres container (see above) and type:
+If you want to pass further lessons as an admin you need to manually edit database (since no one can review your submissions). Connect to postgres container and type:
```
ggwadera, line 80: I think it would be better to have an user with all lessons completed already, so we don't need to go over these steps. Could be the `admin` you already have in the dump.

Maybe include a normal user account as well.

docs/developmentSetup.md
```diff
@@ -7,22 +7,96 @@ To begin we will need to set up the code base on your own machine. To do this we
...
+
+To submit challenges:
+1. Logout `c0d3 logout`.
+2. Login to your local server `c0d3 login --url http://localhost:4000/api/graphql`.
+3. Submit your code `c0d3 submit --url http://localhost:4000/api/graphql`
+It should be visible on your local setup.
+
+To connect to postgres database inside container:
+1. Check your current active docker containers `docker ps`. Notice docker_postgres container id (first few characters are enough).
+2. Open bash console inside postgres container `docker exec -it *postgres id* /bin/bash`.
```
ggwadera, line 75: The container could be named, then the container name can be used instead of the id.

Use [container_name](https://docs.docker.com/compose/compose-file/compose-file-v3/#container_name) in the compose file for that.

ggwadera, line 75: The container could be named, so the container name can be used instead of the id.

To do this add a [container_name](https://docs.docker.com/compose/compose-file/compose-file-v3/#container_name) property in the compose file.

It could be helpful to suggest some graphical database management tools as well, such as [DBeaver](https://dbeaver.io/) or [pgAdmin](https://www.pgadmin.org/).

docs/developmentSetup.md
Ulisseus: >Might be interesting to provide a pre-configured config.json too for convenience.
I thought about that, but is it really the same config for all users regardless of external environment? E.g. won't my premade config break something for people with macs. I can't test this. 
Probably not since it's the whole point of containers but I am not 100% sure.

helpers/dbload.ts
Ulisseus: Better safe than sorry. Done.

docs/developmentSetup.md
Ulisseus: There are two premade users now. `admin:password` which can review any submission and `newbie:password` which has just registered.

docs/developmentSetup.md
Ulisseus: Containers are now named (c0d3_db and mattermost) and there is a mention of pgAdmin and dBeaver in the end.

ggwadera (approved):

helpers/dbload.ts
```diff
@@ -107,4 +112,4 @@ export default {
   sequelize
 }
 
-export { User, UserLesson }
+export { User, UserLesson, options }
```
anthonykhoa, line 115: Do we plan to use the `options` object in the future, or in other places in the codebase? I'm not able to think of a scenario where we would use the `options` object in another file. Seems to me that it is a private variable whose only purpose is to be used as the third parameter when calling the Sequelize function in this file, so exporting 'options' makes things more confusing. It makes me think that the variable is meant to be used in other places in the codebase

Ulisseus: It won't be used in other places at all, I added it as export for easier testing. Otherwise I am not sure how to write tests for dbload without explicitly creating import. 
I could rename it to `testOptions` or something like that.

helpers/dbload.test.js
```diff
@@ -0,0 +1,34 @@
...
+  dialect: 'postgres',
+  pool: {
+    max: 5,
+    min: 0,
+    acquire: 30000,
+    idle: 10000
+  }
+}
+describe('dbload tests', () => {
+  beforeEach(() => {
```
anthonykhoa, line 14: This should work so that you don't have to export `options` from dbload

```
let defaultOptions
beforeEach(() => {
    jest.resetModuleRegistry()
    defaultOptions = {
        host: process.env.DB_HOST || 'city',
        logging: false,
        dialect: 'postgres',
        pool: {
            max: 5,
            min: 0,
            acquire: 30000,
            idle: 10000
        }
        port: ((process.env.DB_PORT as unknown) as number) || 5432
    }
})
```

helpers/dbload.ts
anthonykhoa: I see, thanks for explaining! I could not think of a better way to write the tests either.

anthonykhoa (approved):

ggwadera:
Nice work @Ulisseus, I believe the dev setup will be much easier now.

@all-contributors please add @Ulisseus for code and documentation.

PR #466, opened 2021-01-14
closes #398 - Creation of GiveStarCard component

### LINKED ISSUE #398: [Feature]Bringing give star feature back to life
Complete details on how to make this happen is in this [design doc](https://github.com/garageScript/c0d3-app/wiki/Design-Docs)

anthonykhoa:
closes #398 

Try out this component in the storybook, please let me know if there are any changes to the UI I should make!
https://story-code.khoa.app/iframe.html?id=components-givestarcard--basic&viewMode=story

## This PR will:
* Create `getLessonMentors` dummy values file
* Export the ModalCardProps from ModalCard file
* Create GiveStarCard component
* Make tests and stories for GiveStarCard
* Add a `left arrow` svg to use, as well as an `scss` file for GiveStarCard
* Update the graphql/index.tsx file since `graphql/queries/setStar.ts` was added(`needed for GiveStarCard`)

## Architectural decisions
1) The `StarCard` and `GiveStarCard` component were originally just one component. However, I turned it into two components, because there is an unnecessary function(`handleClose`) and 2 states(`done, mentor`) that are not needed when someone has already given a star to another person. Thus I separated into two components for increased clarity, readability, and organization of code.

2) Another way to update the front-end information after giving someone a star, is to directly update the cache. However, I felt that this would make the `GiveStarCard` component less modular -- if we ever decide to use another query other than `getApp` in order to retrieve the stars given to a user, then the `GiveStarCard` will also be have to updated as well. This is the same thought process I had with using react context to update a parent component's state as well -- it would make the `GiveStarCard` component less modular, and more prone to breaking when changes in the codebase happen. Thus, I settled for having a `setStarGiven` prop that someone can pass in to update the starGiven value that will be passed into the `GiveStarCard` as a prop.

Screen Shot 2021-01-24 at 8 08 59 PM
Screen Shot 2021-01-24 at 8 09 07 PM
Screen Shot 2021-01-18 at 6 11 25 PM
Screen Shot 2021-01-18 at 6 11 47 PM
Screen Shot 2021-01-18 at 6 48 23 PM


components/__snapshots__/GiveStarCard.test.js.snap
```diff
@@ -0,0 +1,43 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`GiveStarCard Component should display \`already given star to\` display when starGiven prop does not equal empty string 1`] = `
+<div
+  aria-hidden="true"
+/>
```
ggwadera, line 6: I think the snapshots weren't generated correctly because every snap only has a div in it.

Look at the [curriculum page snap](https://github.com/garageScript/c0d3-app/blob/master/__tests__/pages/__snapshots__/curriculum.test.js.snap) for example, it contains the entire rendered component.

If this is incorrect it may lead to a false believe that the component is working as intended, when in reality the test is not doing anything.

anthonykhoa: Ah, thanks for the catch!

ggwadera (approved):

PR #464, opened 2021-01-11
Update project NPM dependencies

ggwadera:
In this PR I will be trying to update the project dependencies to their latest versions, and including fixes to breaking changes when needed. This should fix #364 if I'm able to update Next.js too.

After each incremental update, I run the tests to see if nothing broke, and also run the app and storybook on localhost to see if everything is working as it was.

### Note

After this PR is merged, it's recommended to remove the `node_modules` and `.next` folders in your local environment before running `yarn install` to avoid any issues from the updates.

### Needs more work

- All the Apollo dependencies have been merged into a single one: `@apollo/client`. I haven't tried to update it yet, as it may cause a lot of breaking changes that will require a lot of changes and testing. See the [migration guide](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/) for more details.
- Also, the `next-with-apollo` dependency may not be required anymore, see this [example with Apollo from Vercel](https://github.com/vercel/next.js/tree/canary/examples/with-apollo) where Next is integrated with a GraphQL back-end using the Apollo client.
- The `@zeit/next-sass` and `node-sass` packages are not needed anymore because [NextJS has built-in SASS/SCSS support](https://nextjs.org/docs/basic-features/built-in-css-support#sass-support), requiring only the `sass` package. But this requires to import all the style files only in `_app.tsx`, or use component-level SCSS, read the previous link for more details.
- Sequelize 6 caused a lot of type errors when I installed it. Would need more work in the future to get everything in order.

### Removed

- @types/winston: types are included in the package now.
- @types/yup: types are included in the package now.
- @types/markdown-to-jsx: types are included in the package now.

### Added

- [@testing-library/user-event](https://github.com/testing-library/user-event)
  - This package makes some improvements to the `fireEvent` API, better simulating real user actions.
  - Used in the tests for the signup and login pages.

### Updated

- [Storybook](https://storybook.js.org/) ([Changelog](https://github.com/storybookjs/storybook/blob/next/CHANGELOG.md))
- [Typescript](https://www.typescriptlang.org/) ([Changelog](https://www.typescriptlang.org/docs/handbook/release-notes/overview.html))
- [ESLint](https://eslint.org/) ([Changelog](https://github.com/eslint/eslint/blob/master/CHANGELOG.md))
- [Prettier](https://prettier.io/) ([Changelog](https://github.com/prettier/prettier/blob/master/CHANGELOG.md))
[...]

PR #471, opened 2021-01-20
closes#467 - improve error screens

Ulisseus:
This PR adds one new component - generic [Error](https://www.chromatic.com/component?appId=600846bf978ae90021e98307&name=Components%2FError&mode=snapshot&buildNumber=1&specName=Not+Found&viewport=1200&browser=chrome). 

- It fixes dynamic routing. Right now all links like [https://www.c0d3.com/review/1000Test](https://www.c0d3.com/review/1000Test) are valid. This PR restricts them to lessonID routes (added new type and type guard function).
-  Unauthorized access to [https://www.c0d3.com/curriculum](https://www.c0d3.com/curriculum) should redirect to login page, otherwise it should show 500 Error component. 
-  `No data...' in withQueryLoader wrapper is replaced by new custom Error component. 
- Updated snapshots.


helpers/isLessonID.tsx
```diff
@@ -0,0 +1,4 @@
+import { LessonID } from '../@types/lesson'
+export default function isLessonID(arg: LessonID | string): arg is LessonID {
+  return ['1', '2', '3', '4', '5', '24', '25', '27', '28', '29'].includes(arg)
```
anthonykhoa, line 3: Lesson Ids should not be hardcoded, because if any lesson are added in the future, this file will have to be updated.

__tests__/pages/curriculum.test.js
```diff
@@ -139,4 +142,20 @@ describe('Curriculum Page', () => {
...
+          errors: [new GraphQLError("Cannot read property 'id' of null")]
+        }
+      }
+    ]
+    const { container } = render(
+      <MockedProvider mocks={mocks} addTypename={false}>
+        <Curriculum />
+      </MockedProvider>
+    )
+    await wait(() => expect(container).toMatchSnapshot())
```
anthonykhoa, line 159: `wait` is deprecated, and `waitFor` should be used instead

https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

pages/review/[lesson].tsx
```diff
@@ -33,6 +33,10 @@ const Review: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
   const { lessons, session } = queryData
   const router = useRouter()
   const currentlessonId = router.query.lesson as string
+  if (!isLessonID(currentlessonId)) {
```
Comment (anthonykhoa) [inline pages/review/[lesson].tsx:36]: There is a `validateLessonId` function that has already been made

https://github.com/garageScript/c0d3-app/blob/master/helpers/validateLessonId.ts

Comment (anthonykhoa) [reply pages/review/[lesson].tsx:36]: hmm this is on the front-end though, so we shouldn't use the function i mentioned.

I believe all the `lessonIds` should be contained in the `session`, so you can use that in order to determine if the current route is a valid lessonId or not

__tests__/pages/curriculum.test.js
Ulisseus: `waitFor` requires more modern version of testing-library (c0d3-app uses ^9.4.0). Upgrading it to the latest 11something version breaks all snapshots, so it probably needs to be a separate pr for this.

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx]: There was lesson info in session indeed. I removed unnecessary lessonID type.

__tests__/pages/curriculum.test.js
ggwadera: update your fork, PR #464 was merged and updated a lot of deps

Ulisseus: Updating local branch causes some weird errors in tests `TestingLibraryElementError: Unable to find role="heading"` even if I completely override local files when merging. Quick search gives this [issue](https://github.com/testing-library/dom-testing-library/issues/578) but according to it this problem should be fixed in modern versions of testing library. 

I'll try to figure it out today but closing this pr and opening the new one from synced master could be faster in the end.

Ulisseus:
I'm closing this pr in favor of creating new one instead of trying to solve npm/yarn conflict between dependencies on merging.

PR #472, opened 2021-01-22
closes#467 - improve error screens

Ulisseus:
Unfortunately, I needed to create a new pr due to npm/yarn conflict when merging with local branch.

This PR adds one new component - generic [Error](https://www.chromatic.com/component?appId=600846bf978ae90021e98307&name=Components%2FError&buildNumber=4). 

- It fixes dynamic routing. Right now all links like [https://www.c0d3.com/review/1000Test](https://www.c0d3.com/review/1000Test) and https://www.c0d3.com/curriculum/random are valid. This PR restricts them to lessonID routes.
-  Unauthorized access to [https://www.c0d3.com/curriculum](https://www.c0d3.com/curriculum) should redirect to login page, otherwise it should show 500 Error component. 
-  `No data...' in withQueryLoader wrapper is replaced by new custom Error component. 
- Test for challenge page under `__tests__/[lesson].test.js` was renamed to `challengePage.test.js` since jest could not recognize its path (`jest test __tests__/[lesson].test.js` could not find any tests, `jest test __tests__/*` worked fine)
- Updated snapshots.


__tests__/pages/curriculum.test.js
```diff
@@ -39,13 +42,13 @@ describe('Curriculum Page', () => {
       }
     ]
 
-    const { findByRole } = render(
+    const { findByText } = render(
```
Ulisseus, line 45: I know that testing-library recommends using findByRole/getByRole first, but since my tree looks like this only text queries work:
names

components/Error.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <Layout>
+      <div className="container">
+        <div className="row">
+          <div className="d-flex col-sm-3 align-items-center justify-content-center">
+            <div className="text-center">
+              <Text size="xl" bold={true}>
+                {title}
+              </Text>
+              <Text size="md">{message}</Text>
+              <Navlink path="/" className="btn btn-primary py-3 px-5 mt-3">
```
anthonykhoa, line 21: We have a button component made already
https://github.com/garageScript/c0d3-app/blob/master/components/theme/Button.tsx

pages/curriculum.tsx
```diff
@@ -13,8 +15,32 @@ export const Curriculum: React.FC<GetAppProps> = ({ data }) => {
...
+  }
+  if (!session || !lessons || !alerts) {
+    return (
+      <Error
+        title="Internal server error"
+        message="Bad data"
+        src="/500.png"
+      ></Error>
+    )
+  }
```
anthonykhoa, line 43: Instead of two `if statements`, you can use just one in order to reduce duplicated code
```
if (error || !session || !lessons || !alerts) {
    return (
      <Error
        title="Internal server error"
        message={error && error.message || "Bad data"}
        src="/500.png"
      ></Error>
    )
  }
```

pages/review/[lesson].tsx
```diff
@@ -43,6 +43,10 @@ const Review: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
     router.push('/login')
     return null
   }
+  if (!lessons.map((l: Lesson) => l.id.toString()).includes(currentlessonId)) {
```
Comment (anthonykhoa) [inline pages/review/[lesson].tsx:46]: Can we use a better variable name than `l`?

Is the `toString` needed too? In our typedefs, we define the `id` field as a string, so it should be a string already
https://github.com/garageScript/c0d3-app/blob/master/graphql/typeDefs.ts#L130

components/Error.tsx
```diff
@@ -0,0 +1,34 @@
...
+                Back
+              </Navlink>
+            </div>
+          </div>
+          <div className="col-sm-9">
+            <img src={src} alt="" className="mb-3 img-fluid"></img>
```
anthonykhoa, line 27: The `alt` attribute is there but it's an empty string -- should either get rid of it or give it value

pages/review/[lesson].tsx
Comment (anthonykhoa) [reply pages/review/[lesson].tsx:46]: `if (!lessons.map((l: Lesson) => l.id.toString()).includes(currentlessonId))`

I don't think the line above is doing what you want it to. `map` function returns an array, so it's never going to be truthy, it can only be falsy.

This use of the `find` function should return `undefined` if the lessonId is not found: 
```
lessons.find((lesson)=>lesson.id === currentlessonId)
```

containers/withQueryLoader.tsx
```diff
@@ -23,8 +24,13 @@ const withQueryLoader = <T extends {}>(
...
     return <Component queryData={data} {...props} />
   }
-
-  return <h1>No Data...</h1>
+  return (
+    <Error
+      title="Internal server error"
+      message="No data"
+      src="/500.png"
+    ></Error>
```
anthonykhoa, line 32: `</Error>` should not be needed, you can just use a self enclosing tag since the component does not display children. 

It is in our engineering practices:
https://github.com/garageScript/c0d3-app/wiki/Engineering-Practices

pages/404.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react'
-import Error from 'next/error'
+import Error from '../components/Error'
 
 export default function NotFound() {
-  return <Error statusCode={404} />
+  return (
+    <Error title="404 error" src="/404.png" message="Page not found"></Error>
```
anthonykhoa, line 6: self enclosing tag

pages/curriculum.tsx
```diff
@@ -13,8 +15,32 @@ export const Curriculum: React.FC<GetAppProps> = ({ data }) => {
...
+    error &&
+    _.get(error, 'graphQLErrors[0].message') ===
+      "Cannot read property 'id' of null"
+  ) {
+    Router.push('/login')
+    return <LoadingSpinner />
+  }
+  if (error) {
+    return (
+      <Error
```
anthonykhoa, line 28: self enclosing tag

components/Error.tsx
Ulisseus: I remade Error component with Button (see storybook above). It looks a bit small on big screen but neither Button not Text components allow bigger padding/margins.

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx]: I redirected properly when I tested locally. First map returned an array and then includes checked if currentlessonId is in this array. But I agree `!lessons.find((lesson: Lesson) => lesson.id.toString() === currentlessonId)` is probably more readable. Updated.

components/Error.tsx
Ulisseus: Fixed (and all other self enclosing tags).

pages/review/[lesson].tsx
Comment (anthonykhoa) [reply pages/review/[lesson].tsx]: `l.id.toString())` <-- ah, thanks for explaining! i indeed read the function wrong because I didnt notice this extra `)`

pages/curriculum/[lesson].tsx
```diff
@@ -15,10 +16,20 @@ import _ from 'lodash'
...
-  const lessonStatus: LessonStatus[] = _.get(session, 'lessonStatus', [])
   const router = useRouter()
   const currentlessonId = router.query.lesson as string
+  if (!session || !lessons || !alerts) {
+    return (
+      <Error title="Internal server error" message="Bad data" src="/500.png" />
+    )
+  }
+  if (
+    !lessons.find((lesson: Lesson) => lesson.id.toString() === currentlessonId)
```
Comment (anthonykhoa) [inline pages/curriculum/[lesson].tsx:27]: `lesson.id.toString()` is the `toString` needed? In our typedefs, `id` is defined as a string already, so id should already be a string
https://github.com/garageScript/c0d3-app/blob/master/graphql/typeDefs.ts#L130

components/Error.tsx
anthonykhoa: Our button component does seem too limited for the context of the error component. You could use a `div` to wrap the button in order to add margin, and also wrap the text inside with an html element(`like h1`) to change the font size. But then the size of the button would still be too small even with the largest size available. Considering we would have to make an `xl` class for the button component too, I think the code would be more simple and the UI would look better with the Navlink component!

Ulisseus: OK, I reverted it back to NavLink, see storybook.

pages/curriculum/[lesson].tsx
Comment (Ulisseus) [reply pages/curriculum/[lesson].tsx:27]: It's a bug, thanks for noticing.
 Lesson was typed twice, first as gql query in your link and then as `Lesson` type in https://github.com/garageScript/c0d3-app/blob/master/%40types/lesson.d.ts. Since app.d.ts [uses](https://github.com/garageScript/c0d3-app/blob/master/%40types/app.d.ts) latter Lesson type with number id I need to convert it to string first otherwise I get typerror. Meanwhile actual query comes with string ids. 
I changed Lesson id type to string and removed all `toString` conversions.

stories/components/Error.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { MockedProvider } from '@apollo/react-testing'
+import GET_APP from '../../graphql/queries/getApp'
+
+export default {
+  components: Error,
+  title: 'Components/Error'
+}
+
+export const NotFound: React.FC = () => {
+  const mocks = [
```
anthonykhoa, line 12: are all the mocks necessary? It seems that the `Error` component does not depend on any data received from a graphql request

pages/curriculum/[lesson].tsx
Comment (anthonykhoa) [reply pages/curriculum/[lesson].tsx]: Thanks for fixing the bug!
Blah, I think it is confusing to have `ids` as strings instead of numbers.. but all the ids in our typedefs file are defined as strings, even though they are an integer in the database. I find it confusing -- I think that our typedefs should match the data types of what is defined in our database for increased clarity.

pages/curriculum/[lesson].tsx
```diff
@@ -15,12 +16,20 @@ import _ from 'lodash'
...
-  const userSubmissions: UserSubmission[] = _.get(session, 'submissions', [])
-  const lessonStatus: LessonStatus[] = _.get(session, 'lessonStatus', [])
   const router = useRouter()
   const currentlessonId = router.query.lesson as string
+  if (!session || !lessons || !alerts) {
+    return (
+      <Error title="Internal server error" message="Bad data" src="/500.png" />
+    )
+  }
+  if (!lessons.find((lesson: Lesson) => lesson.id === currentlessonId)) {
```
Comment (anthonykhoa) [inline pages/curriculum/[lesson].tsx:26]: `lessons.find((lesson: Lesson) => lesson.id === currentlessonId)` 

This line is repeated on line 32. You can clean this up by declaring `currentLesson` before this `if` statement.
```
const currentLesson: Lesson | undefined = lessons.find(
    (lesson: Lesson) => lesson.id === currentlessonId	
)
if (!currentLesson) {
         ........
}
```

pages/review/[lesson].tsx
```diff
@@ -40,8 +41,12 @@ const Review: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
     return <LoadingSpinner />
   }
   if (!session) {
-    router.push('/login')
-    return null
+    return (
+      <Error title="Internal server error" message="Bad data" src="/500.png" />
+    )
+  }
+  if (!lessons.find((lesson: Lesson) => lesson.id === currentlessonId)) {
```
Comment (anthonykhoa) [inline pages/review/[lesson].tsx:48]: see comment in `curriculum/[lesson]`, code is repeated here too

stories/components/Error.stories.tsx
Ulisseus: They were unnecessary indeed. Fixed.

pages/curriculum/[lesson].tsx
Comment (Ulisseus) [reply pages/curriculum/[lesson].tsx]: I removed duplicate code in both cases.

stories/components/Error.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { MockedProvider } from '@apollo/react-testing'
+
+export default {
+  components: Error,
+  title: 'Components/Error'
+}
+
+export const NotFound: React.FC = () => {
+  return (
+    <MockedProvider addTypename={false}>
```
anthonykhoa, line 12: the `mockedprovider` shouldnt be needed since the `Error` component doesn't depend on any graphql related things

pages/review/[lesson].tsx
```diff
@@ -40,19 +41,22 @@ const Review: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
     return <LoadingSpinner />
   }
   if (!session) {
-    router.push('/login')
-    return null
+    return (
```
Comment (anthonykhoa) [inline pages/review/[lesson].tsx:44]: When there's no session that means the user is not logged in, so I think it would be a better `user experience` for people to be redirected to the login screen instead of receiving an `Error` display.

__tests__/pages/[lesson].test.js
```diff
@@ -1,183 +0,0 @@
-import React from 'react'
```
Comment (anthonykhoa) [inline __tests__/pages/[lesson].test.js:1]: why is this test file deleted?

__tests__/pages/curriculum.test.js
hwong0305: Please add this as a comment to the code. You don't need to include the image but explain the reasoning.

containers/withQueryLoader.tsx
```diff
@@ -23,8 +24,9 @@ const withQueryLoader = <T extends {}>(
   if (data) {
     return <Component queryData={data} {...props} />
   }
-
-  return <h1>No Data...</h1>
+  return (
```
hwong0305, line 27: If it's no data that means the user is not signed in. We should redirect to sign in form.

package.json
```diff
@@ -92,6 +92,7 @@
     "babel-plugin-inline-react-svg": "^1",
     "babel-plugin-prismjs": "^2",
     "babel-preset-react-app": "^10",
+    "chromatic": "^5",
```
hwong0305, line 95: Why is this needed?

pages/review/[lesson].tsx
Comment (hwong0305) [reply pages/review/[lesson].tsx:44]: This same comment applies to the curriculum page when no data is shown.

package.json
Ulisseus: It's a module for publishing storybook online (see Error component in the first comment). While it's possible to self host storybooks many people don't have their own server, so I used and added chromatic.

__tests__/pages/curriculum.test.js
Ulisseus: Is it really necessary? The code above is rather self explanatory and it's a pretty common pattern in tests (render, query, compare to snapshot). I provided this screenshot to show that I used this query not because it's easier to use but because it's the only way to find correct components. 
If anything I need to learn more about accessibility and change `Error` component to allow proper naming.

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx:44]: No session redirects to login page now.

containers/withQueryLoader.tsx
Ulisseus: It redirects to login page now.

Ulisseus:
This PR got too big, I'm splitting it in two.

PR #475, opened 2021-01-25
closes #468 - migrate to apollo client 3

### LINKED ISSUE #468: Migrate to Apollo Client 3.0
All the Apollo Client packages have been merged into a single one for version 3. This means that all the packages we use from Apollo are now included within `@apollo/client`. Fortunately, they have published a [migration guide](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/), where it says that there have been breaking changes from version 2. While there's a chance that nothing breaks on our end, I opted to leave this update out of the previous updates PR (#464) because it may involve a little more work.

On this same subject, I noticed we use [next-with-apollo](https://github.com/lfades/next-with-apollo) to provide a HOC for our components. While this is useful, it might also be not necessary, as I found this [example with Apollo from Vercel](https://github.com/vercel/next.js/tree/canary/examples/with-apollo) where Next is integrated with a GraphQL back-end using the Apollo client.

### To do

- [ ] Remove all current Apollo client packages
- [ ] Install only `@apollo/client`
- [ ] Update all the Apollo client imports
- [ ] Check if there's breaking changes (tests, type-check, check everything on localhost, etc...)
  - [ ] Tests
  - [ ] Type-check
  - [ ] Check app on localhost to see if it's working correctly
  - This includes checking if `next-with-apollo` is working correctly, as there is an [open issue](https://github.com/lfades/next-with-apollo/issues/147) about supporting Apollo Client v3.
- [ ] If something breaks, fix it

Ulisseus:
closes #468 

In newer version many apollo packages were merged into one. This pr is mostly about using proper imports.
Locally I tested creating new users, submitting and reviewing challenges.


components/ChallengeMaterial.tsx
```diff
@@ -304,8 +304,10 @@ const ChallengeMaterial: React.FC<ChallengeMaterialProps> = ({
     },
     {}
   )
+  //chain sort-map causes errors in tests without a copy
+  const challengesCopy = [...challenges]
   //create a new Challenges array with user submission data integrated and sorted
-  const challengesWithSubmissionData: ChallengeSubmissionData[] = challenges
+  const challengesWithSubmissionData: ChallengeSubmissionData[] = challengesCopy
```
Ulisseus, line 310: I am surprised why this didn't caused any problems before. I created copy of challenges to use in map-sort, otherwise it was causing type errors.

helpers/withApollo.ts
```diff
@@ -16,6 +16,8 @@ export default withApollo(
       // Cache is used to rehydrating the store for Server Side Rendering.
       // Apollo Explanation https://www.apollographql.com/docs/react/performance/server-side-rendering/
       // Example from : https://medium.com/swlh/create-a-killer-frontend-for-2020-setup-next-js-graphql-styled-components-typescript-and-ssr-fe66cffd7d94
+      //next-with-appollo doesn't support apollo3 yet, so some types are not up to date
+      //@ts-ignore
```
Ulisseus, line 20: next-with-apollo wasn't updated for apollo 3 yet. It works fine but some types are wrong, hence ignore.

helpers/withApollo.ts
```diff
@@ -1,7 +1,7 @@
 import withApollo from 'next-with-apollo'
-import { ApolloClient } from 'apollo-client'
-import { createHttpLink } from 'apollo-link-http'
-import { InMemoryCache } from 'apollo-boost'
+import { ApolloClient } from '@apollo/client'
```
anthonykhoa, line 2: can we do just 1 import for `ApolloClient, createHttpLink, and InMemoryCache`?

package.json
```diff
@@ -14,15 +14,9 @@
     "generate": "graphql-codegen --config codegen.yml"
   },
   "dependencies": {
-    "@apollo/react-components": "^3",
-    "@apollo/react-hoc": "^3",
-    "@apollo/react-hooks": "^3",
-    "@apollo/react-testing": "^3",
+    "@apollo/client": "^3.3.7",
```
anthonykhoa, line 17: Since we are using whole number `^N` versions in the rest of our package.json, can you turn this into version into a whole number?

pages/_app.tsx
```diff
@@ -1,10 +1,10 @@
+import { ApolloProvider } from '@apollo/client'
 import React, { useEffect } from 'react'
 // TODO: type the posthog library
 // @ts-ignore
 import posthog from 'posthog-js'
-import { ApolloProvider } from '@apollo/react-hooks'
 import withApollo from '../helpers/withApollo'
-import { ApolloClient, NormalizedCacheObject } from 'apollo-boost'
+import { ApolloClient, NormalizedCacheObject } from '@apollo/client'
```
anthonykhoa, line 7: can we move `ApolloProvider, ApolloClient, and NormalizedCacheObject` to be imported from the same line?

anthonykhoa (approved):

helpers/withApollo.ts
hwong0305: Wouldn't it better to wait until it's official supported before we use in prod? Implementing new library updates early means that we're doing the testing for other people. 

It may not be a good idea for new students to be exposed to this side of development.

Ulisseus: It is possible to remove `next-with-apollo` completely and use next.js fetching methods with graphql as mentioned in the linked issue but it will require a lot of work. On the other hand, `next-with-apollo` only uses apollo for types and tests [according](https://github.com/lfades/next-with-apollo/pull/132#issuecomment-702773588) to its creator.

Ulisseus:
I accidentally submitted some commits as GH Pages Bot, so if anyone decides to merge this pr please squash it into Ulisseus and not to the bot.

ggwadera (approved):

helpers/withApollo.ts
hwong0305: That is fine. I am not asking to remove `next-with-apollo`. 

I think it's better to delay the implementation with apollo3 until `next-with-apollo` is updated so we don't have any workaround.  Are we getting any benefit out of migrating to apollo3?

ggwadera: `next-with-apollo` won't be getting a update any time soon I think, since there's a PR opened 9 months ago to support Apollo 3. Even the maintainer, who works at Vercel, recommends using the next.js fetch methods now instead of the module HOC.

Besides that I believe we would just be delaying the inevitable as things move fast nowadays. Maybe it's better to be up to date now or we might need to spend more time migrating stuff later to support new features.

Would like to hear everyone's opinion on this.

hwong0305: It would have been more clear if the benefits were explained from the start:

- Apollo Client 3 has already been out for six months and `next-with-apollo` has no priority in being replaced soon. It would be worth to implement client 3 due to the improvements below from Apollo 3
- The app can be refactored to import from one package. It makes it easier for new users to understand.
- The improvements to the cache makes implementation a lot more clean and way less cumbersome. 

For next time, when requesting to upgrade to a new version, it would be helpful to list the benefits so other people can understand the need to go to a new version.

hwong0305 (approved):

helpers/withApollo.ts
ggwadera: Yeah that's a fair point, will keep that in mind 👍

anthonykhoa: @songz

Issue #467, opened 2021-01-15
[Feature] Better error screens


Ulisseus:
Right now when you open  [https://www.c0d3.com/curriculum](https://www.c0d3.com/curriculum) without signing up first you get this basic error screen:
![curriculum](https://user-images.githubusercontent.com/3845012/104737911-ea51d400-5755-11eb-9f9d-ea6478cc1dce.png)
Review error screen [https://www.c0d3.com/review/2](https://www.c0d3.com/review/2):
![reviewError](https://user-images.githubusercontent.com/3845012/104738433-8ed41600-5756-11eb-850a-14d509e0ba3a.png)
And 404 error screen is probably the default one from next.js:
![404](https://user-images.githubusercontent.com/3845012/104738888-233e7880-5757-11eb-84bc-673524020191.png)

ggwadera:
Closed by #484 and #485.

PR #484, opened 2021-02-02
add Error component

Ulisseus:
This PR adds new [Error](https://www.chromatic.com/component?appId=600846bf978ae90021e98307&name=Components%2FError&buildNumber=5) component. Part of #467.


components/Error.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import Layout from './Layout'
+import NavLink from './NavLink'
+import { Text } from './theme/Text'
+type ErrorProps = {
+  title: string
+  src: string
+  message?: string
+}
+const Error: React.FC<ErrorProps> = ({ title, message, src }) => {
```
anthonykhoa, line 10: Does the `message` and image `src` have a relationship with one another? Like, every time message equals `No data`, do we want to display the 500 error image?

Similarly, when the message equals `404 error`, do we want to `always` display the `404` error image? If we have images that are `only` used for specific error messages or titles, then I don't think we should pass in `src` as a prop. Instead, you can have some kind of table where you select the `src` based on the `message` or `title` passed in. 

If it is the case that the image `src` is not bound to any error message or title(`like display 404 error image when message equals 404 error`), then I think it would be better to have `src` have only a limited number of options(`notFound, 404`) as a valid input. And instead of `src` it should be renamed something like `type`. A table can be made to match the corresponding image `src` depending on value of `type` passed in. This would make the component easier for others to use in the future, since they won't have to determine the correct paths to use as a value for `src`.

Ex
```
const imageSources = {
  notFound: 'notFound.png',
  404: '404.png'
}
```

Ulisseus: Titles and background images are tied together. `Page not found` is shown with a confused man, `Internal server error` results in generic server image, so it makes sense to merge them in a new type somehow. I am not sure if enum is the best way to do so though, why not a simple union?
```
 type ErrorProps = {
  type: '404' | '500'
  message?: string
}
const Error: React.FC<ErrorProps> = ({ type, message }) => {
       ...
              <Text component="div" size="xl" bold={true}>
                {type === '404' ? 'Page not found' : 'Internal server error'}
              </Text
              ....
              ```

Comment (anthonykhoa) [reply components/Error.tsx:10]: Yep, unions look like they work too! I apologize for the misunderstanding, I should have been more clear in my language. I only used `enums` as an example, which is why I said `something like this`

Comment (Ulisseus) [reply components/Error.tsx]: OK, I changed it to union type. If we need more error pages we could extend this union and change ternary operator for a switch.

**Review by ggwadera — APPROVED** [review]

File: components/Error.tsx [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+              </Text>
+              <Text size="md">{message}</Text>
+              <NavLink path="/" className="btn btn-primary py-3 px-5 mt-3">
+                <h3 className="font-weight-bold">Back</h3>
+              </NavLink>
+            </div>
+          </div>
+          <div className="col-sm-9">
+            <img
+              src={type === '404' ? '/404.png' : '/500.png'}
```
anthonykhoa, line 28: Based on the choices for `type`, the `image src paths` shouldn't have to be hardcoded.
```
src={`/${type}.png`}
```

components/Error.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+const Error: React.FC<ErrorProps> = ({ type, message }) => {
+  return (
+    <Layout>
+      <div className="container">
+        <div className="row">
+          <div className="d-flex col-sm-3 align-items-center justify-content-center">
+            <div className="text-center mt-3">
+              <Text component="div" size="xl" bold={true}>
+                {type === '404' ? 'Page not found' : 'Internal server error'}
```
anthonykhoa, line 18: I see a high potential for someone else updating this component in the future, to add another type of error(`for example, admin pages can have an unauthorized error for non-admins`).

Therefore, I think this piece of logic can be refactored so that it is easier for others to add another error type in the future.

Right now, the next person to add onto this file may experience these difficulties:
1) Having to visually filter through `html` code to determine where to edit in the file
  * `Layout -> div -> div -> div -> div -> Text`: Have to dig all the way here to find the logic
  
2) Figuring out that `500` is the error number associated with `Internal server error`
  * With a ternary operator, it's not explicitly clear that `Internal server error` is the intended title for error `500`. A person would have to analyze the entire component as a whole, and then make the connection of `Oh, there's only one other error type, and it is error 500. So I guess Internal server error is the title for 500`.
  
3) Figuring out the best way to add a third option
  * Continuing to use a ternary operator to support a third error type makes the code unnecessarily complex and more unreadable. The next person to add another type to this file would have to go through the process of deciding how to best support a third error type.

ggwadera: Your initial enum suggestion would be the best approach here in my opinion. With a enum we can directly map an error code to it's associated image, not even needing a ternary operator or switch. And it's also easy to modify in the future by adding new error status codes.

```ts
enum ErrorStatusCode {
	NOT_FOUND = '404',
	INTERNAL_SERVER_ERROR = '500'
}
```

Ulisseus: OK, I changed it to enum. It's easier to add new pages this way, on the other hand, using it would be easier with unions IMO since you don't need to export (and figure out extra type). But it's such a minor thing so I'm fine with either.

components/Error.tsx
```diff
@@ -2,20 +2,27 @@ import React from 'react'
...
+  INTERNAL_SERVER_ERROR = '500'
+}
 type ErrorProps = {
-  type: '404' | '500'
+  code: StatusCode
   message?: string
 }
-const Error: React.FC<ErrorProps> = ({ type, message }) => {
+const Error: React.FC<ErrorProps> = ({ code, message }) => {
+  const errorMessages: ReadonlyMap<StatusCode, string> = new Map([
```
anthonykhoa, line 14: This hash table should be on the global scope. The table has constant values that will not change unless we manually update the file ourselves to add another error status code. As such, there is no need to generate another `errorMessages` table every time this component is used.

components/Error.tsx
anthonykhoa: Hmm I think using the `map` object makes things more complex than they have to be. `Map` is meant to be used when we have to key memory addresses such as objects, but all the status codes are just strings.

This should work:

```
const errorMessages: Readonly<{ [key in StatusCode]: string }> = {
  [StatusCode.NOT_FOUND]: 'Page not found',
  [StatusCode.INTERNAL_SERVER_ERROR]: 'Internal server error'
}
``

Comment (Ulisseus) [reply components/Error.tsx]: Makes sense, I thought that map was faster than object for accessing values (turns out it is faster but for inserting/removing key-values pairs). It's an object now.

File: components/Error.tsx (hunk 4 above)
Comment (Ulisseus) [reply components/Error.tsx]: Mapping from status codes is out of component scope and it's renamed to errorTitle.

File: stories/components/Error.stories.tsx [hunk 5]
```diff
@@ -0,0 +1,49 @@
...
+  ]
+  return (
+    <MockedProvider mocks={mocks} addTypename={false}>
+      <Error code={StatusCode.NOT_FOUND} message="404 error" />
+    </MockedProvider>
+  )
+}
+
+export const Internal: React.FC = () => {
+  const mocks = [
```
anthonykhoa, line 32: I think this file can be refactored so that there is less duplicated code. For example, these `mocks` are exactly the same value in each storybook entry(`Internal, NotFound`)

anthonykhoa: This should make it easier for the next person who has to update this file, since they don't have to make their own `mocks`, and they don't have to fill in `MockedProvider` with the correct prop values.

```
const mocks = [
  {
    request: { query: GET_APP },
    result: {
      data: {
        lessons: [],
        session: null,
        alerts: []
      }
    }
  }
]

const ErrorWrapper: React.FC = ({ children }) => (
  <MockedProvider mocks={mocks} addTypename={false}>
    {children as React.ReactElement}
  </MockedProvider>
)

export const NotFound: React.FC = () => (
  <ErrorWrapper>
    <Error code={StatusCode.NOT_FOUND} message="404 error" />
  </ErrorWrapper>
)

export const Internal: React.FC = () => (
  <ErrorWrapper>
    <Error code={StatusCode.INTERNAL_SERVER_ERROR} />
  </ErrorWrapper>
)
```

anthonykhoa: Something similar to this:
```
const mocks = [
  {
    request: { query: GET_APP },
    result: {
      data: {
        lessons: [],
        session: null,
        alerts: []
      }
    }
  }
]

export const NotFound: React.FC = () => (
  <MockedProvider mocks={mocks} addTypename={false}>
    <Error code={StatusCode.NOT_FOUND} message="404 error" />
  </MockedProvider>
)

export const Internal: React.FC = () => (
  <MockedProvider mocks={mocks} addTypename={false}>
    <Error code={StatusCode.INTERNAL_SERVER_ERROR} />
  </MockedProvider>
)
```

components/Error.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import Layout from './Layout'
+import NavLink from './NavLink'
+import { Text } from './theme/Text'
+export enum StatusCode {
+  NOT_FOUND = '404',
```
anthonykhoa, line 6: Why are the `status code numbers` strings? I think it would make more sense to use numbers as numbers, rather than as strings.

Additionally, this removes the need to have a `toStatusCode` function that you have in your PR#485

components/Error.tsx
```diff
@@ -0,0 +1,42 @@
...
+              <Text component="div" size="xl" bold={true}>
+                {errorTitle[code]}
+              </Text>
+              <Text size="md">{message}</Text>
+              <NavLink path="/" className="btn btn-primary py-3 px-5 mt-3">
+                <h3 className="font-weight-bold">Back</h3>
+              </NavLink>
+            </div>
+          </div>
+          <div className="col-sm-9">
```
anthonykhoa, line 34: This `div` is unneeded, you can just add the classname(`col-sm-9`) to the image's classes

anthonykhoa:
The two images should use `svg`'s instead of `png`'s. This is so that the image can easily scale without losing quality

Ulisseus:
> The two images should be `svg`'s because:
> 
>     1. `svg`'s scale easily without losing quality, while `png`'s lose quality upon scaling.
> 
>     2. The rest of the images in our `public` folder are `svg`'s, so for the sake of codebase consistency these two images should also be `svg`'s as well.

Resaving bitmap as svg won't help much. I don't have original source so it will look look as bad as upscaled png. Should I change it just for the sake of consistency? It would be misleading IMO, it's still a bitmap just in different format. I also added used [next/image component](https://nextjs.org/docs/basic-features/image-optimization) which should help with image optimization.

ggwadera:
> Resaving bitmap as svg won't help much. I don't have original source so it will look look as bad as upscaled png. Should I change it just for the sake of consistency? It would be misleading IMO, it's still a bitmap just in different format. I also added [next/image component](https://nextjs.org/docs/basic-features/image-optimization) which should help with image optimization.

You're right that just converting the png won't help. But even the Next image component won't help with scaling issues. Could you search for some another illustrations then? You can find some free resources [here](https://nodesign.dev/illustrations).

Ulisseus:
OK, I changed images to svg, see[ new storybook](https://www.chromatic.com/component?appId=600846bf978ae90021e98307&name=Components%2FError&buildNumber=5&specName=Not+Found).

ggwadera:
Will you be adding tests for this component as well?

Ulisseus:
> Will you be writing unit tests for this component as well?
But what is there to test? Error component has no internal logic, storybook snapshots should cover everything.

public/500.svg
```diff
@@ -0,0 +1,553 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
anthonykhoa, line 1: We shouldn't put all the `error` images into the root of the public folder. To be more organized, they should be placed into a folder of their own. This will make it easier for others to find the `error` images in the future

songz (approved):

hwong0305 (approved):

PR #485, opened 2021-02-02
updates #487 - update pages routes

Ulisseus:
#484 needs to be merged first. Part of #467.

This PR updates page routes. If user is unauthorized he will be redirected to login page. If the page does not exist he will get 404 Error. Graphql error will result in 500 Internal Server Error.


@types/lesson.d.ts
```diff
@@ -12,7 +12,7 @@ export type User = {
 }
 
 export type Lesson = {
-  id: number
+  id: string
```
Ulisseus, line 15: See #486. Such small change doesn't warrant a separate PR.

pages/404.tsx
```diff
@@ -1,6 +1,5 @@
 import React from 'react'
-import Error from 'next/error'
-
+import Error from '../components/Error'
 export default function NotFound() {
-  return <Error statusCode={404} />
+  return <Error title="404 error" src="/404.png" message="Page not found" />
```
hwong0305, line 4: Next.JS has a default 404 page. Please utilize this instead of creating a 404 page.

pages/500.tsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react'
+import Error from '../components/Error'
+
+export default function NotFound() {
```
hwong0305, line 4: A status code of `500` is a server error and isn't `not found`. This function name is misleading.

pages/curriculum/[lesson].tsx
```diff
@@ -15,13 +17,26 @@ import _ from 'lodash'
...
-  const lessonStatus: LessonStatus[] = _.get(session, 'lessonStatus', [])
   const router = useRouter()
   const currentlessonId = router.query.lesson as string
+  if (!session) {
+    router.push('/login')
+    return <LoadingSpinner />
+  }
+  if (!lessons || !alerts) {
+    return (
+      <Error title="Internal server error" message="Bad data" src="/500.png" />
```
Comment (hwong0305) [inline pages/curriculum/[lesson].tsx:28]: Shouldn't this mean the user is not logged in?

pages/curriculum/[lesson].tsx
```diff
@@ -15,13 +17,26 @@ import _ from 'lodash'
...
+    )
+  }
   const currentLesson: Lesson | undefined = lessons.find(
-    (lesson: Lesson) => lesson.id.toString() === currentlessonId
+    (lesson: Lesson) => lesson.id === currentlessonId
   )
+  if (!currentLesson) {
+    return <Error title="404 error" message="Page not found" src="/404.png" />
```
Comment (hwong0305) [inline pages/curriculum/[lesson].tsx:35]: 404 should be reserved for the page is not found. Not an error.

hwong0305 (dismissed):
See comments in PR.

pages/404.tsx
Ulisseus: I don't understand, according to docs next.js [suggests](https://nextjs.org/docs/advanced-features/custom-error-page) to edit `pages/404.tsx` to create custom error page. Do you want me to edit next.js error component instead of creating my own?

pages/curriculum/[lesson].tsx
Comment (Ulisseus) [reply pages/curriculum/[lesson].tsx:28]: No session means that user is not logged in. This case happens if there is a session but no lessons or no alerts.

pages/curriculum/[lesson].tsx
Comment (Ulisseus) [reply pages/curriculum/[lesson].tsx:35]: This case happens if user tries to open lessonID which does not exist, something like https://www.c0d3.com/curriculum/100ShouldNotExist. I could redirect him to 404 page but this way he won't be able to check his incorrect url. 
Do you want me to rename Error component for something else?

pages/500.tsx
Ulisseus: Looks reasonable, will be fixed.

pages/curriculum/[lesson].tsx
Comment (hwong0305) [reply pages/curriculum/[lesson].tsx:28]: Wouldn't it better to redirect to login if this is the case?

pages/404.tsx
hwong0305: That sounds like a much better way and less confusing with the code. For the error page, you can just render a link to login or the landing page.  It's better if you don't pull any data into it since it's statically generated. 

I like the idea of using next.js's error capability. Since it will go to the page whenever there's an error and you don't have to manually route to `/500`

https://nextjs.org/docs/advanced-features/custom-error-page

hwong0305 (commented):
Updating my comment to amend the point on the 404 page.

pages/curriculum/[lesson].tsx
Comment (Ulisseus) [reply pages/curriculum/[lesson].tsx:28]: I think it would  be more confusing. Since there is nothing wrong with session already authorized user could be redirected to login page. He relogins, login redirects him to curriculum page, no lessons/alerts, he is on log in page again. 
This way he'll get a 500 error screen and won't waste time logging in/logging out.

pages/404.tsx
Ulisseus: >Otherwise, users won't be redirected to 404 when they go to c0d3.com/500 for whatever reason.
c0d3.com/500 won't redirect to 404 page, but you're right, I forgot to update _error page. I updated it and it should display new Error screens now too.

pages/500.tsx
Ulisseus: Fixed.

pages/_error.tsx
```diff
@@ -21,14 +25,15 @@ const MyError = ({
...
-  return <NextErrorComponent statusCode={statusCode} />
+  return <ErrorComponent code={toStatusCode(statusCode)} />
 }
 
 MyError.getInitialProps = async ({ res, err, asPath }: MyErrorContext) => {
-  const errorInitialProps = (await NextErrorComponent.getInitialProps({
-    res,
-    err
-  } as MyErrorContext)) as MyErrorContext
+  const statusCode = res ? res.statusCode : err ? err.statusCode : 404
```
anthonykhoa, line 32: why are we setting the default error code to 404?

pages/_error.tsx
```diff
@@ -21,14 +25,15 @@ const MyError = ({
...
+  const errorInitialProps = {
+    statusCode,
+    hasGetInitialPropsRun: false
```
anthonykhoa, line 35: On line 40, you set this value to true. So it's redundant here

songz (approved):

hwong0305 (approved):

ggwadera (approved):

Issue #409, opened 2020-09-25
Write tests for all files and add all relevant files to code coverage


songz:
This PR does not have tests: https://github.com/garageScript/c0d3-app/pull/401/files

And yet, coverage is 100%. This means many other pages probably lack tests as well. 

For this issue, first figure out how to include pages to be tracked in code coverage. 

Then, figure out how to be amazing and get coverage to a more sensible 100%

hwong0305:
Test needs to be written for the files below to covered all the lines before we can add all relevant files for code coverage. Please create individual PRs to close each file and tag this issue and we will cross it off as we go.

----------------------------|---------|----------|---------|---------|-------------------
File                        | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------------------------|---------|----------|---------|---------|-------------------
All files                   |   74.98 |    73.12 |   69.85 |   74.72 |                   
 components                 |   91.95 |    95.41 |    90.7 |   92.43 |                   
  AdditionalResources.tsx   |     100 |      100 |     100 |     100 |                   
  Alert.tsx                 |     100 |      100 |     100 |     100 |                   
  AlertsDisplay.tsx         |     100 |      100 |     100 |     100 |                   
  AnnouncementCard.tsx      |     100 |      100 |     100 |     100 |                   
  AppNav.tsx                |     100 |      100 |     100 |     100 |                   
  Card.tsx                  |     100 |      100 |     100 |     100 |                   
  ChallengeMaterial.tsx     |     100 |      100 |     100 |     100 |                   
  ContributorCard.tsx       |     100 |      100 |     100 |     100 |                   
  DropdownMenu.tsx          |     100 |      100 |     100 |     100 |                   
  FilterButtons.tsx         |     100 |      100 |     100 |     100 |                   
  Footer.tsx                |     100 |      100 |     100 |     100 |                   
  FormCard.tsx              |     100 |      100 |     100 |     100 |                   
  Input.tsx                 |     100 |      100 |     100 |     100 |                   
  InputGroup.tsx            |     100 |      100 |     100 |     100 |                   
  LandingPage.tsx           |     100 |      100 |     100 |     100 |                   
  Layout.tsx                |     100 |      100 |     100 |     100 |                   
  LessonCard.tsx            |     100 |      100 |     100 |     100 |                   
  LessonTitleCard.tsx       |     100 |      100 |     100 |     100 |                   
[...]

PR #489, opened 2021-02-07
add profile/username test

Ulisseus:
Adds test for `profile/username`. Part of #409 and #433. 

There are many tests for null values. Should we tighten up graphql types someday? Almost everything is optional for no reason. 
Also, in [userlesson](https://github.com/garageScript/c0d3-app/blob/master/pages/profile/%5Busername%5D.tsx#L60) file there is a comment about fixing id types and this expression ` status === 'passed' && parseInt(lessonId || '') === parseInt(lesson.id + ''`. Do we still need to cast these types into Ints?

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "@storybook/addon-storysource": "^6",
     "@storybook/addon-viewport": "^6",
     "@storybook/react": "^6",
+    "@testing-library/jest-dom": "^5.11.9",
     "@testing-library/react": "^11",
     "@testing-library/user-event": "^12",
     "@types/bcrypt": "^3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     "@storybook/addon-storysource": "^6",
     "@storybook/addon-viewport": "^6",
     "@storybook/react": "^6",
+    "@testing-library/jest-dom": "^5.11.9",
```
Ulisseus, line 70: I added this optional testing-library module for easier appearance-disappearance testing. See official [docs](https://testing-library.com/docs/guide-disappearance/). Other methods didn't work in some cases (I still got loading spinner when querying with findBy).

ggwadera: The `waitForElementToBeRemoved()` is not from `jest-dom`. You can see the import is from the React testing library.

Nonetheless, it's a useful library which includes some [useful DOM testing helpers](https://github.com/testing-library/jest-dom#jest-dom).

For example, instead of testing `expect(element).toBeTruthy()`, you can test for [`expect(element).toBeInTheDocument()`](https://github.com/testing-library/jest-dom#tobeinthedocument)

ggwadera (approved):

songz (approved):

hwong0305 (approved):

PR #490, opened 2021-02-07
add ReviewCard test and story

Ulisseus:
Adds test and story for ReviewCard component. Part of #409 and #433.


components/ReviewCard.tsx
```diff
@@ -100,48 +100,44 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
   }
 
   return (
-    <>
-      {diff && (
```
Ulisseus, line 104: I deleted this check since it created an untestable branch. Local component `DiffView` already has default case when no diff is provided but since it's called as part of this expression you can't access it.

anthonykhoa: Will this change affect the front end user experience?

components/ReviewCard.test.js
```diff
@@ -0,0 +1,133 @@
...
+      <MockedProvider mocks={mocks}>
+        <ReviewCard submissionData={submissionData} addTypeName={false} />
+      </MockedProvider>
+    )
+    fireEvent.change(getByRole('textbox', { name: '' }), {
+      target: { value: 'Good job!' }
+    })
+    await waitFor(() =>
+      fireEvent.click(getByRole('button', { name: 'Accept' }))
+    )
```
ggwadera, line 130: I changed these two events to the ones below and it seems to fix the issue on my end. Try it to see if it works here too.

```js
//add the import below on the test file
import userEvent from '@testing-library/user-event'

// change fireEvent to these ones
userEvent.type(getByRole('textbox', { name: '' }), 'Good job!')
userEvent.click(getByRole('button', { name: 'Accept' }))
```

components/ReviewCard.test.js
```diff
@@ -0,0 +1,133 @@
...
+    expect(container).toMatchSnapshot()
+  })
+
+  test('Should render no diff input', async () => {
+    const NoDiffSumbisson = _.cloneDeep(submissionData)
+    delete NoDiffSumbisson.diff
+    const { container } = render(
+      <MockedProvider mocks={mocks}>
+        <DiffView submissionData={NoDiffSumbisson} addTypeName={false} />
```
ggwadera, line 80: The `addTypeName={false}` prop should be placed on the `MockedProvider` component.

```jsx
<MockedProvider mocks={mocks} addTypeName={false}>
  <ReviewCard submissionData={{ ...submissionData, diff: NonJsDiff }} />
</MockedProvider>
```

ggwadera (changes requested):

ggwadera:
Running `yarn test --no-cache --runInBand` on my machine resulted in the same error from the CI logs.

components/ReviewCard.test.js
Ulisseus: I thought that userEvent and fireEvent are interchangeable. Thanks, that fixed it.

components/ReviewCard.test.js
Ulisseus: Fixed.

components/ReviewCard.tsx
Ulisseus: It's from an old commit, current pr added one extra export instead of changing component. It should not affect end use either way.

ggwadera (approved):

components/ReviewCard.test.js
ggwadera: I think they're interchangeable, but [`userEvent`](https://testing-library.com/docs/ecosystem-user-event) seems to be a more sophisticated implementation made to approach browser events more realistically. Also the API is a bit easier to use too.

ggwadera:
Whoever merges this PR please squash the commits so we don't have a bunch of `fix build test` in the log.

anthonykhoa (approved):

PR #493, opened 2021-02-10
Fix Give Star feature

### LINKED ISSUE #492: [Bug] GiveStarCard not working in production
The _Give Star_ feature from PR #481 is not working on the production deployed version, while on a development server it works fine.

Sample error from the network response:

```json
{
  "errors": [
    {
      "message": "TypeError: Cannot read property 'username' of undefined",
      "locations": [{ "line": 2, "column": 3 }],
      "path": ["getLessonMentors"],
      "extensions": { "code": "INTERNAL_SERVER_ERROR" }
    }
  ],
  "data": { "getLessonMentors": null }
}
```

ggwadera:
Closes #492

The issue with the linked issue was that the webpack build step replaces the model names in the transpiled script file. Examining the built file, the `User` model was renamed into `User_User` for some reason I'm not aware of. So this resulted in the entity from the sequelize query having a property `User_User` instead of `User`, to get the user entity from the table join.

To solve this, I added an alias to the sequelize association, so the user instance is always called `user`.

What this PR does:

- Adds an alias for the `User` model in the `UserLesson -> belongs to -> User` sequelize association.
- Changes the sequelize query to include the alias and only the required attributes (username, name and id) for the user.
- Fixes the tests that broke because of this.
- Filters out the current user so you don't see your own name in the list.

### DIFF
--- a/helpers/dbload.ts
+++ b/helpers/dbload.ts
@@ -96,7 +96,8 @@ User.hasMany(UserLesson, {
 })
 
 UserLesson.belongsTo(User, {
-  foreignKey: 'userId'
+  foreignKey: 'userId',
+  as: 'user'
 })
 
 sequelize.sync({ alter: !!process.env.ALTER_DB })
--- a/graphql/queryResolvers/getLessonMentors.test.js
+++ b/graphql/queryResolvers/getLessonMentors.test.js
@@ -3,6 +3,7 @@ jest.mock('../../helpers/validateLessonId')
 import { getLessonMentors } from './getLessonMentors'
 import { User, UserLesson } from '../../helpers/dbload'
 import { validateLessonId } from '../../helpers/validateLessonId'
+import { Op } from 'sequelize'
 describe('getLessonMentors resolver', () => {
   beforeEach(() => {
     validateLessonId.mockReturnValue(true)
@@ -49,5 +50,23 @@ describe('getLessonMentors resolver', () => {
     //rejects: checks for promise rejection
     //which would be the case if an error was thrown in a Promise
     await expect(getLessonMentors(null, { lessonId: '3' })).rejects.toThrow()
+  })
+
+  test('should include userId in the query if session exists', async () => {
+    UserLesson.findAll = jest
+      .fn()
+      .mockReturnValue([{ user: { username: 'user1', name: 'lol', id: 2 } }])
+    const res = await getLessonMentors(
+      null,
+      { lessonId: '3' },
+      { req: { user: { id: 1 } } }
+    )
+    expect(res).toEqual([{ username: 'user1', name: 'lol', id: 2 }])
+    expect(UserLesson.findAll).toHaveBeenCalledWith({
+      where: { lessonId: '3', userId: { [Op.ne]: 1 } },
+      include: [
+        { model: User, as: 'user', attributes: ['username', 'name', 'id'] }
+      ]
+    })
   })
 })
--- a/graphql/queryResolvers/getLessonMentors.test.js
+++ b/graphql/queryResolvers/getLessonMentors.test.js
@@ -10,11 +10,10 @@ describe('getLessonMentors resolver', () => {
 
   test('should return an array of User objects', async () => {
     UserLesson.findAll = jest.fn().mockReturnValue([
-      { User: { username: 'user1', email: 'abc@mail', name: 'lol', id: 2 } },
+      { user: { username: 'user1', name: 'lol', id: 2 } },
       {
-        User: {
+        user: {
           username: 'user2',
-          email: 'xyz@mail',
           name: 'potato',
           id: 240
         }
@@ -27,7 +26,9 @@ describe('getLessonMentors resolver', () => {
     ])
     expect(UserLesson.findAll).toHaveBeenCalledWith({
       where: { lessonId: '3' },
-      include: [{ model: User }]
+      include: [
+        { model: User, as: 'user', attributes: ['username', 'name', 'id'] }
+      ]
     })
   })
 
--- a/graphql/queryResolvers/getLessonMentors.ts
+++ b/graphql/queryResolvers/getLessonMentors.ts
@@ -1,5 +1,8 @@
 import { User, UserLesson } from '../../helpers/dbload'
 import { validateLessonId } from '../../helpers/validateLessonId'
+import { Op } from 'sequelize'
+import { Context } from '../../@types/helpers'
+import _ from 'lodash'
 
 type ArgsGetLessonMentors = {
   lessonId: string
@@ -7,14 +10,17 @@ type ArgsGetLessonMentors = {
 
 export const getLessonMentors = async (
   _parent: void,
-  args: ArgsGetLessonMentors
+  args: ArgsGetLessonMentors,
+  context: Context
 ): Promise<User[]> => {
   const { lessonId } = args
+  const userId: number | undefined = _.get(context, 'req.user.id', undefined)
   try {
     await validateLessonId(parseInt(lessonId))
     const results: [UserLesson & { user: User }] = await UserLesson.findAll({
       where: {
-        lessonId
+        lessonId,
+        ...(userId && { userId: { [Op.ne]: userId } })
       },
       include: [
         { model: User, as: 'user', attributes: ['username', 'name', 'id'] }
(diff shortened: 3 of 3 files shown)


anthonykhoa (approved):

anthonykhoa (approved):

helpers/dbload.ts
```diff
@@ -96,7 +96,8 @@ User.hasMany(UserLesson, {
 })
 
 UserLesson.belongsTo(User, {
-  foreignKey: 'userId'
+  foreignKey: 'userId',
```
hwong0305, line 99: Does this mean the test database is different than the production database? If that is the case, that should be fixed too.

ggwadera: No, the database is unaltered because I didn't change the foreign key.

The `as` value is just the internal name that Sequelize will use when defining the associations. By default it uses the model name, but since the build step changes that we need to set an alias to make sure the property name is the same in dev and prod. ([Sequelize guide](https://sequelize.org/v5/manual/associations.html#naming-strategy), [API reference](https://sequelize.org/v5/class/lib/model.js~Model.html#static-method-belongsTo))

hwong0305: I understand what happens in sequelize. I am concerned about why this works in development but not in production.

ggwadera: It breaks in prod because webpack changes the model names so `User` becomes `User_User`.

If you run `yarn build` and look at the transpiled files you'll see this for the `User` model definition:

```js
class User_User extends external_sequelize_["Model"] {}
```

And the `getLessonMentors` gets transpiled to this:

```js
const getLessonMentors = async (_parent, args) => {
  const {
    lessonId
  } = args;

  try {
    await validateLessonId(parseInt(lessonId));
    const results = await dbload["b" /* UserLesson */].findAll({
      where: {
        lessonId
      },
      include: [{
        model: dbload["a" /* User */]
      }]
    });
    return results.map(({
      User: {
        username,
        name,
        id
      }
    }) => {
      return {
        username,
        name,
        id
      };
    });
  } catch (err) {
    throw new Error(err);
  }
};
```

The part that breaks in prod in that the `User` field doesn't exist because it changed to `User_User`.

ggwadera: This will work in kinda of a improvised fix until we can find a better way to make this work more easily.

hwong0305: Since `sequelize` is backend code, it shouldn't be transpired by webpack. What you see above looks like it's code that is generated by `graphql-codegen` 

This tells me that our development database may have a property called `User` in `user`. We should check that and fix this. We possibly may  want to want `alter` in development to correct this.

ggwadera: `graphql-codegen` generates only one file: [`graphql/index.tsx`](https://github.com/garageScript/c0d3-app/blob/master/graphql/index.tsx), and that file only has the types for the resolver ([1](https://github.com/garageScript/c0d3-app/blob/70c5ccbed18bae71f7604e64ca4aaccf9351a5cf/graphql/index.tsx#L198), [2](https://github.com/garageScript/c0d3-app/blob/70c5ccbed18bae71f7604e64ca4aaccf9351a5cf/graphql/index.tsx#L205), [3](https://github.com/garageScript/c0d3-app/blob/70c5ccbed18bae71f7604e64ca4aaccf9351a5cf/graphql/index.tsx#L509), [4](https://github.com/garageScript/c0d3-app/blob/70c5ccbed18bae71f7604e64ca4aaccf9351a5cf/graphql/index.tsx#L1151)). And besides that, it only generates that file if you run the command for it, it's not a build step. Also the graphql types and sequelize model definitions are independent of each other right now.

The code above is generated after running `yarn build` (which is [aliased](https://github.com/garageScript/c0d3-app/blob/70c5ccbed18bae71f7604e64ca4aaccf9351a5cf/package.json#L6) to `next build`) like I already said before. And the code above is from the production-ready files inside the `.next` folder. That specific code is in the file `.next/server/pages/api/graphql.js`, that is generated from [this](https://github.com/garageScript/c0d3-app/blob/master/pages/api/graphql.ts) file. You can run `yarn build` and see for yourself. That is the code that is running on the deployed version on Vercel right now.

Even if it's backend code it's still managed by the Next.js framework, because that is how Next.js [works](https://nextjs.org/docs/api-routes/introduction). I don't know if it's webpack or babel that makes the code like that, but it's not `graphql-codegen`. @Ulisseus also sent a related issue with next-auth and TypeORM, [see here](https://chat.c0d3.com/c0d3/pl/ecnzz4itubywxqe6k31c19zwxh).

The development database has nothing different as well, it's the same columns, I didn't even touch the database. The alias is just for sequelize to put the value in the right property on the resulting JS object. [There's even another aliases already in use](https://github.com/garageScript/c0d3-app/blob/70c5ccbed18bae71f7604e64ca4aaccf9351a5cf/helpers/dbload.ts#L67-L83).

Running the server with `next start` and attaching a debugger, these are the fields on one of the values of `results`:

```js
console.log(JSON.stringify(results[0]))
{
  "isPassed": "1613224106554",
  "isTeaching": "1613224113460",
  "isEnrolled": "1613224116772",
  "lessonId": 5,
  "createdAt": "2021-01-01T02:04:36.339Z",
  "updatedAt": "2021-01-01T11:57:00.101Z",
  "userId": 1,
  "User_User": {
    "id": 1,
    "name": "Name",
    "username": "username",
    "password": "redacted",
    "email": "user@email.com",
    "gsId": null,
    "isOnline": null,
    "isAdmin": "false",
    "forgotToken": null,
    "cliToken": null,
    "emailVerificationToken": null,
    "tokenExpiration": null,
    "createdAt": "2021-01-01T01:24:43.803Z",
    "updatedAt": "2021-01-01T11:46:18.082Z"
  }
}
```

And here is that same value but running with `next dev`:

```js
console.log(JSON.stringify(results[0]))
{
  "isPassed": "1613224106554",
  "isTeaching": "1613224113460",
  "isEnrolled": "1613224116772",
  "lessonId": 5,
  "createdAt": "2021-01-01T02:04:36.339Z",
  "updatedAt": "2021-01-01T11:57:00.101Z",
  "userId": 1,
  "User": {
    "id": 1,
    "name": "Name",
    "username": "username",
    "password": "redacted",
    "email": "user@email.com",
    "gsId": null,
    "isOnline": null,
    "isAdmin": "false",
    "forgotToken": null,
    "cliToken": null,
    "emailVerificationToken": null,
    "tokenExpiration": null,
    "createdAt": "2021-01-01T01:24:43.803Z",
    "updatedAt": "2021-01-01T11:46:18.082Z"
  }
}
```

So you see that the expected `User` field is not present in the production build, instead we have `User_User` because it's the model name that was changed in the files bundled by `next build`.

> This tells me that our development database may have a property called `User` in `user`.

Why would you think that?

Besides trying to fix this on our side, the only other viable solution I see would be to open a issue on the Next.js repo and see what it's developers have to say about that.

hwong0305: It looks like it's babel running through webpack that transforms the code like this.

The **root cause** isn't with the backend code but just with now `Next.JS` wants to compile everything since it's running through the serverless function with `Vercel`.

I was simply worried that our dev environment has `userLesson` schema of below. 

```
{
  "isPassed": "1613224106554",
  "isTeaching": "1613224113460",
  "isEnrolled": "1613224116772",
  "lessonId": 5,
  "createdAt": "2021-01-01T02:04:36.339Z",
  "updatedAt": "2021-01-01T11:57:00.101Z",
  "userId": 1,
  "User_User": {
    "id": 1,
    "name": "Name",
    "username": "username",
    "password": "redacted",
    "email": "user@email.com",
    "gsId": null,
    "isOnline": null,
    "isAdmin": "false",
    "forgotToken": null,
    "cliToken": null,
    "emailVerificationToken": null,
    "tokenExpiration": null,
    "createdAt": "2021-01-01T01:24:43.803Z",
    "updatedAt": "2021-01-01T11:46:18.082Z"
  }
}
```

With your comments above and I also checked at @anthonykhoa's development server and it looks good. 

I didn't actually have any concerns with the PR itself, but I just wanted to make sure we're doing due diligence on managing the db side between prod and app. I also don't regard this as an improvised fix. We are simply being more explicit in what we're calling our relations in SQL.

hwong0305 (approved):

PR #497, opened 2021-02-11
pin npm packages version to minor patch

### LINKED ISSUE #482: change package versions to include minor and patch versions inside package.json
it is safer than using whole number caret versions

ggwadera:
Closes #482

Just reverted the previous choice of pinning only the major version of our dependencies.

Now it's locked up to the minor patch version.


anthonykhoa (approved):

anthonykhoa (approved):

__tests__/pages/__snapshots__/curriculum.test.js.snap
```diff
@@ -776,7 +776,7 @@ exports[`Curriculum Page Should render with basic dummy data 1`] = `
           />
           <a
             class="mx-3"
-            href="#"
+            href="/#"
```
anthonykhoa, line 779: All the versions are still the same(`the latest version`) right? So why did this snapshot change?

ggwadera: Who knows? hahaha

I think that today it felt that having the slash there was better.

ggwadera: Maybe now that we have the versions pinned to a single one this will go away.

songz (approved):

PR #498, opened 2021-02-13
add admin/users test

Ulisseus:
Adds tests for `admin/users page` and its components. Part of #409 and #433.

### DIFF
--- a/components/admin/AdminLayout.tsx
+++ b/components/admin/AdminLayout.tsx
@@ -3,8 +3,10 @@ import LoadingSpinner from '../../components/LoadingSpinner'
 import Layout from '../../components/Layout'
 import _ from 'lodash'
 import { GetAppProps } from '../../graphql'
-
+import Error, { StatusCode } from '../../components/Error'
+import { useRouter } from 'next/router'
 export const AdminLayout: React.FC<GetAppProps> = ({ data, children }) => {
+  const router = useRouter()
   const { loading, error, session } = data
 
   if (loading) {
@@ -12,14 +14,10 @@ export const AdminLayout: React.FC<GetAppProps> = ({ data, children }) => {
   }
 
   if (error) {
-    return (
-      <Layout>
-        <h1>Error</h1>
-      </Layout>
-    )
+    return <Error code={StatusCode.INTERNAL_SERVER_ERROR} />
   }
 
-  if (!session) window.location.href = '/login'
+  if (!session) router.push('/login')
 
   const isAdmin = _.get(session, 'user.isAdmin', false)
 
--- a/components/admin/users/AdminUsersTable.tsx
+++ b/components/admin/users/AdminUsersTable.tsx
@@ -5,6 +5,7 @@ import changeAdminRights from '../../../graphql/queries/changeAdminRights'
 import { User } from '../../../graphql'
 import { AdminUsersSplitSearch } from './AdminUsersSplitSearch'
 import { filter } from '../../../pages/admin/users'
+import _ from 'lodash'
 
 type UsersListProps = {
   users: User[]
@@ -56,9 +57,9 @@ const AdminOption: React.FC<AdminOptionProps> = ({
     }
   }
 
-  const changeButton = async () => {
-    await changeRights(mutationVariable)
-    const newUsers = [...users]
+  const changeButton = () => {
+    changeRights(mutationVariable)
+    const newUsers = _.cloneDeep(users)
     newUsers[usersIndex].isAdmin = newAdminRights
     setUsers(newUsers)
   }
--- a/pages/admin/users.tsx
+++ b/pages/admin/users.tsx
@@ -70,10 +70,15 @@ const AdminUsers: React.FC<QueryDataProps<AllUsersData>> = ({ queryData }) => {
           onClick={(value: string) => changeFilter(value, 'option')}
           currentOption={searchOption.option}
         >
-          Search By:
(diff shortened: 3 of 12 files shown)


components/admin/AdminLayout.tsx
```diff
@@ -3,23 +3,21 @@ import LoadingSpinner from '../../components/LoadingSpinner'
...
   if (error) {
-    return (
-      <Layout>
-        <h1>Error</h1>
-      </Layout>
-    )
+    return <Error code={StatusCode.INTERNAL_SERVER_ERROR} />
   }
 
-  if (!session) window.location.href = '/login'
```
Ulisseus, line 22: Changed to next/router push

components/admin/AdminLayout.tsx
```diff
@@ -3,23 +3,21 @@ import LoadingSpinner from '../../components/LoadingSpinner'
...
   const { loading, error, session } = data
 
   if (loading) {
     return <LoadingSpinner />
   }
 
   if (error) {
-    return (
-      <Layout>
-        <h1>Error</h1>
```
Ulisseus, line 17: Changed to new Error component.

components/admin/users/AdminUsersTable.tsx
```diff
@@ -56,9 +57,9 @@ const AdminOption: React.FC<AdminOptionProps> = ({
     }
   }
 
-  const changeButton = async () => {
-    await changeRights(mutationVariable)
-    const newUsers = [...users]
+  const changeButton = () => {
+    changeRights(mutationVariable)
+    const newUsers = _.cloneDeep(users)
```
Ulisseus, line 62: Fixed a bug, right now if you open this page and try to add admin rights it will throw an error, something about `newUsers[usersIndex].isAdmin` is readonly. Deepcloning fixed it.

components/admin/users/AdminUsersTable.tsx
```diff
@@ -56,9 +57,9 @@ const AdminOption: React.FC<AdminOptionProps> = ({
     }
   }
 
-  const changeButton = async () => {
-    await changeRights(mutationVariable)
```
Ulisseus, line 60: There is no reason to wait for this mutation since its return value isn't used anywhere.

pages/admin/users.tsx
```diff
@@ -70,10 +70,15 @@ const AdminUsers: React.FC<QueryDataProps<AllUsersData>> = ({ queryData }) => {
           onClick={(value: string) => changeFilter(value, 'option')}
           currentOption={searchOption.option}
         >
-          Search By:
+          <label htmlFor="search">Search By:</label>
```
anthonykhoa, line 73: What does this do, does it make the UI look better?

pages/admin/users.tsx
```diff
@@ -70,10 +70,15 @@ const AdminUsers: React.FC<QueryDataProps<AllUsersData>> = ({ queryData }) => {
...
         </FilterButtons>
       </div>
-      <input type="text" className="form-control" onChange={handleChange} />
+      <input
+        type="text"
+        id="search"
```
anthonykhoa, line 78: Why is the `id` added? Does it have something to do with the label being added?

components/admin/users/AdminUsersTable.tsx
anthonykhoa, line 62: Was there a problem that was occurring? Why is the spread operator replaced with `._cloneDeep` ?

pages/admin/users.tsx
Ulisseus: UI looks the same but accessibility is better, you can select input by label and use better testing library query (and it's [recommended](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names) label inputs anyway).

components/admin/users/AdminUsersTable.tsx
anthonykhoa: I tried to add and remove admin rights on the website and its working perfectly. Can you take a screenshot of the error?

pages/admin/users.tsx
Ulisseus: There are two ways to add label to html tag, wrap it around and use `for` tag (`htmlFor` in case of React). I can't simply wrap input with new label because it will change UI so I used the latter method. `Search by` is now label for `input` despite being separate non nested tags.

anthonykhoa: Ah, thanks for the detailed explanation!

components/admin/users/AdminUsersTable.tsx
Ulisseus: adminError
I haven't thought about why it happens, I guess spread didn't go deep enough and we're trying to reassign original users value which is used in a hook?

anthonykhoa: This is really strange, because I just tested on my local, and it doesn't work -- I get the same error you do. But in production it works fine. And before, when I created this page, it worked fine on local. Maybe an update to our dependencies caused this error? Weird... thanks for noticing the error and fixing it :)

anthonykhoa: More Info:
I tried it with our production build (`yarn start -p 3000`), and it worked fine.

anthonykhoa (approved):

ggwadera (approved):

hwong0305 (approved):

PR #501, opened 2021-02-20
fix admin/lessons page

Ulisseus:
This is part of #409. As it is production works fine but development build fails because of runtime typescript type errors (trying to change useState value directly instead of creating a copy). I guess typescript or hooks were added later and since there were no tests nobody noticed. Production also won't be able to render empty lessons. 
Without fixing this first I won't be able to submit tests (they will fail, first from of type errors, second from incorrect index errors).

### DIFF
--- a/components/admin/lessons/AdminLessonInfo.tsx
+++ b/components/admin/lessons/AdminLessonInfo.tsx
@@ -152,7 +152,7 @@ export const AdminLessonInfo: React.FC<LessonInfoProps> = ({
   selectedLesson
 }) => {
   // true when user clicks on `create new lesson` button
-  if (lessons && selectedLesson === lessons.length - 1) {
+  if (lessons && selectedLesson === lessons.length) {
     return <NewLesson setLessons={setLessons} />
   }
 
--- a/components/admin/lessons/AdminLessonsSideBar.tsx
+++ b/components/admin/lessons/AdminLessonsSideBar.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import _ from 'lodash'
 import { Lesson } from '../../../graphql/index'
 
 type SideBarLessonProps = {
@@ -13,14 +14,17 @@ export const AdminLessonsSideBar: React.FC<SideBarLessonProps> = ({
   setSelectedLesson,
   selectedLesson
 }) => {
-  const lessonListData = lessons || []
-
+  const lessonListData = _.cloneDeep(lessons) || []
+  if (lessonListData.length === 0) {
+    lessonListData.push({ title: 'Create New Lesson' })
+  }
   //check if create new lesson has already been pushed to the lessons array
   const lastIndex = lessonListData.length - 1
   const { title } = lessonListData[lastIndex]
   if (title !== 'Create New Lesson') {
     lessonListData.push({ title: 'Create New Lesson' })
   }
+
   const lessonList = lessonListData.map((obj: any, arrIndex: number) => (
     <a
       key={arrIndex}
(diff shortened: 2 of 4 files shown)


components/admin/lessons/AdminLessonInfo.tsx
```diff
@@ -152,7 +152,7 @@ export const AdminLessonInfo: React.FC<LessonInfoProps> = ({
   selectedLesson
 }) => {
   // true when user clicks on `create new lesson` button
-  if (lessons && selectedLesson === lessons.length - 1) {
+  if (lessons && selectedLesson === lessons.length) {
```
Ulisseus, line 155: There is lessons useState value at lessons page. It's passed to AdminLessonsSideBar where it is cloned and `Create new lesson` title is appended. Lessons themselves are not modified. So if you render 3 lessons, AdminLessonsSideBar will have 4 buttons and etc.

components/admin/lessons/AdminLessonsSideBar.tsx
```diff
@@ -13,14 +14,17 @@ export const AdminLessonsSideBar: React.FC<SideBarLessonProps> = ({
   setSelectedLesson,
   selectedLesson
 }) => {
-  const lessonListData = lessons || []
-
+  const lessonListData = _.cloneDeep(lessons) || []
```
Ulisseus, line 17: Without cloning it this function tries to edit useState value which will cause runtime typescript errors.

components/admin/lessons/AdminLessonsSideBar.tsx
```diff
@@ -13,14 +14,17 @@ export const AdminLessonsSideBar: React.FC<SideBarLessonProps> = ({
...
+  if (lessonListData.length === 0) {
```
Ulisseus, line 18: Without this check AdminLessonSideBar component won't be able to render empty array of lessons, `lastIndex` will be `-1` and you can get `title` value of undefined resulting in error. This probably won't happen in production (we are not going to delete all lessons), but since tests should cover 100% I needed to add it.

helpers/admin/adminHelpers.ts
```diff
@@ -7,8 +8,9 @@ export const getPropertyArr = (
   deleteProps?: string[],
   dropdownChange?: Function
 ) => {
-  ;(deleteProps || []).forEach(prop => delete options[prop])
-  const keys = Object.keys(options)
+  const cloneOptions = _.cloneDeep(options)
```
Ulisseus, line 11: Trying to edit useState value directly again hence cloning to avoid type errors.

pages/admin/lessons.tsx
```diff
@@ -4,11 +4,18 @@ import { AdminLessonsSideBar } from '../../components/admin/lessons/AdminLessons
 import { withGetApp, GetAppProps } from '../../graphql'
 import { Lesson } from '../../graphql/index'
 import { AdminLayout } from '../../components/admin/AdminLayout'
-
+import Error, { StatusCode } from '../../components/Error'
+import LoadingSpinner from '../../components/LoadingSpinner'
 const Lessons: React.FC<GetAppProps> = ({ data }) => {
+  const { loading, error, lessons } = data
+  if (loading) return <LoadingSpinner />
```
anthonykhoa, line 11: This is unnecessary, it is handled in AdminLayout
https://github.com/garageScript/c0d3-app/blob/master/components/admin/AdminLayout.tsx#L16

There error is handled in AdminLayout as well, except one difference is that there is no `message` prop value passed into the Error component that is returned

Ulisseus: Fixed.

ggwadera (approved):

anthonykhoa (approved):

PR #504, opened 2021-02-23
add AdminLessonChallenges test

Ulisseus:
Adds test for `admin/AdminLessonChallenges` component. Part of #409 and #433.

`try{someMutation}catch(e){throw new Error(e)}` pattern was very hard to test, most of the time it crashed the whole test suit, so I logged error to Sentry straight away. This should not change anything as far as logging goes, just some duplicate code. Right now every uncaught error is handled by `_error` page which uses the same `Sentry.catchException` method. Also, with manual logging we can show notifications for user if something went wrong.

### DIFF
--- a/components/admin/lessons/AdminLessonChallenges.tsx
+++ b/components/admin/lessons/AdminLessonChallenges.tsx
@@ -2,6 +2,7 @@ import { useMutation } from '@apollo/client'
 import React, { useEffect, useState } from 'react'
 import updateChallenge from '../../../graphql/queries/updateChallenge'
 import { FormCard } from '../../FormCard'
+import * as Sentry from '@sentry/browser'
 import createNewChallenge from '../../../graphql/queries/createChallenge'
 import { Lesson, Challenge, Maybe } from '../../../graphql/index'
 import {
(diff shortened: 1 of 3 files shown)


components/admin/lessons/AdminLessonChallenges.tsx
```diff
@@ -156,15 +157,21 @@ export const AdminLessonChallenges: React.FC<LessonChallengesProps> = ({
...
   }
-
   const allChallenges = !challenges
     ? []
-    : challenges.map((challenge: Maybe<Challenge>, key: number) => (
-        <LessonChallenge challenge={challenge} alter={alter} key={key} />
-      ))
-
+    : challenges.map((challenge: Maybe<Challenge>, key: number) => {
+        if (challenge !== null) {
```
Ulisseus, line 166: Lesson type is rather loose so challenges could be null or array of nulls. First ternary operator checks for null, if condition inside map checks every challenge. Since we're sure that every single challenge is not null we can pass `challenge as Challenge` to `LessonChallenge` component instead of `Maybe<Challenge>`. 
This won't change anything as far as UX goes.

ggwadera (approved):

songz (approved):

PR #509, opened 2021-03-01
Closes #547  - fix navBar highlighting 

### LINKED ISSUE #547: Navigation bar highlighting doesn't work
Curriculum button is always highlighted no matter what. 
Contributors button doesn't get highlighted when you go to `https://www.c0d3.com/contributors`.

Ulisseus:
There are many snapshots updates but they all are exactly the same (change curriculum classname from active to inactive). I added named useRouter mock because you need to mock it in most tests. It also made `withTestRouter` wrapper unnecessary because you can specify queries using new router mock. The same goes for testing routes, no need to define `window.location.pathname` every time you need to test redirection. Actual highligting fix is only two files - `AppNav` and `NavLink`. 

Big yarn diff is minor dependencies updates. I guess nobody run `yarn install` for a while. You can pull fresh master, delete `yarn.lock` and you will get exactly the same difference.


components/AppNav.tsx
```diff
@@ -25,39 +27,38 @@ const dropdownMenuItems = [
 
 const NavBar: React.FC<AuthLinkProps> = ({ session }) => {
   const isAdmin = _.get(session, 'user.isAdmin', '')
+  //window object is not avaible in server side rendering phase
+  const location: string =
+    window && '/' + window.location.pathname.split('/')[1]
```
Ulisseus, line 32: Without first check next will print errors (since static bundling is done server side and there is no window object). Splitting path by `/` will always produce array with more than one item.

hwong0305: Should get the location from `router` instead

helpers/dbload.ts
```diff
@@ -19,7 +19,7 @@ const options: Options = {
   }
 }
 
-if (process.env.NODE_ENV != 'production') {
+if (process.env.NODE_ENV === 'development') {
```
hwong0305, line 22: We need more clarify on why this check is added. It should be always used.

components/AppNav.tsx
Ulisseus: What are advantages of getting pathname from router? Because if I use router I need to change every test which uses this component otherwise they will fail because router is null.

helpers/dbload.ts
Ulisseus: I removed this check altogether. Not dbload uses `DB_PORT` env variable and next config has fallback in case if this variable was not configured. It looks much cleaner IMO and it should work in production (assuming postgres is running on default 5432 port).

ggwadera (commented):
Additionally, since a lot of tests will require a `useRouter` mock now, and there's a lot of repeated boilerplate code (like mocking the same return values), I think it would be better to have a [manual mock](https://jestjs.io/docs/en/manual-mocks) with these things already done.

For example, we can have a file `__mocks__/next/router.js` like the one below, and then this file will be the one imported whenever you write `import { useRouter } from 'next/router'`. Also, this way we don't need to use `jest.mock('next/router')` since that will be done automatically by Jest with the manual mock.

```js
const push = jest.fn()

export const useRouter = jest.fn().mockImplementation(() => {
  return {
    route: '/',
    pathname: '/',
    query: {},
    asPath: '/',
    push,
    replace: async () => true,
    reload: () => null,
    back: () => null
  }
})
```

Then on the tests that you need to mock or assert something about the mocks, you can do this for example:

```js

describe('router mock tests', () => {
  beforeEach(() => {
    jest.clearAllMocks()
  })

  const { push } = useRouter()

  it('should be able to mock implementation', () => {
    let called = false
    push.mockImplementationOnce(() => {
      called = true
    })
    useRouter().push()
    expect(called).toBe(true)
  })

  it('should be able to assert mock call', () => {
    const path = '/login'
    useRouter().push(path)
    expect(push).toBeCalledWith(path)
  })
})
```

components/AppNav.tsx
```diff
@@ -27,9 +28,8 @@ const dropdownMenuItems = [
 
 const NavBar: React.FC<AuthLinkProps> = ({ session }) => {
   const isAdmin = _.get(session, 'user.isAdmin', '')
-  //window object is not avaible in server side rendering phase
-  const location: string =
-    window && '/' + window.location.pathname.split('/')[1]
+  const router = useRouter()
+  const location = '/' + router.asPath.split('/')[1]
```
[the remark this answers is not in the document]
Ulisseus: They are interchangeable in this case.
https://www.c0d3.com/curriculum/5:
 - `pathname` returns `/curriculum/[lesson]`
 - `asPath` returns `/curriculum/5`
 Either way you need to split it by `/` and prepend `/` to check against active path.

Ulisseus:
This PR got really big but it hard to split it up, so please review it as it is. Small changes in `AppNav` and `NavLink` mean every single snapshot needs to be updated. Since `AppNav` uses `router`, every test and story need to mock `router` somehow. As **ggwadera** pointed out there is a lot of boilerplate code in test if you mock it explicitly, so I added named `router` mock which simplifies things a lot. 

`dbload` file was changed because of failure of vercel deployment. I added fallback in next.config for `DB_PORT` variable, so everything should be the same.

components/AppNav.tsx
```diff
@@ -25,39 +28,37 @@ const dropdownMenuItems = [
...
+    },
+    {
+      path:
+        'https://www.notion.so/Table-of-Contents-a83980f81560429faca3821a9af8a5e2',
+      name: 'Journey',
+      external: true
+    },
+    { path: 'https://chat.c0d3.com', name: 'Help', external: true },
+    { path: '/contributors', name: 'Contributors' }
+  ]
```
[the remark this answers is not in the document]
Ulisseus: Done.

__tests__/pages/curriculum/[lesson].test.js
```diff
@@ -81,7 +84,8 @@ describe('Lesson Page', () => {
     ]
     useRouter.mockReturnValueOnce({
       query: {
-        lesson: '2'
+        lesson: '2',
+        asPath: '/'
```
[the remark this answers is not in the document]
Comment (Ulisseus) [reply __tests__/pages/curriculum/[lesson].test.js]: I removed `withTestRouter` from tests altogether, named router mock can handle queries too and it's much more succinct. The former is used in stories only now.

testUtil/withNextRouter.js
```diff
@@ -6,9 +6,9 @@ import { RouterContext } from 'next/dist/next-server/lib/router-context'
 export function withTestRouter(tree, router = {}) {
   const {
     route = '',
-    pathname = '',
+    pathname = '/not-root',
```
ggwadera, line 9: Why is this needed?

Ulisseus: It's not needed, and I should probably revert it back. Initially I intended to test redirections with `withTestRouter` and the base route should be different from `/` in case if you need check redirect to/from index file.

hwong0305:
> t

Could you please separate the changes out to `dbload` and the `withRouter` mock in a different PR? That way it's clearer on what needs to happen with the navbar changes.

Ulisseus:
> Could you please separate the changes out to `dbload` and the `withRouter` mock in a different PR? That way it's clearer on what needs to happen with the navbar changes.

I can remove dbload changes from this PR but it break vercel deployment (that's why I added them initially). I have no idea why it behaves that way now. 

Removing manual `router` mock will in fact make things much messier. A lot of components use `AppNav` and it means that every single one of them needs to have `router` mock. I tried to it that way in earlier [commit ](https://github.com/garageScript/c0d3-app/pull/509/commits/003ef14286c435e3bf057a05bb239bb0c9ed468d), it's still one gigantic pr.

components/AppNav.tsx
```diff
@@ -69,9 +71,7 @@ const AuthButton: React.FC<AuthButtonProps> = ({ initial, username }) => {
   const [logoutUser, { data }] = useLogoutMutation()
   useEffect(() => {
     const success = _.get(data, 'logout.success', false)
-    if (success) {
-      window.location.pathname = '/'
-    }
+    if (success) window.location.pathname = '/'
```
anthonykhoa, line 74: This originally used window.location, but can you change it to use `router.push` instead? The consistency in using the router object will make the codebase more clear

Ulisseus: It won't work. Without full refresh the page stays the same after logout. At first I thought it was because next uses fast refresh in development but production build is the same. 
I am not sure what options I need to pass to `router` to force full refresh.
The same thing with `login` page redirection.

anthonykhoa: Oo thanks for detailed explanation! Things seem a lot more complicated than it appears to be on the surface. I'll try having a go at this and see if I can provide a solution

anthonykhoa: I couldn't come up with a solution -- I guess nextjs `router.push` function doesn't like it when we pass in `/` as a value :(
No other simple solution came to mind as well, so it seems we are forced to use `window.location here`

hwong0305 (approved):

hwong0305:
Thanks. It's easier to look at if you only have a single change through multiple files.

ggwadera (approved):

components/AppNav.tsx
```diff
@@ -17,47 +18,48 @@ type AuthLinkProps = {
...
-        external
-      >
-        Help
-      </NavLink>
-      <NavLink path="/contributors" className="nav-item nav-link">
-        Contributors
-      </NavLink>
+      {navItems.map(button => (
+        <NavLink
+          path={button.path}
```
anthonykhoa, line 54: The button object should be deconstructed

PR #516, opened 2021-03-02
Add adminLessonInfo test

Ulisseus:
Just another test for admin panel.


components/admin/lessons/AdminLessonInfo.tsx
```diff
@@ -152,14 +153,15 @@ export const AdminLessonInfo: React.FC<LessonInfoProps> = ({
   selectedLesson
 }) => {
   // true when user clicks on `create new lesson` button
-  if (lessons && selectedLesson === lessons.length) {
+  if (!lessons || selectedLesson === lessons.length) {
```
Ulisseus, line 156: Without check for lessons this component won't handle empty Lesson data. This is unlikely to happen in production but since lesson type is `Lesson[]|undefined` I need to cover for it.

components/admin/lessons/AdminLessonInfo.tsx
```diff
@@ -152,14 +153,15 @@ export const AdminLessonInfo: React.FC<LessonInfoProps> = ({
...
     return <NewLesson setLessons={setLessons} />
   }
 
   // set currently selected lesson
-  const lesson = lessons && lessons[selectedLesson]
+  const lesson = lessons[selectedLesson] ? lessons[selectedLesson] : lessons[0]
```
Ulisseus, line 161: Incorrect lesson id case.

components/admin/lessons/AdminLessonInfo.tsx
```diff
@@ -152,14 +153,15 @@ export const AdminLessonInfo: React.FC<LessonInfoProps> = ({
...
 
-  const lessonId = parseInt(lesson ? lesson.id + '' : '')
+  //TO-DO change Lesson type to include non-nullable Int id field
+  const lessonId = parseInt(lesson.id + '')
```
Ulisseus, line 164: Our grapql types strike again. According to our type system `lesson.id` can be null but every lesson mutation requires id. So this component will render everything correctly but nothing will work. 

Ideally lesson.id should be non-nullable int.

ggwadera (approved):

components/admin/lessons/AdminLessonInfo.tsx
anthonykhoa: This can be shortened to 
```
const lesson = lessons[selectedLesson] || lessons[0]
```
This will make the code shorter and easier to understand. In this scenario, it makes more contextual sense to use the logical OR operator rather than ternary. Your use of ternary operator is just doing what the logical OR operator already does.

anthonykhoa (approved):

components/admin/lessons/AdminLessonInfo.tsx
hwong0305: Put this in the comment for the code and not in the PR.

PR #525, opened 2021-03-05
Added test for pages/_app

kondanna:


Ulisseus:
You have two tests in this PR, one for _app, another for lessons. You should remove the latter. 
And one snapshot for _app [failed](https://github.com/garageScript/c0d3-app/pull/525/checks#step:4:353).

kondanna:
submitting new request with updated snapshot

PR #540, opened 2021-03-09
Adds test for pages/_app

kondanna:
updates #409

This PR adds tests for the `pages/_app` file, bringing it to 100% line coverage!


__tests__/pages/_app.test.js
```diff
@@ -0,0 +1,82 @@
...
+import { MockedProvider } from '@apollo/client/testing'
+import MyApp from '../../pages/_app'
+import Login from '../../pages/login'
+import * as Sentry from '@sentry/browser'
+import GET_APP from '../../graphql/queries/getApp'
+import dummyLessonData from '../../__dummy__/lessonData'
+import dummySessionData from '../../__dummy__/sessionData'
+import dummyAlertData from '../../__dummy__/alertData'
+import posthog from 'posthog-js'
+jest.spyOn(Sentry, 'captureException')
```
ggwadera, line 16: It's better to mock the whole posthog module here instead of that single function.

```js
jest.mock('posthog-js')
import posthog from 'posthog-js'
```

Then you can check with the real function name.

```js
expect(posthog.init).not.toHaveBeenCalled()
```

__tests__/pages/_app.test.js
```diff
@@ -0,0 +1,82 @@
...
+      <MockedProvider mocks={mocks} addTypename={false}>
+        <MyApp Component={Login} />
+      </MockedProvider>
+    )
+    await waitForElementToBeRemoved(() => queryByText('Loading...'))
+    await (() =>
+      expect(posthog.init).toHaveBeenCalledWith(process.env.POSTHOG_API_KEY, {
+        api_host: 'https://app.posthog.com'
+      }))
+    await (() => console.log(process.env))
```
anthonykhoa, line 69: Why are we logging this value?

ggwadera: Also you could remove the `await` on the `expect` call on both tests, it's not doing anything there.

If the tests fails then do `await waitFor(() => expect(...))`

anthonykhoa: Here's an example of when to use `waitFor`:
Let's say we are making a test for the login page such that when someone presses the submit button with incorrect input fields, a popup appears saying `Lol u needa put correct values in the fields bro`. In order to test if the popup appeared with that test, you might want to do something like 
```
await waitFor(() => getByText('Lol u needa put correct values in the fields bro'))
```
This is because that popup might take some time to render, so you might need to wait with `waitFor` until it finishes rendering

Ulisseus (approved):

ggwadera (approved):

PR #544, opened 2021-03-12
Closes #551, #550 - add DB_PORT variable

### LINKED ISSUE #551: No option for postgres port in production
Right now production supposedly runs on default 5432 port (If you don't specify port explicitly sequelize uses default port for the given dialect). Meanwhile some users need to run db on other ports locally. There is currently `process.env.NODE_ENV` check as workaround for `DB_PORT` variable but it could possible cause problems in some non-standard environments (like `preview` on vercel).

We need to add `DB_PORT` globally and remove unnecessary env checks.

Ulisseus:
Initially it wasn't possible to use anything except for default postres port (5432). A few weeks ago I created new `DB_PORT` environment variable and added `process.env.NODE_ENV` check to keep production untouched. But turns out it can cause problems with vercel preview deployment. 

Honestly, I am not sure if it was `DB_PORT` at fault, deployment fail happened only once and I can't get any feedback/logs from vercel. But this PR will make `dbload` more readable anyways. It removes env check and adds fallback for `DB_PORT` in `next.config`.  Other sequelize options have fallbacks in config too, so unreachable `OR` checks were removed as well.


ggwadera (approved):

helpers/dbload.ts
```diff
@@ -7,26 +7,22 @@ import { ChallengeTypes, Challenge } from './models/Challenge'
...
-
-if (process.env.NODE_ENV != 'production') {
-  options.port = ((process.env.DB_PORT as unknown) as number) || 5432
-}
 const sequelize = new Sequelize(
-  process.env.DB_NAME || 'you',
-  process.env.DB_USER || 'failed',
-  process.env.DB_PW || 'this',
-  options
+  process.env.DB_NAME as string,
```
anthonykhoa, line 11: fields of the process.env object are already strings, so the typecasting shouldnt be needed

helpers/dbload.ts
```diff
@@ -7,26 +7,22 @@ import { ChallengeTypes, Challenge } from './models/Challenge'
...
+  process.env.DB_USER as string,
+  process.env.DB_PW,
+  {
+    host: process.env.DB_HOST,
+    logging: false,
+    port: (process.env.DB_PORT as unknown) as number,
```
anthonykhoa, line 17: since  fields in the process.env object are strings, it will be clear if you explicitly convert the string to a number with parseInt, instead of implicitly doing it with typescript

helpers/dbload.ts
Ulisseus: It's `string|undefined` (in case if env variable wasn't configured) so this type assertion is needed.

helpers/dbload.ts
Ulisseus: I still need type assertion for `parseInt`, because it requires string as argument, so it's going to be `parseInt(process.env.DB_PORT as string)`. Should I change it? I don't think there is a clear solution to this.

helpers/dbload.ts
anthonykhoa: In that case it'd be better to be explicitly clear and give it a default string value in case it is undefined. 
Like 
```
process.env.DB_NAME || 'you'
```
This change will also allow us to use only javascript, which simplifies things.

anthonykhoa: Oh hmm I see that in our config file, we already set default values for all the env variables, so it would actually make contextual sense to use `as string` in these instances

anthonykhoa: Is it possible for us to create typescript definitions for our env variables such that they are `string` instead of `string | undefined` ? I think that would increase the clarity of our codebase

helpers/dbload.ts
anthonykhoa: Nice yea, `parseInt(process.env.DB_PORT as string)` looks about as clear as it can get

Ulisseus: Changed.

anthonykhoa (approved):

anthonykhoa:
Can you create an issue for this and tag it in your PR description? This will allow us to have a more organized issue/PR history.

Ulisseus:
> Can you create an issue for this and tag it in your PR description? This will allow us to have a more organized issue/PR history. Each PR should have an issue created for it anyway, since tagging issues in our commit messages is in our engineering practices

This PR was the part of #509 and it was removed because it got too big to review. I am not sure what was causing vercel problems because I can't get any feedback or logs. Creating issue like `dbload.tsx looks ugly, refactor` is probably not a good idea.

anthonykhoa:
> This PR was the part of #509 and it was removed because it got too big to review. I am not sure what was causing vercel problems because I can't get any feedback or logs. Creating issues like `dbload.tsx looks ugly, refactor` is probably not a good idea.

Oh, I thought you made this PR to solve the issue of vercel preview deployment not working. Even if you intended to make this PR to only refactor code to look cleaner, you ended up fixing the issue of `vercel preview deployment` not working too right? We should document that the `vercel preview` issue has occurred by making an issue for it, and additionally tag that issue to this PR. That will create a more organized history documentation for us. 

Documentation is very important and can prove to be useful later on. For example, if anyone else in the future comes across the problem of `vercel deployment` not working, they can search for the issue and its corresponding PR to find meaningful clues.

PR #546, opened 2021-03-13
fix curriculum highlighting

Ulisseus:
This is the lighter version of #509. There are a lot of snapshot changes but they are exactly the same (change styling of `curriculum` button). So there are only small 2 files which need reviewing - `AppNav` and `NavLink`.
I used `window.location.pathname` instead of `router` because latter requires a lot of mocking changes for every test and turns this PR into #509 (and there shouldn't be any difference between `window` and `router` anyways in this case). I won't close #509 so you can review whatever is easier for you.

### DIFF
--- a/components/AppNav.tsx
+++ b/components/AppNav.tsx
@@ -49,10 +49,9 @@ const NavBar: React.FC<AuthLinkProps> = ({ session }) => {
     <div className="navbar-nav collapse navbar-collapse">
       {navItems.map(button => (
         <NavLink
-          path={button.path}
+          {...button}
           className="nav-item nav-link"
           key={button.name}
-          external={button.external}
           activePath={location === button.path}
         >
           {button.name}
--- a/components/AppNav.tsx
+++ b/components/AppNav.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from 'react'
-import NavLink from './NavLink'
+import NavLink, { NavLinkProps } from './NavLink'
 import LoadingSpinner from './LoadingSpinner'
 import { Button } from './theme/Button'
 import { DropdownMenu } from './DropdownMenu'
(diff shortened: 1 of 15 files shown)


components/AppNav.tsx
```diff
@@ -17,47 +17,47 @@ type AuthLinkProps = {
...
-        path="https://chat.c0d3.com"
-        className="nav-item nav-link"
-        external
-      >
-        Help
-      </NavLink>
-      <NavLink path="/contributors" className="nav-item nav-link">
-        Contributors
-      </NavLink>
+      {navItems.map(button => (
```
anthonykhoa, line 50: You should destructure the button object here, that will make the code more concise

components/AppNav.tsx
```diff
@@ -17,47 +17,47 @@ type AuthLinkProps = {
...
+    name: 'Journey',
+    external: true
+  },
+  { path: 'https://chat.c0d3.com', name: 'Help', external: true },
+  { path: '/contributors', name: 'Contributors' }
+]
 
 const NavBar: React.FC<AuthLinkProps> = ({ session }) => {
   const isAdmin = _.get(session, 'user.isAdmin', '')
+  const location = window && '/' + window.location.pathname.split('/')[1]
```
anthonykhoa, line 46: We should stick to just using `useRouter` path related things in order to maintain codebase consistency.

components/AppNav.tsx
Ulisseus: Changed.

components/AppNav.tsx
Ulisseus: Using `router` turns this pr into #509. Considering that it took more than 12 days to review it I think people don't really like to review 40 files at once.

anthonykhoa: It's unfortunate that it may take longer to get reviewed, but that factor is not a good enough reason to decrease the quality of the codebase

Ulisseus: Well, my intend was to fix highlighting with this pr and add router, router mocks and etc with the next pr.
Also, we are already using `window.location` for changing routes in some cases.

ggwadera: > Also, we are already using `window.location` for changing routes in some cases.

That needs to be fixed as well, changing routes without the router forces the browser to refresh instead of just modifying the DOM.

anthonykhoa: > Well, my intend was to fix highlighting with this pr and add router, router mocks and etc with the next pr.

Ah, I see! It makes sense that this is its own separate PR then. Can you make separate issues for each of the problems you are fixing, and tag the issues you are tackling in each PR description? Like for example putting a `closes #33` in the PR description. That will provide us a more organized way of letting us know which specific issue each PR is handling. Tagging issues in your PR description will also automagically close those tagged issues as well, which is really convenient:
https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue

Furthermore, each commit message should have an issue number(s) tagged too -- it's in our engineering practices:
https://github.com/garageScript/c0d3-app/wiki/Engineering-Practices

If I don't see the issue number of `changing window.location to router` being in the PR description, then I'll know the PR doesn't have the intention of doing that. And then maybe to add additional clarity for the reviewer, you can note in your PR description that are not going to change `window.location` to `router` in this PR, but that you've already made an issue for it so it will be addressed in a later PR.

Issue #555, opened 2021-03-17
Self-host lessons and other documents [feature]


Ulisseus:
Currently all c0d3 documents are on notion which works fine but it's strange to use 3rd party site to host relatively simple text documents. 

* There is a difference in styles and design between c0d3 and notion
* No version control
* Limited functionality (with self-hosted documents you can add all sorts of interactive elements, animations, react components and etc)
* Everyone can fix a typo/update something by submitting a PR which could be great introduction to github for new students

Moving away from notion is a big change for students and mentors both, so I propose starting with simpler pages like development setup and journey. 

Current plan is:
* change withSass to CssModules
* add new document components like spoiler, modal image and etc
* add [mdx](https://mdxjs.com/) support

anthonykhoa:
Testing to see if github bot works

Ulisseus:
All lessons are complete. 

- [JS0 ](https://www.c0d3.com/curriculum/lessons/js0/primitive_data_and_operators)
- [JS1](https://www.c0d3.com/curriculum/lessons/js1/hypertext_markup_language)
- [JS2](https://www.c0d3.com/curriculum/lessons/js2/introduction_to_testing_inner_properties)
- [JS3](https://www.c0d3.com/curriculum/lessons/js3/preflight)
- [JS4](https://www.c0d3.com/curriculum/lessons/js4/interactive_elements)
- JS5 (pending to be merged)

What's next? A few rounds of proofreading of course. Also, we need to update challenges in some lessons. Only then we could start changing notion links.

kondanna:
For JS0 Part 1:
![JS0 - Objects](https://user-images.githubusercontent.com/16906103/121796608-0ed9ab80-cc23-11eb-815f-084aaa96639e.PNG)
* 'The last data type you will need to know is Objects' feels strangely placed in that part of the section
* it says 'and in this section' but the functions are actually introduced in Part 2 (next page)
![JS0 - Strings - answer](https://user-images.githubusercontent.com/16906103/121796610-0f724200-cc23-11eb-8065-4970bf2ccf4b.PNG)
either 'd takes the value of c _joined with_ the value of message' or 'd takes the value of c _and joins it with_ the value of message'
![JS0 - Primitives - spacing](https://user-images.githubusercontent.com/16906103/121796611-0f724200-cc23-11eb-90b8-ac406898756f.PNG)
the spacing should probably be adjusted

kondanna:
For JS0 Part 2:

[...]

PR #556, opened 2021-03-17
Update #555 - add css modules, remove next-sass

Ulisseus:
`next-sass` is incompatible with `mdx`. Also, modern Next [supports](https://nextjs.org/blog/next-9-3#built-in-sass-css-module-support-for-component-level-styles) global sass styles anyways. 
>If you are currently using @zeit/next-sass we recommend removing the plugin from your next.config.js and package.json, thereby moving to the built-in Sass support.

Adding local css modules requires a bit more work. Previously you could `import '../scss/foobar.scss'`  and all `foobar` styles would be apllied automatically to this component. Unfortunately, this works only for global styles, for local ones you need to import style object `import style from '../scss/foobar.scss'` and apply it to every className. See [docs](https://nextjs.org/docs/basic-features/built-in-css-support#adding-component-level-css).


components/AnnouncementCard.tsx
```diff
@@ -1,16 +1,19 @@
...
 
 type Props = {
   announcements: string[]
 }
 
 const AnnouncementCard: React.FC<Props> = ({ announcements }) => {
   return (
-    <div className="announcement-card__container card shadow-sm mt-3 p-2 d-flex border-0">
+    <div
+      className={`${styles['announcement-card__container']} card shadow-sm mt-3 p-2 d-flex border-0`}
```
Ulisseus, line 11: I choose to use camelCase everywhere because otherwise you need to change most css names (you can't use `styles.announcement-card__container` and etc). It's rather verbose but on the bright side it's clear that this rule comes from custom css file and not from bootstrap.

package.json
```diff
@@ -46,6 +45,7 @@
     "react-diff-viewer": "^3.1.1",
     "react-dom": "^17.0.1",
     "react-feather": "^2.0.9",
+    "sass": "^1.32.8",
```
Ulisseus, line 48: `node-sass` was causing problems with css modules. It's [recommended](https://github.com/sass/node-sass) to move to `dart-sass` (`sass`) anyways.

package.json
```diff
@@ -95,6 +95,7 @@
     "jest": "^26.6.3",
     "prettier": "^2.2.1",
     "react-test-renderer": "^17.0.1",
+    "sass-loader": "^10.0.1",
```
Ulisseus, line 98: It's `sass` dependency but you need to add it manually to config file.

scss/fonts.scss
```diff
@@ -0,0 +1,4 @@
+@font-face {
+  font-family: "PT MON";
```
Ulisseus, line 2: It's the same font. Css nodules use different format for importing fonts. You can't just `@import "../public/PTMono-Regular.ttf";`

scss/index.scss
```diff
@@ -46,4 +46,5 @@ p {
 
 // Local SCSS Imports
 @import './challengeMaterial.scss';
+@import './modalCard.scss';
```
Ulisseus, line 49: `modalCard` is global because it can't access 3rd party component. It modifies `react-bootstrap` styles and not its own.

scss/index.scss
```diff
@@ -46,4 +46,5 @@ p {
 
 // Local SCSS Imports
 @import './challengeMaterial.scss';
```
Ulisseus, line 48: What was the reason for making `challengeMaterial` styles global?

scss/fonts.scss
hwong0305: Please fix typo.

ggwadera: I think this font-family is incorrect, should actually be this:

```suggestion
  font-family: "PT Mono";
```

Ulisseus: Fixed.

ggwadera (approved):

songz (approved):

Issue #508, opened 2021-02-27
c0d3 doesn't work without javascript


Ulisseus:
Eternal loading screen.
It would be nice if at least landing page and lessons would be usable without js since they're almost fully static.

PR #557, opened 2021-03-20
update #508 - make landing page static

Ulisseus:
Static pages are great, they are fast and seo friendly. Currently when you open `c0d3.com` for the first time you need to wait for `getAppQuery` while staring at loading spinner. It's completely unnecessary.  

Index page has three components - `AppNav`, `Footer` and the `LandingPage`, but only `AppNav` needs query data. I changed `AppNav` to return unauthorized navigation bar by default, run query and update session state with the new data. Index page also handles redirects based on session but since there is no session I added redirect based on cookie (logging in creates simplest `loggedIn` cookie, logging out destroys it). 

`curriculum` page was changed from `withApollo` HOC to simple apollo query because it needs to specify fetch-policy. Without it curriculum page would show default unauthorized state to logged in users. 

All redirects are handled by next router now.

There are quite a few updated snapshots but they all are of the same pattern (from loading spinner to default unauthorized navigation bar).


__tests__/pages/profile/username.test.js
```diff
@@ -38,14 +38,13 @@ describe('user profile test', () => {
...
     ]
-    const { container, findByRole, queryByText } = render(
+    const { findByRole, queryByText } = render(
       <MockedProvider mocks={mocks} addTypename={false}>
         <UserProfile />
       </MockedProvider>
     )
     await findByRole('heading', { name: /loading/i })
     await waitForElementToBeRemoved(() => queryByText('Loading...'))
-    expect(container).toMatchSnapshot()
```
Ulisseus, line 48: I deleted this snapshot. It's flaky (it fails if you pc is under heavy load). It also unnecessary since I only need to check for loading spinner.

stories/pages/Curriculum.stories.tsx
```diff
@@ -54,27 +54,36 @@ export const WithAlerts: React.FC = () => {
   )
 }
 
-export const CompletedLessons: React.FC = () => {
-  const session = {
-    ...dummySessionData,
+export const CompletedLessons: React.FC<{}> = () => {
+  const session: Session = {
```
Ulisseus, line 58: It's actually quite funny. We used `dummySessionData` data in so many tests but its types are incorrect. `id` filed needs to be a string.

Ulisseus:
Check out vercel [deployment](https://c0d3-app-fzuywetdm-c0d3.vercel.app/). It loads faster, it works with javascript disabled and transitions between pages (logging in, logging out) are a bit smoother. 

Cons. Clicking on `C0d3` button when you're authorized triggers loading spinner. Unfortunately, this is  result of using `cache and network` fetch policy.

ggwadera:
* We do have session, and it's managed by `express-session` that already set some cookies on login (like `connect.sid`) so setting another cookie is not necessary imo. I was thinking of doing what you done to check the cookie for redirection, nice one. Only thing that should be fixed is to make the logout process clear the cookies.

* It would have been nice if you explained in the PR description what is the fetch policy you changed for the Apollo hook. For future readers, I found it [here](https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies):

  > `cache-and-network`: Apollo Client executes the full query against both the cache and your GraphQL server. The query automatically updates if the result of the server-side query modifies cached fields. Provides a fast response while also helping to keep cached data consistent with server data.

  * I'm on the fence about changing this cache policy. This means that every time it needs to ping the server to see if there's updated data for that huge `getApp` query, which only a small number of data might be modified.

* The nav bar being unauthenticated by default is bad UX imo. I think we should move the authenticated user query into that component so it doesn't affect static generation for other pages. We could make use of React's contexts to avoid having to pass that data as props if other components need it.

Ulisseus:
I thought about using `connect.sid` instead of creating new cookie and decided against it. It's a server only cookie for a reason and opening it to a browser could lead to security problems. 

About new fetch-policy, I can't find a better way to do it. Currently we're fetching this huge `getAppQuery` once and then it's being served from cache. It's great if data stays the same, but it won't work if you need to update something without clearing the cache first. I'd like to decouple lesson/challenge data from session in the future though. At least nothing will change for the end user except for more traffic. 

AppNav component already uses query. Default isn't hardcoded, it's react state. This component runs its own `getAppQuery` and updates the state when it returns. Content provider is unnecessary IMO because client cache is shared between all components.

songz (approved):

Ulisseus:
I remember that we decided to wait before merging this pr last spring but what for? Decoupling session from getAppQuery? Dropping `withApollo`? Something else? I forgot.

__tests__/pages/404.test.js
```diff
@@ -1,22 +1,15 @@
...
 
 describe('404 Error Page', () => {
   it('should render', async () => {
-    const { container, queryByRole } = render(
+    const { container } = render(
       <MockedProvider>
         <NotFound />
       </MockedProvider>
     )
-    await waitForElementToBeRemoved(() =>
```
Ulisseus, line 17: This PR updates NavBar so there is no loading spinner anymore.

pages/login.tsx
```diff
@@ -78,13 +79,14 @@ export const Login: React.FC<LoginFormProps> = ({
 }
 
 const LoginPage: React.FC = () => {
+  const router = useRouter()
   const [loginErrors, setLoginErrors] = useState<string[]>([])
   const [loginUser, { data, error }] = useMutation(LOGIN_USER)
```
ggwadera, line 84: Found a solution for the stale cache after login. See the Apollo links below for a more detailed explanation.

After completing the login mutation, we can tell Apollo to refetch a query and update its cache using the `refetchQueries` option. So instead of needing to check to server to update the cache with that fetch policy, we can simply update it after login since a session will now exist. And also no more loading spinner when on the curriculum page 🙂

Reference: 
- [When To Use Refetch Queries in Apollo Client](https://www.apollographql.com/blog/when-to-use-refetch-queries-in-apollo-client/)
- [Advanced topics on caching - Rerunning queries after a mutation](https://www.apollographql.com/docs/react/caching/advanced-topics/#rerunning-queries-after-a-mutation)

```suggestion
  const [loginUser, { data, error }] = useMutation(LOGIN_USER, {
    refetchQueries: [
      {
        query: GET_APP
      }
    ]
  })
```

pages/curriculum.tsx
```diff
@@ -7,27 +7,28 @@ import AnnouncementCard from '../components/AnnouncementCard'
...
+const announcements = [
+  'To make space for other students on our servers, your account will be deleted after 30 days of inactivity.',
+  'Take each lesson challenge seriously and do them over and over again until you can solve them. With the exception End to End, all challenges are questions and exercises taken from real interviews.',
+  'This lesson will not only prepare you for interviews, but it will also help teach you the skills that you need to become an effective engineer.',
+  'After completing Foundations of JavaScript, Variables & Functions, Array, Objects, End to End, HTML/CSS/JavaScript, React/GraphQL/SocketIO, you will be technically ready to contribute to our codebase.'
+]
+export const Curriculum: React.FC<{}> = () => {
+  const { loading, error, data } = useGetAppQuery({
+    fetchPolicy: 'cache-and-network'
+  })
```
ggwadera, line 21: This won't be needed anymore, see my other comment on the login page.

```suggestion
  const { loading, error, data } = useGetAppQuery()
```

@types/helpers.d.ts
```diff
@@ -25,4 +25,5 @@ export interface UserSession extends Session {
 
 export interface Context {
   req: LoggedRequest
+  res: Response
```
ggwadera, line 28: We should be using Next's API types here instead of Express. `NextApiRequest` and `NextApiResponse`.

https://nextjs.org/docs/basic-features/typescript#api-routes
https://nextjs.org/docs/api-routes/api-middlewares#extending-the-reqres-objects-with-typescript

components/AppNav.tsx
```diff
@@ -99,19 +111,27 @@ const UnAuthButton = () => (
...
+    if (!session || _.get(session, 'user.username', null) === null) {
       return <UnAuthButton />
+    }
 
     const initial = ''
     // TODO: replace with typing
     const username = _.get(session, 'user.username', '')
 
-    return <AuthButton username={username} initial={initial} />
+    return <AuthButton username={username} initial={initial} router={router} />
```
ggwadera, line 132: Why pass the router as props? There's no problem in using the `useRouter` hook in more than one component.

helpers/controllers/authController.ts
```diff
@@ -61,6 +62,11 @@ export const login = async (_parent: void, arg: Login, ctx: Context) => {
     const cliToken = { id: user.id, cliToken: user.cliToken }
 
     session.userId = user.id
+    ctx.res.setHeader(
+      'Set-Cookie',
+      `loggedIn=true; expires=${new Date(Date.now() + ONE_WEEK)}; Path=/`
+    )
```
ggwadera, line 68: Why not directly set this cookie client-side instead of making the server set it through the response? It kinda makes more sense too because it's only a client-side check. You can do it with `document.cookie`.

helpers/controllers/authController.ts
```diff
@@ -84,7 +90,10 @@ export const logout = async (_parent: void, _: void, ctx: Context) => {
         error: 'Session Error'
       })
     }
-
+    ctx.res.setHeader(
+      'Set-Cookie',
+      `loggedIn=false; expires=${new Date(Date.now())}; Path=/`
+    )
```
ggwadera, line 96: Why have a false value? Does it mean something else? Can't we simply clear the cookie on logout?

Ulisseus:
I really like `refetchQueries` option, it makes everything so much simpler. Thanks. I moved `loggedIn` state to localstorage too, if redirects are handled client side it makes sense to keep data locally. 

Redirects need some work but I can't figure out how to make them flawless without server side rendering (and using cookies again). 

Go to vercel, login, and try to open index page in new tabs a few times (maybe with no cache and with throttling). It goes like this, you see `Become a Software Engineer the Hard Way`, then it's loading spinner and then it's curriculum page. What I'd like to do is to redirect to curriculum even before index page loaded. Also, browser downloads all landing pages images despite redirect (maybe it depends on browser). I tried to use `window.stop` but it wasn't perfect (page still loads but not the images). It also caused rare errors so I choose not to use it.

ggwadera:
> Also, browser downloads all landing pages images despite redirect

You can try making the images lazy loaded, I think `next/image` does that by default but not 100% sure.

https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes

pages/index.tsx
```diff
@@ -2,20 +2,14 @@ import * as React from 'react'
...
-import _ from 'lodash'
-
-const IndexPage: React.FC<GetAppProps> = ({ data: { loading, session } }) => {
+import LoadingSpinner from '../components/LoadingSpinner'
+const IndexPage: React.FC<{}> = () => {
   const router = useRouter()
-
-  if (loading) return <LoadingSpinner />
-  if (session) {
+  if (process.browser && window.localStorage.getItem('loggedIn')) {
```
ggwadera, line 9: More assured way to check if it's a browser.

```suggestion
  if (window && window.localStorage.getItem('loggedIn')) {
```

Ulisseus: This was my first approach but next throws errors `ReferenceError: window is not defined`. I'm not sure why it works this way now because I remember using this check in `AppNav` component for redirection. 
Alternatively I could use useEffect on mount([]) but process.browser should be a bit faster (it doesn't wait for a full load).

Ulisseus:
>You can try making the images lazy loaded, I think next/image does that by default but not 100% sure.

We need to remade all landing page images into next image someday but it should be another pr.

pages/index.tsx
ggwadera: Alright I see what you mean, but still `process` is not a standard object on browsers so it can confuse someone else that may work on this. There's a workaround you can use with `typeof` that won't cause that error because it doesn't evaluate `window`, only its type. See the suggestion below.

```suggestion
  if (
    typeof window !== 'undefined' &&
    window.localStorage.getItem('loggedIn')
  ) {
```

Besides that I think the `useEffect` would be more elegant and the difference in speed (if it exists) would be negligible since with this approach the redirection still only happens after the page is loaded, in both cases the JS redirection will only occur after the static HTML is rendered and the React scripts are downloaded and executed.

```suggestion
  useEffect(() => {
    localStorage.getItem('loggedIn') && router.push('/curriculum')
  }, [])
```

I did some quick tests here to compare total load times with redirection for these 3 cases. For the tests I disabled the cache and used the fast 3G network speed throttling. Then I went to the `/` location and waited until it redirected and the entire curriculum page loaded.

- With `useEffect`

  |Run|DOMContentLoaded|Load|Finish|
  |----|----------------|----|------|
  |1|1.04 s|9.91 s|10.03 s|
  |2|1.01 s|9.88 s|9.92 s|
  |3|1.02 s|9.85 s|10.01 s|

- With `if (typeof ...)` (without the loading spinner return)

  |Run|DOMContentLoaded|Load|Finish|
  |----|----------------|----|------|
  |1|1.02 s|9.86 s|9.95 s|
  |2|1.01 s|8.56 s|9.99 s|
  |3|1.01 s|8.59 s|9.98 s|

- With `if (process.browser ...)` (with the loading spinner return)

  |Run|DOMContentLoaded|Load|Finish|
  |----|----------------|----|------|
  |1|1.00 s|4.41 s|9.98 s|
  |2|1.02 s|4.43 s|10.02 s|
  |3|1.02 s|4.43 s|10.01 s|

As you can see the difference is not that significant and could be attributed to random variance. Also the DOMContentLoaded and Load times sometimes varied significantly in all cases, which in some runs the DOMContentLoaded even got below 1 s. The biggest difference is for the Load time in the last case, but that is due to returning the loading spinner, if I include the return in the second case the Load times are similar.

But still, the landing page flash after redirection is an issue we should try to find a solution, but it can be done later.

Ulisseus: OK, I changed it to `useEffect`.

ggwadera (approved):

Issue #499, opened 2021-02-16
[Feature] See star comments on user profile


ggwadera:
Currently the comments that the users submit when giving a star aren't available through the app UI.

These comments should be publicly displayed on the profile of the user that received the star.

kondanna:
it's not giving me the option to assign to myself, so I'll comment that I'm working on this issue

ggwadera:
> it's not giving me the option to assign to myself, so I'll comment that I'm working on this issue

I invited you to the organization, check your email. You should be able to assign yourself after accepting.

kondanna:
feature merged into master.

PR #561, opened 2021-03-22
update #499 - adds comments to user profile

kondanna:


@types/lesson.d.ts
```diff
@@ -27,5 +27,9 @@ export type Star = {
   studentId: number
   mentorId: number
   lessonId: number
+  lessonTitle: string
```
anthonykhoa, line 30: We shouldn't have to be using from this file. Typescript definitions for data returned from graphql requests are in our codegen generated [graphql/index.tsx](https://github.com/garageScript/c0d3-app/blob/master/graphql/index.tsx) file

pages/profile/[username].tsx
```diff
@@ -78,6 +79,16 @@ const UserProfile: React.FC = () => {
     }
   })
 
+  const lessonStatus: LessonStatus[] = _.get(data, 'userInfo.lessonStatus', [])
+  const profileStars: Star[] = []
+  lessonStatus.forEach(({ starsReceived }) => {
```
Comment (anthonykhoa) [inline pages/profile/[username].tsx]: You are using the `profileStars` variable and the `forEach` to do a combination of reducing and filtering. You should separate out this logic, and use `filter` along with a more appropriate higher order function so that your code becomes more explicit and clear about what is happening. It can be made even more simple by removing the nested `forEach` inside with a spread operator instead.

```
const validProfiles = lessonStatus.filter(({ starsReceived }) => starsReceived)
const profileStars: Star[] = validProfiles.reduce((acc, { starsReceived }) => {
  acc.push(...starsReceived)
  return acc
}, [])
```

Comment (anthonykhoa) [reply pages/profile/[username].tsx]: Ah, I see you changed your code before I finished making this comment. Lemme look the changes now

helpers/controllers/userInfoController.ts
```diff
@@ -5,7 +5,7 @@ type Username = {
   username: string
 }
 
-const { User, UserLesson, Submission, Star } = db
+const { Lesson, User, UserLesson, Submission, Star } = db
```
anthonykhoa, line 8: Can you separate out the back end and front end code for this PR? That will make things more organized and easier to review.

pages/profile/[username].tsx
```diff
@@ -107,11 +110,20 @@ const UserProfile: React.FC = () => {
       },
       {}
     )
+
     let starsReceived = [] as Star[]
     if (lessonStatusMap[String(lesson.id)]) {
       starsReceived = lessonStatusMap[String(lesson.id)].starsReceived as Star[]
     }
 
+    lessonStatus.forEach(({ starsReceived }) => {
```
Comment (anthonykhoa) [inline pages/profile/[username].tsx:119]: You are using the profileStars variable and the forEach to do a combination of reducing and filtering. You should separate out this logic, and use filter along with a more appropriate higher order function so that your code becomes more explicit and clear about what is happening. It can be made even more simple by removing the nested forEach inside with a spread operator instead.

const validProfiles = lessonStatus.filter(({ starsReceived }) => starsReceived)
const profileStars: Star[] = validProfiles.reduce((acc, { starsReceived }) => {
  acc.push(...starsReceived)
  return acc
}, [])

kondanna:
Closing this to submit backend & frontend logic separately under branch 499 (backend) & 499/comments (subbranch)

PR #562, opened 2021-03-22
Refactor loading spinner tests

ggwadera:
This PR just reorganizes some of the `expect loading spinner to render` tests as they don't need snapshots and are essentially the same repeated code, with only the rendered component differing between each one. As seen from some of the previous PRs, these snapshots were causing some issues where the tests failed for some reason. Removing these snapshots should help with this.

* Created new `expectLoading(Component: React.ComponentType)` test utility function to simplify loading spinner tests.
  * It uses a single mocked request and response for `MockedProvider` because it doesn't really make a difference to test only for the loading spinner to appear.
  * I replaced every test for this to use this new function and removed the obsolete snapshots.

* Moved `withNextRouter` from `testUtils/` to `__tests__/utils/` to help with organization.
  * Updated all the imports that used this function.

* Changed Jest configuration to ignore the `__tests__/utils/` folders for tests and coverage.


ggwadera:
The Vercel check is not updating for some reason even after I approved the deployment. Just ignore that since the build check passed this can be merged with no problems.

Ulisseus:
Can `components/admin/AdminLayout.test.js` and `containers/withQueryLoader.test.js` be refactored with new `expectLoading` function?

ggwadera:
> Can `components/admin/AdminLayout.test.js` and `containers/withQueryLoader.test.js` be refactored with new `expectLoading` function?

Changed `components/admin/AdminLayout.test.js` with my new function. Thanks 👍

`containers/withQueryLoader.test.js` I think is fine the way it is since it only checks for the return value. When I tried to render it errors out.

songz (approved):

anthonykhoa (approved):

PR #570, opened 2021-03-24
[DNM] Update #499 - adds backend for star comments in user profile

kondanna:
adds additional fields: studentName, studentUsername, lessonTitle, lessonOrder,  for
query { userInfo(<user>) { lessonStatus { starsReceived { <additional fields> } } } }
in order to show student initials & username, lesson name, and comment on user profile

### DIFF
--- a/graphql/typeDefs.ts
+++ b/graphql/typeDefs.ts
@@ -170,8 +170,12 @@ export default gql`
   type Star {
     id: String!
     studentId: Int
+    studentUsername: String
+    studentName: String
     mentorId: Int
     lessonId: Int
+    lessonTitle: String
+    lessonOrder: Int
     comment: String
   }
 `
--- a/helpers/controllers/userInfoController.ts
+++ b/helpers/controllers/userInfoController.ts
@@ -5,7 +5,7 @@ type Username = {
   username: string
 }
 
-const { User, UserLesson, Submission, Star } = db
+const { Lesson, User, UserLesson, Submission, Star } = db
 
 export const userInfo = async (_parent: void, args: Username) => {
   const username = _.get(args, 'username')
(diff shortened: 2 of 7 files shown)


graphql/typeDefs.ts
```diff
@@ -170,8 +170,12 @@ export default gql`
   type Star {
     id: String!
     studentId: Int
+    studentUsername: String
+    studentName: String
     mentorId: Int
     lessonId: Int
+    lessonTitle: String
+    lessonOrder: Int
```
anthonykhoa, line 178: It looks like you use `lessonOrder` as a string in the `helpers/controllers/userInfoController.ts` file, but here you are declaring it as an integer. According to graphql specs, graphql will coerce data types as needed into the one defined in the schema in order to `uphold the contract the scalar type describes`

See here, and search `result coercion` for more information:
https://spec.graphql.org/June2018/

kondanna: thanks, fixed!

also just added id to userInfo/starsReceived

helpers/controllers/userInfoController.ts
```diff
@@ -35,11 +35,19 @@ export const userInfo = async (_parent: void, args: Username) => {
...
+      ]
     })
   ])
   const starMap = starsReceived.reduce((map: any, star: any) => {
     map[star.lessonId] = map[star.lessonId] || []
+    star.dataValues.studentUsername = _.get(star, 'student.username', '')
+    star.dataValues.studentName = _.get(star, 'student.name', '')
+    star.dataValues.lessonOrder = _.get(star, 'lesson.order', 0) // lessonOrder: Int
+    star.dataValues.lessonTitle = _.get(star, 'lesson.title', '')
     map[star.lessonId].push(star.dataValues)
```
ggwadera, line 51: I sent PR #596 to replace Sequelize on this file, could you wait until that PR is merged and then use Prisma for your code? Sorry about that, but this way we don't add any more tech debt to be fixed later. 

Just FYI, the lesson join for a star won't work on Prisma until we change it to be a foreign key on the table. But the stars are already grouped by lesson in this `starMap` object and also on the `lessonStatus` value on the response. So having another table join here for the lesson is not needed in my opinion. The front-end component already does some work to join the lessons data in order to display the progress from the `lessonStatus` list.

In the future I think this entire query should go through some refactoring. There's a lot of values that go in the response that are not used and just add to the size. For example, the code for each submission from the user is in the response, yet it serves no purpose right now. Also the work that the front-end could be moved into the server response. But that's a thing for another PRs.

kondanna: yup, that works.

kondanna:
new version will use Prisma

PR #572, opened 2021-03-24
Updates 571 - ProfileSubmissions Component under pages/profile now styled with scss module

kondanna:


kondanna:
gotta resubmit, i accidentally pushed then deleted !! yarn.lock. we don't want that.

PR #583, opened 2021-03-25
[DNM] Add scripts to setup migrations

ggwadera:
# 🚨🚨🚨 ALERT 🚨🚨🚨

**BEFORE MERGING THIS PR A DATABASE BACKUP WILL BE NEEDED. AFTER APPROVING I WILL DO THE BACKUP AND MERGE THE PR SO WE CAN RECOVER IF SOMETHING BREAKS**

# PR Description

This PR adds a new script to help run Prisma migrations with some checks to avoid breaking stuff.

Since some commands can alter the database, and now that the preview deployment is connected to the database, it's important that these are only executed on the production deployment in case a PR is accidentally deployed to preview. This is done by checking the `VERCEL_ENV` environment variable that Vercel provides (https://vercel.com/docs/environment-variables#system-environment-variables).

Also a little workaround is needed to make Prisma CLI work on Vercel right now without adding a new environment variable. This is explained on the `setup` section below, but basically it's just using our current envs to generate a Postgres URL connection string. Fortunately the environment variables are available outside the build and runtime so this can be done. This is executed only when running at Vercel by checking the `VERCEL` environment variable.

The `vercel-build` script on `package.json` is called during the project build when deploying at Vercel. This was used to insert the migrations commands.

I tested this script on my Vercel account deploying this project and it worked without any issues.

What the script does:

* `baseline`: https://www.prisma.io/docs/guides/application-lifecycle/add-prisma-migrate-to-a-project#baseline-your-production-environment
  * To start doing migrations Prisma needs to create a new table on the database to keep track of the migrations that were applied. The baseline will do this.
  * Because applied migrations will make the command throw an error, I added a check for this to ignore this and only throw if it's a different error.
  * **RUNS ONLY ON PRODUCTION**

* `migrate`: https://www.prisma.io/docs/guides/deployment/deployment/#deploying-database-changes-with-prisma-migrate
  * This will deploy the pending migrations to the production database.
  * **THIS PR DOES NOT INCLUDE ANY MIGRATIONS**, this is only the initial step so we can write migrations in the future, so no data will be modified.
  * On my PR #566 that I added Prisma I had included some migrations, I removed them here to be brought back when appropriate.
  * **RUNS ONLY ON PRODUCTION**

* `setup`: this creates an `.env` inside the `prisma` folder so the CLI can connect to the database, using our current environment variables for the database. This was done for the reasons below:
  * Prisma CLI uses the database configuration from the `schema.prisma` file to connect we can't override it programmatically like I did for the client.
  * Prisma also requires an URL like `postgresql://blabla` to connect.
  * Our current production database envs are stored as secrets on Vercel and I can't recover them to insert another environment variable for that.
  * **RUNS ONLY ON VERCEL**, this is executed on the `postinstall` script which is called after installing the project dependencies.


kondanna (approved):

Ulisseus:
Maybe it's just me but I think that things are moving too fast. 

Currently have have two problems with db - crazy schemas and typeless data from sequelize. I expected that we will slowly depreciate sequelize first and then start fixing schemes. Even if migration is successful how it will interact with current sequelize setup?

ggwadera:
As I see it, any migration will also come with a replacement for Sequelize code to avoid breaking any related thing. These migrations will be needed because the type safety Prisma gives to us doesn't allow some of the things that are currently done with Sequelize which are handled kinda behind the scenes.

Ulisseus:
> These migrations will be needed because the type safety Prisma gives to us doesn't allow some of the things that are currently done with Sequelize which are handled kinda behind the scenes.

Makes sense. One thing, before actually doing any migrations we should make sure that everyone have access to production database/logs so we won't repeat bcrypt incident. I believe our production db is on digital ocean which still requires song's email. Is it possible to ssl to prod db? Is there enough info in credentials channel to do so?

ggwadera:
I have ssh access. If anyone needs it me or Song can give access, just need your ssh public key.

Ulisseus (approved):

PR #588, opened 2021-03-26
Update husky to v5, add lint-staged

ggwadera:
- Updated [`husky`](https://github.com/typicode/husky) to version 5.
  - [Summary of changes](https://blog.typicode.com/husky-5/)
    - Git hooks are much faster.
    - Optional install through the `postinstall` or `prepare` script in `package.json`.
    - Hooks are now configured in their corresponding files (like `.husky/pre-commit`) instead of in `package.json`
  - Changed the hooks scripts to lint/format using `lint-staged`.

- Installed [`lint-staged`](https://github.com/okonet/lint-staged)
  - `lint-staged` run linters against staged git files and helps keep the code formatted correctly when commiting.
  - Instead of running the linter against the whole project to check formatting like we used to, `lint-staged` makes it run only on the staged files making it faster.
  - Added automatic lint/formatting to `*.{js,jsx,ts,tsx}`, `*.{json,md,scss}` and `*.svg` for using svg optimizer.
  - Added some more overrides on `.prettierrc.js` to make prettier work with markdown and json.


package.json
```diff
@@ -122,20 +124,17 @@
...
     ]
   },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm run lint && npm run test && npm run type-check",
-      "pre-push": "npm run lint && npm run test && npm run type-check"
-    }
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx}": "eslint --fix",
+    "*.{json,md,scss}": "prettier --write",
```
Ulisseus, line 137: Does it support `mdx` files formatting?

.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run type-check && npm test
```
Ulisseus, line 4: Currently there is lint check in pre push. Can dropping it cause any problems?

package.json
ggwadera: I'm not sure. I'll look into it.

.husky/pre-push
ggwadera: The linter will run on the files you have staged when you commit, so I don't see the need to run before pushing, only if you skip the hook when commiting.

package.json
ggwadera: There's a mdx plugin for eslint that we can use

https://github.com/mdx-js/eslint-mdx

Maybe you could include in your mdx PR.

Ulisseus (approved):

songz (approved):

kondanna (approved):

Issue #545, opened 2021-03-12
Fix Sequelize model and database types


ggwadera:
# The issues

* Sequelize currently does not offer any type hint at all, which may lead to bugs in code working with the database.
  * Every Sequelize query in the codebase returns `any` as the type, needing a cast to be able to work with types.
  * From the Sequelize [docs](https://sequelize.org/v5/manual/typescript.html): *"As Sequelize heavily relies on runtime property assignments, TypeScript won't be very useful out of the box. A decent amount of manual type declarations are needed to make models workable."*

* Some of the database types are confusing or just not the appropriate ones. Examples: 
  1. Users table:
      * ~~`isAdmin` is defined as `varchar` containing the values `true` or `false` as strings, but should be `boolean` and also `NOT NULL`. It should also default to `false`.~~ Fixed in #646.
      * ~~`email`, `username` and `name` which are required at signup are also nullable.~~ Fixed in #748
      * `email` should be unique.
      * `username` should be unique.
  2. Lessons table:
      * ~~`title`, `description` and `order` fields are required but are nullable.~~ Fixed in #749
  3. Challenges table:
      * ~~`title`, `description` and `order` fields are required but are nullable.~~ Fixed in #746
  4. Submissions table:
      * ~~`userId`, `challengeId` and `lessonId` (foreign keys) are nullable but are required to create a submission.~~ Fixed in #716.
      * `diff` column could probably also be not null.
      * ~~`status` column only holds 4 unique values: `needMoreWork`, `open`, `passed` and `underReview`. Default value should be `open`. This column should be a Enum type or use check constraints:~~ Fixed in #716.
        ```sql
		CREATE TABLE submissions (
		    ...
		    status VARCHAR CHECK (status IN ('needMoreWork', 'open', 'passed', 'underReview'))
		    ...
		)
        ```
  5. UserLessons table:
      * `isPassed`, `isTeaching` and `isEnrolled` columns which from the name makes you think are `boolean`, actually are `varchar` that holds a timestamp string like `1588199982766`. #988 
  6. Stars table:
      * ~~`lessonId`, `studentId` and `mentorId` should be not null since they are foreign keys.~~
      * ~~Inspecting the table actually reveals that `lessonId` is not even a foreign key and only holds the lesson ID int value.~~
  7. All tables:
      * ~~`createdAt` column does not have a default value. I don't know if Sequelize does something in the background to assign a value for this, but we never set it when creating something.~~ Fixed in #608.

[...]

PR #592, opened 2021-03-28
Refactor alerts query resolver and alerts controller with Prisma

ggwadera:
Part of #545

# Changes in this PR

- Refactored `alerts` query resolver with Prisma, replacing Sequelize.
- Refactored `alertsController` file with Prisma, replacing Sequelize.
  - Also changed types to use the ones from `graphql/index.tsx`, generated from the query and mutation.
- Change GraphQL's `Alert` type and `removeAlert` mutation to make the id as `Int!` instead of `String!`
  - Updated related React components to use the new type to fix type errors from Typescript.
- Regenerated `graphql/index.tsx` file
- Updated tests

# TODO

- Because the `createdAt` columns in the database tables does not have a default value, currently we need to include it in the entity creation data by calling `new Date()` since it expects a `string` or `Date` value, and `Date.now()` returns a `number`. This requires writing a mutation to add a default value on the tables so we don't need to do that anymore.

### DIFF
--- a/components/Alert.tsx
+++ b/components/Alert.tsx
@@ -11,7 +11,7 @@ type Props = {
   icon?: string
   type?: string
   prompt?: string
-  onDismiss?: (id: string) => void
+  onDismiss?: (id: number) => void
 }
 
 const alertIconMap: { [type: string]: string } = {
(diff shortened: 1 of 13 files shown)


components/Alert.tsx
```diff
@@ -39,7 +39,7 @@ const Alert: React.FC<Props> = ({ alert, onDismiss }) => {
           </NavLink>
         )}
       </div>
-      {id && onDismiss && (
+      {onDismiss && (
```
ggwadera, line 42: `id` type is non-null, this check is unnecessary

kondanna (approved):

PR #595, opened 2021-03-29
Update #545 - fixes typeDefs for ChallengeMaterial component and switch imports to graphql/index instead

kondanna:
So I'm starting to work on the bug where reviewer's name stops showing up after a student clicks on their own profile, then goes to curriculum > lesson and clicks on a challenge.

Before that, I figure I'll change the typeDefs related to that file.

  type Challenge {
    id: String!
    description: String! << required, so not null
    lessonId: String! << foreign key, so not null
    title: String! << required, so not null
    order: Int! << required, so not null
  }

### DIFF
--- a/components/ChallengeMaterial.tsx
+++ b/components/ChallengeMaterial.tsx
@@ -1,10 +1,5 @@
 import React, { useState } from 'react'
-import {
-  Challenge,
-  ChallengeSubmissionData,
-  UserSubmission,
-  UserSubmissionsObject
-} from '../@types/challenge'
+import { Challenge, Submission } from '../graphql/index'
 import { LessonStatus } from '../@types/lesson'
 import NavLink from './NavLink'
 import Markdown from 'markdown-to-jsx'
@@ -23,6 +18,19 @@ type CurrentChallengeID = string | null
 type ReviewStatusProps = {
   status: string
   reviewerUserName: string | null
+}
+
+export type ChallengeSubmissionData = {
+  title: string
+  id: string
+  order: number
+  description: string
+  status: string
+  submission?: Submission
+}
+
+export type UserSubmissionsObject = {
+  [submissionId: string]: Submission
 }
 
 export const ReviewStatus: React.FC<ReviewStatusProps> = ({
@@ -294,7 +302,7 @@ export const ChallengesCompletedCard: React.FC<ChallengesCompletedCardProps> = (
 
 type ChallengeMaterialProps = {
   challenges: Challenge[]
-  userSubmissions: UserSubmission[]
+  userSubmissions: Submission[]
   lessonStatus: LessonStatus
   chatUrl: string
   lessonId: string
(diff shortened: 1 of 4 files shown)


graphql/typeDefs.ts
```diff
@@ -152,11 +152,11 @@ export default gql`
   }
 
   type Challenge {
-    id: String
-    description: String
-    lessonId: String
-    title: String
-    order: Int
+    id: String!
```
ggwadera, line 155: Let's start fixing this stupid thing of ids being strings haha

This will probably break some stuff after generating a new index file, you can then run `yarn type-check` to see what's broken.

```suggestion
    id: Int!
```

graphql/typeDefs.ts
```diff
@@ -152,11 +152,11 @@ export default gql`
...
+    description: String!
+    lessonId: String!
```
ggwadera, line 157: Same as the other comment.

```suggestion
    lessonId: Int!
```

components/ChallengeMaterial.tsx
```diff
@@ -25,6 +20,19 @@ type ReviewStatusProps = {
...
 }
 
+export type ChallengeSubmissionData = {
+  title: string
+  id: string
+  order: number
+  description: string
+  status: string
+  submission?: Submission
+}
```
ggwadera, line 30: I'd suggest changing this with a compound type since there's a lot of redundant stuff but it breaks some things here. If you wanna try to do it then great, but it's not needed in this PR.

```suggestion
export type ChallengeSubmissionData = Challenge & { submission: Submission }
```

Would probably be good to rename with with a better name too like `ChallengeWithSubmission`.

components/ChallengeMaterial.tsx
```diff
@@ -312,8 +320,8 @@ const ChallengeMaterial: React.FC<ChallengeMaterialProps> = ({
   }
   //create an object to evaluate the student's status with a challenge
   const userSubmissionsObject: UserSubmissionsObject = userSubmissions.reduce(
-    (acc: UserSubmissionsObject, submission: UserSubmission) => {
-      acc[submission.challengeId] = submission
+    (acc: UserSubmissionsObject, submission: Submission) => {
```
ggwadera, line 323: The type here is inferred from the array type so you don't need to specify it.

```suggestion
    (acc: UserSubmissionsObject, submission) => {
```

graphql/typeDefs.ts
kondanna: Good idea, I had to change a bunch of other things (as expected).

ggwadera:
dude, what happened to your PR? hahaha

you're changing like every file on the repo lol

image

kondanna:
Oh, I see now. those other files, yeah, that rebase really messed me up. I'll start over, shouldn't take *too* long.

kondanna:
gonna try to fix one type at a time

Issue #590, opened 2021-03-27
Curriculum sidebar is misaligned


Ulisseus:
It looks bad once you see it. 
![misaligned](https://user-images.githubusercontent.com/3845012/112723791-5ce2eb00-8f21-11eb-9579-1f58a03e586b.png)

kondanna:
now I can't unsee it

Ulisseus:
We decided to keep it as it is.

PR #597, opened 2021-03-29
close #548, #590 - update additional resources

### LINKED ISSUE #548: Update Additional Resources menu
"Github basics" link doesn't go anywhere. 
"Help & Support" and "Setup Instructions" are duplicates. Should we remove them? 
And even new students probably already know about stackoverflow (it's one of the top search results for anything coding related). 

I propose to changes it to something like this.
* [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Learn)
* [The Modern JavaScript Tutorial](https://javascript.info/). It's not really a tutorial, more like a guidebook but it's rather in-depth. 
* [The missing semester](https://missing.csail.mit.edu/). Various tools, git, vim, grep, tmux and etc.
* [Full Stack open](https://fullstackopen.com/en/)
* [Teach Yourself Computer Science](https://teachyourselfcs.com/)

Feel free to add/remove anything.

Ulisseus:
## What does this PR do?

It's various curriculum front end changes, so check [vercel](https://c0d3-app-cvc66di03-c0d3.vercel.app/curriculum) deployment first.

* Additional resources panel was updated
* Lesson images are no longer stretched when lesson description is longer than two lines
* ~~Sidebar always follows grid layout. Progress bar has the same height as first two lessons, announcements are four lessons long~~
* Curriculum page is finally mobile friendly (at least for unathorized users, lessons for logged in users still need some work). Progress bar is still there but announcement and additional resources are hidden on mobile. It was possible to insert them between lessons but I like this cleaner look more. TO-DO add a way to see them on mobile.


ggwadera:
What's up with this huge whitespace?

image

Ulisseus:
> What's up with this huge whitespace?

It's not supposed to be like this. This is how it looks on my local machine. Weird. I'll check what's wrong tomorrow.
sidebar

ggwadera:
I get your intentions to get everything aligned, but unfortunately this won't work for our current layout. The two columns dimensions varies depending on the viewport and if there's an authenticated user so more things are rendered, making the flex rows get separated like on my previous comment.

The code also gets pretty ugly and unnecessarily complicated with a lot of divs and slices just to keep the two things aligned. The lessons and the sidebar layouts should be independent of each other because they're different things.

pages/curriculum.tsx
```diff
@@ -82,15 +82,28 @@ export const Curriculum: React.FC<GetAppProps> = ({ data }) => {
...
+      <div className="row">
+        <div className="col-xl-8">{lessonsToRender.slice(5, 7)}</div>
+        <div className="d-none d-xl-block col-xl-4">
           <AdditionalResources />
         </div>
       </div>
+      <div className="row">
+        <div className="col-xl-8">{lessonsToRender.slice(7)}</div>
+      </div>
     </Layout>
```
ggwadera, line 107: Sorry but this is bad. Instead of 2 divs you got a bunch of them now, and the layout also depends on the number of lessons, which I agree won't change on a short time, but when it changes then this will be another source of issues.

ggwadera (changes requested):

pages/curriculum.tsx
songz: the rest of the PR is great, I don't think its a good idea for annoucements to be aligned with each lesson... if announcement changes, you get weird white spaces.

Ulisseus:
>I get your intentions to get everything aligned, but unfortunately this won't work for our current layout. The two columns dimensions varies depending on the viewport and if there's an authenticated user so more things are rendered, making the flex rows get separated like on my previous comment.

Check updated vercel. Logged in user, 1080p resolution, I slowly reduce viewport width. Nothing breaks until I hit mobile breakpoint, then layout changes. The same thing happens if I gradually increase font size. 
Maybe it's a browser thing? To my surprise there is a minor but annoying difference in how [chromium](https://user-images.githubusercontent.com/3845012/112964049-a2392f80-9150-11eb-96a7-36cb50811b0a.png) and [mozilla](https://user-images.githubusercontent.com/3845012/112964167-bed56780-9150-11eb-8c02-b6e4ac8b805f.png) handle flexboxes. 

I agree that curriculum page becomes more complicated but it's pretty much standard bootsrap grid layout. 
```
<div className="row">
  <div className="column A style">{children}</div>
  <div className="column B style">{children}</div>
</div>
```
It would be nice to keep simple functional `lessons.map` and three components but I doubt it's possible even with implicit grid layout magic. Slicing lessons won't break anything, if index is out of range slice will create empty array which will result in empty div. 

So question is do you guys like the look of new layout or should I revert it back? Because I don't see technical reasons why it's worse.

pages/curriculum.tsx
Ulisseus: Announcement change is a problem yes. I used vertical flexbox so it should look fine as long as total announcement height won't become bigger than four lessons. As far as I know it's not possible to create flexbox of two items with total height equal the smaller one (with pure CSS at least). 

Was announcement message ever changed?

ggwadera:
> Because I don't see technical reasons why it's worse.

- You're coupling the entire layout together making it hard to change something in the future. If, for example, we change the announcement then we would also have to worry about the layout not breaking and keeping everything aligned. Right now this is not something we have to worry about.

- The same argument applies for the lessons card. You're coupling the layout to the lessons data we get from the DB and is this case it's worse, because the data being from outside the code, the layout can break unintentionally by a simple change on the database. If someday we decide that we should add another lesson, this will require fixing the page to display it, which is something we shouldn't have to worry about.

While on the visual side this looks good, on the development side this is not a good idea overall. This will just make the code more complicated and fragile to changes. A good practice is to make code as modular as possible so a change in one component will not affect another unrelated one. My suggestion is to revert this change for the moment until there's a better solution.

Ulisseus:
> If, for example, we change the announcement then we would also have to worry about the layout not breaking and keeping everything aligned.

Announcement card is '3 lessons long', it's also a column flex and in will scroll in case of overflow. So there is some future proofing for simple changes. It won't fit something really big but it wasn't designed for it. Ideally I'd love it to "stick" to closest lesson but I don't know how to do it (if it's even possible). 

 >If someday we decide that we should add another lesson, this will require fixing the page to display it, which is something we shouldn't have to worry about.

Will it? Why? I add new lessons/change their order locally and everything is working as it should.

I don't feel strongly about this change, so I won't argue further. It's very easy to revert it to old layout while keeping it mobile friendly if needed.

Ulisseus:
I reverted [the old layout](https://c0d3-app-cvc66di03-c0d3.vercel.app/curriculum) back. I changed css for announcement titles a bit.

So does anyone know if we really delete inactive users after 30 days or I should remove this section?

ggwadera (approved):
Thanks for reverting that layout. The additional resources looks great, nice work.

ggwadera:
> So does anyone know if we really delete inactive users after 30 days or I should remove this section?

Honestly? We don't delete anything haha. I'm not even sure how we would track inactivity to do that. For me you could remove it, what do you think @songz ?

Ulisseus:
>We don't delete anything

I guess it's anti procrastination message. I'll merge it as it is and edit announcement if needed later.

PR #605, opened 2021-03-31
Updates #545 - fixes Challenge typeDef

kondanna:
updated:
 
 type Challenge {
    id: Int!
    description: String! <-- required in form
    lessonId: Int! <-- foreignkey not null
    title: String! <-- required in form
    order: Int! <-- required in form
  }

### DIFF
--- a/pages/profile/[username].tsx
+++ b/pages/profile/[username].tsx
@@ -84,7 +84,7 @@ const UserProfile: React.FC = () => {
     const challengesStatus = challengeList.map(challengeInfo => {
       const challenge = challengeInfo || ({} as Challenge)
       const challengeSubmission = userSubmissions.find(
-        submission => +challenge.id === +submission.challengeId
+        submission => challenge.id === Number(submission.challengeId)
       )
 
       return {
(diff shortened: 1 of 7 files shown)


pages/profile/[username].tsx
```diff
@@ -82,9 +82,9 @@ const UserProfile: React.FC = () => {
     const lesson = lessonInfo || {}
     const challengeList = lesson.challenges || []
     const challengesStatus = challengeList.map(challengeInfo => {
-      const challenge = challengeInfo || {}
+      const challenge = challengeInfo || ({} as Challenge)
       const challengeSubmission = userSubmissions.find(
-        submission => challenge.id === submission.challengeId
+        submission => +challenge.id === +submission.challengeId
```
Comment (ggwadera) [inline pages/profile/[username].tsx:87]: Avoid using implicit conversion. Having an explicit conversion makes the code clearer as that little `+` might go unnoticed.

Since `challenge.id` is already `number`, you only need to convert `submission.challengeId`.

```suggestion
        submission => challenge.id === Number(submission.challengeId)
```

Comment (kondanna) [reply pages/profile/[username].tsx]: Ah, VSCode was slow in updating the Challenge type so it still showed Maybe string and flagged it. Thanks.

ggwadera (commented):
run `yarn lint:fix`, the `index.tsx` file is not following the lint rules.

ggwadera (approved):

PR #606, opened 2021-03-31
Updates #545 - fixes User typeDef

kondanna:
type User {
    id: Int!
    username: String!
    userLesson: UserLesson
    email: String!
    name: String!
    isAdmin: Boolean! 
    cliToken: String
  }

### DIFF
--- a/components/AppNav.tsx
+++ b/components/AppNav.tsx
@@ -115,7 +115,7 @@ const AppNav: React.FC<{}> = () => {
   const { data } = useGetAppQuery()
   useEffect(() => {
     if (data && data.session) {
-      setSession(data.session)
+      setSession(data.session as Session)
     }
   }, [data])
   const renderButtons = () => {
--- a/graphql/typeDefs.ts
+++ b/graphql/typeDefs.ts
@@ -112,11 +112,11 @@ export default gql`
 
   type User {
     id: Int!
-    username: String
+    username: String!
     userLesson: UserLesson
-    email: String
-    name: String
-    isAdmin: String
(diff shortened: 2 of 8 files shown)


graphql/typeDefs.ts
```diff
@@ -112,11 +112,11 @@ export default gql`
...
     id: Int!
-    username: String
+    username: String!
     userLesson: UserLesson
-    email: String
-    name: String
-    isAdmin: String
+    email: String!
+    name: String!
+    isAdmin: Boolean!
```
ggwadera, line 119: Until we make the `username`, `email` and `name` fields non-nullable on the database, this will conflict with the return types from prisma queries, and putting `as any` after it is not a good solution.

components/AppNav.tsx
```diff
@@ -115,7 +115,7 @@ const AppNav: React.FC<{}> = () => {
   const { data } = useGetAppQuery()
   useEffect(() => {
     if (data && data.session) {
-      setSession(data.session)
+      setSession(data.session as Session)
```
Ulisseus, line 118: Why is this needed? Apollo hook returns properly typed `GetAppQuery | undefined`, after check for `data&&data.session` typescript should infer proper type by itself.

kondanna: Ah, ignore this PR. The more recent submissions and challenge one takes priority. 

I probably needed to add annotation because typescript was yelling at me and that was the only way to make it disappear

kondanna:
see challengeType and submissionType PR. also starType will be updated first

PR #608, opened 2021-04-01
[DNM] [Migration] Add default value for `createdAt` column

ggwadera:
### 🚨 ATTENTION 🚨

Until we can merge migrations PRs like this with confidence, I wanna make sure that our production database stays safe. So just like the other one, I want to make a backup just before merging this. That's why the `dont-merge` tag is here.

---

Part of #545 

Now that we can apply migrations, let's try it out with a pretty simple and harmless one :)

This migration only changes the `createdAt` columns to have a default value of [`CURRENT_TIMESTAMP`](https://www.postgresql.org/docs/12/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT), which as the name already says, is the current timestamp when creating an entry in the database.

If we don't have that default value, Prisma complains about needing to have a `createdAt` field when saving a new entry in the database. Without that we need to do something like the code below everytime we create a new entry:

```js
prisma.user.create({
  data: {
    // ... all the user data
    createdAt: new Date()
  }
})
```

Since this migration only changes the default value, all our data in the database should be safe. As for the remaining Sequelize models, I believe this won't cause any issues too as I think Sequelize already sets a timestamp by itself.


package.json
```diff
@@ -21,7 +21,8 @@
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": "eslint --fix",
     "*.{json,md,scss}": "prettier --write",
-    "*.svg": "svgo --multipass"
+    "*.svg": "svgo --multipass",
+    "prisma/schema.prisma": "npx prisma format"
```
ggwadera, line 25: added this to format the schema automatically when commiting

songz (approved):

Issue #622, opened 2021-04-04
Migrate from withApollo and add static generation


Ulisseus:
**Why?**
Currently c0d3 is rather slow. Browser sends queries to graphql server and waits for them to come back while user is staring at loading spinner. Not optimal. 

Next js is capable of fetching data at build time. This way curriculum page would come with predefined lesson/challenge data (no loading spinner at all). 

**How?**
Unfortunately, withApollo package doesn't support new API, so first I need to add another way to provide browser with apolloClient. This will be the first PR.
Another PR will add server side redirects to index page. This will remove 'flickering' when authorized users go to c0d3.com.
And the last one will add static generation to curriculum page.

Ulisseus:
Closed. Curriculum fetches lesson data at build time. And there is no 'flickering' on index page anymore.

PR #623, opened 2021-04-04
update #622 - migrate from withApollo

Ulisseus:
**What does this PR do?**
It changes the way c0d3 server provides end user with apollo client. Previously we used `withApollo` package. I'm adding a custom integration (mostly taken from this [example](https://github.com/vercel/next.js/tree/canary/examples/with-apollo) with some small additions). This should not affect end users in any way.

~~I'm also adding persistent grapql [cache](https://github.com/apollographql/apollo-cache-persist). Previously if you hit f5 all your grapql data disappeared, now it's saved/restored from localStorage. This should decrease loading times a bit.~~


graphql/schema.test.js
```diff
@@ -0,0 +1,6 @@
+describe('schema generation', () => {
+  test('should create schema', () => {
+    const schema = require('./schema')
```
Ulisseus, line 3: Schema is used for server side generation. To generate data at build time you need active graphql endpoint but next js doesn't start api routes until it's finished building the app.

helpers/apolloClient.ts
```diff
@@ -0,0 +1,96 @@
...
+      })
+    } catch (error) {
+      console.error('Error restoring Apollo cache', error)
+    }
+  }
+}
+loadCache()
+function createIsomorphLink() {
+  if (typeof window === 'undefined') {
+    const { SchemaLink } = require('@apollo/client/link/schema')
```
Ulisseus, line 30: SchemaLink is used while generating data server side. While normal HttpLink is used in browser apollo client.

helpers/mail.ts
```diff
@@ -2,7 +2,7 @@ import _ from 'lodash'
 import mailgun from 'mailgun-js'
 
 const CLIENT_URL = _.get(process, 'env.CLIENT_URL', 'https://c0d3.devwong.com')
-const MAILGUN_API_KEY = _.get(process, 'env.MAILGUN_API_KEY', '')
+const MAILGUN_API_KEY = _.get(process, 'env.MAILGUN_API_KEY', 'test-key')
```
Ulisseus, line 5: There was already default `''` value, so I changed it a bit to simplify testing. Otherwise hoisting would complicate imports by a lot (to import schema you need to have valid API_KEY, but you can't define it before imports because every import is moved to the top).

pages/api/graphql.ts
```diff
@@ -18,7 +18,8 @@ const handler: any = nextConnect() // For session middleware. TODO: Need to defi
 const apolloServer = new ApolloServer({
   typeDefs,
   resolvers,
-  context: ({ req }) => ({ req }) // This lets GraphQL have access to sessions
+  context: ({ req }) => ({ req }), // This lets GraphQL have access to sessions
+  uploads: false
```
Ulisseus, line 22: This fixes discrepancy between schema generated and normal api routes graphql endpoint (which we'll be used for server side data fetching in the future). By default apolloServer accepts uploads, while schema-generated server does not .

Ulisseus:
It looks cache is causing bugs with loading spinner. I'll remove it until I change curriculum to server generation. 
spinner

pages/api/graphql.ts
hwong0305: Can you put this in the comment in the code?

graphql/schema.test.js
hwong0305: I think we should disable test for schemas. It doesn't make sense to test this.

ggwadera: @Ulisseus on `package.json` you can add the file on the ignore coverage path list to disable required tests

pages/api/graphql.ts
Ulisseus: Done.

ggwadera (approved):

PR #644, opened 2021-04-10
Remove `babel-plugin-inline-react-svg`

ggwadera:
Removed the [`babel-plugin-inline-react-svg`](https://github.com/airbnb/babel-plugin-inline-react-svg) dependency since it was being used basically for a single component: `Card`.

Replaced the inline svgs for the card icons with `</img>` tags.

That plugin is used to generate inline SVGs when transpiling with Babel, while also applying optimizations to the SVG image. But since we already have all our SVGs optimized with `svgo` this plugin is kinda useless. The `img` tag also helps with accessibility by having the `alt` attribute.

### DIFF
--- a/components/Card.tsx
+++ b/components/Card.tsx
@@ -1,25 +1,28 @@
 import React from 'react'
-import { ReactComponent as Checked } from '../public/assets/curriculum/icons/checked.svg'
-import { ReactComponent as Fail } from '../public/assets/curriculum/icons/exclamation.svg'
 
 type Props = {
   primary?: boolean
-  children?: React.ReactNode
   type?: 'success' | 'fail'
   text?: string
   classes?: string
   title: string
 }
 
-const iconProps: { [type: string]: string } = {
+const iconProps = {
   className: 'mb-4',
   width: '100px',
   height: '100px'
 }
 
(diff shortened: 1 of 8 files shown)


components/Card.tsx
```diff
@@ -1,25 +1,28 @@
...
 type Props = {
   primary?: boolean
-  children?: React.ReactNode
   type?: 'success' | 'fail'
   text?: string
   classes?: string
   title: string
 }
 
-const iconProps: { [type: string]: string } = {
```
ggwadera, line 14: The type here is inferred automatically, so I removed the manual declaration.

components/Card.tsx
```diff
@@ -1,25 +1,28 @@
 import React from 'react'
-import { ReactComponent as Checked } from '../public/assets/curriculum/icons/checked.svg'
-import { ReactComponent as Fail } from '../public/assets/curriculum/icons/exclamation.svg'
 
 type Props = {
   primary?: boolean
-  children?: React.ReactNode
```
ggwadera, line 7: The `children` type already comes with `React.FC<>`, so I removed this.

songz (approved):

hwong0305 (approved):

PR #645, opened 2021-04-10
Add custom discord component

Ulisseus:
Turns out our server has [public api](https://discord.com/api/guilds/828783458469675019/widget.json) on which we can build any component we like. Current discord widget is great but it looks a bit out of place with different styling and colors.  

[Preview](https://c0d3-app-dil9k70q9-c0d3.vercel.app/curriculum). [Storybook](https://deploy-preview-645--youthful-thompson-448332.netlify.app/?path=/story/components-discord--twenty-users).


components/DiscordBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+  id: string
+  username: string
+  status: string
+  avatar_url: string
+}
+interface discordJSON {
+  presence_count: number
+  members: Member[]
+}
+const ONE_MINUTE = 60 * 1000
```
Ulisseus, line 13: Members list is being updated every minute. Too fast?

components/DiscordBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+      setData((data as unknown) as discordJSON)
+      setTimeout(getData, ONE_MINUTE)
+    } catch (e) {
+      setTimeout(getData, ONE_MINUTE)
+    }
+  }
+  useEffect(() => {
+    getData()
+  }, [])
+  const reshuffle = (data: discordJSON) => {
```
Ulisseus, line 41: This reshuffles members list, slices the first 5 and renders it, so you will see different users in different order every time. Sorting isn't very expensive with our members count and it looks nicer this way.

components/DiscordBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <div className={`${styles['discord-list']}`}>{reshuffle(data)}</div>
+      </div>
+      <div className={`${styles['discord-join']} bg-primary text-center`}>
+        <a
+          className="font-weight-bold text-white"
+          href="https://discord.com/invite/5KZ99KAs"
+          target="_blank"
+          rel="noreferrer"
+        >
+          JOIN
```
ggwadera, line 79: Would be nice to have the entire `JOIN` section clickable instead of just the text. Move the anchor tag one level up.

components/DiscordBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+    try {
+      let data = await fetch(
+        'https://discord.com/api/guilds/828783458469675019/widget.json'
+      )
+      data = await data.json()
+      setData((data as unknown) as discordJSON)
+      setTimeout(getData, ONE_MINUTE)
+    } catch (e) {
+      setTimeout(getData, ONE_MINUTE)
+    }
```
ggwadera, line 36: Use `finally` so you don't have to repeat the timeout.

```suggestion
    } catch (e) {
      // maybe show some error message in this case?
    } finally {
      setTimeout(getData, ONE_MINUTE)
    }
```

components/DiscordBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+  return (
+    <div className="bg-white mt-3">
+      <div className="container">
+        <img src="/assets/discordLogo.svg" />
+        <div className="text-primary mt-n3 text-center">
+          <span className="font-weight-bold">
+            {data.members ? data.members.length : '0'}
+          </span>
+          {data.members && data.members.length === 1
```
ggwadera, line 64: From what I've seen `data.members` isn't gonna be null or undefined in the API response, just empty when no one's online. These checks here are unnecessary, since you already have it as non-null in the interface.

components/DiscordBar.tsx
```diff
@@ -0,0 +1,86 @@
...
+    presence_count: 1,
+    members: []
+  })
+  const getData = async () => {
+    try {
+      let data = await fetch(
+        'https://discord.com/api/guilds/828783458469675019/widget.json'
+      )
+      data = await data.json()
+      setData((data as unknown) as discordJSON)
```
ggwadera, line 32: This way it's better by using a const and without the double type casting.

```suggestion
      const data: discordJSON = await fetch(
        'https://discord.com/api/guilds/828783458469675019/widget.json'
      ).then(r => r.json())
      setData(data)
```

components/DiscordBar.tsx
Ulisseus: I probably need to update interface as well. I'm not sure if `data.members` field can't be undefined. From[ discord developer portal](https://discord.com/developers/docs/resources/guild) ` * These fields are only sent within the GUILD_CREATE event ` so under some rare scenario (server outage for example) it's possible to get a response without `guild.members' field.

components/DiscordBar.tsx
Ulisseus: I feel like there is no good way to do it. Ideally I'd simple chain promises but I'm forced to use `try/catch` for consistency. I removed ugly typescript casting by adding additional variable but it doesn't feel right either way.

components/DiscordBar.tsx
Ulisseus: Done.

components/DiscordBar.tsx
Ulisseus: I added error state. See [new storybook](https://deploy-preview-645--youthful-thompson-448332.netlify.app/?path=/story/components-discord--error).

components/DiscordBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [data, setData] = useState<State>({
+    members: []
+  })
+  const getData = async () => {
+    try {
+      const data = await fetch(
+        'https://discord.com/api/guilds/828783458469675019/widget.json'
+      )
+      const json: discordJSON = await data.json()
+      setData(json)
```
ggwadera, line 33: Check if `members` exists in the data you fetched to simplify the checks in the rendering code. You can chain the `.then()` part after the fetch, the `await` command will work the same. You can then define `members` as a non null value.

P.S.: I didn't test this code, let me know if it doesn't work.

```suggestion
      const json = await fetch(
        'https://discord.com/api/guilds/828783458469675019/widget.json'
      ).then(r => r.json())
      if ('members' in json) setData(json as discordJson)
      else throw new Error() // fallback to the catch block
```

components/DiscordBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <div className="d-inline-block">
+      <img src={avatar} className={`${styles['member-avatar']} mx-1`} />
+    </div>
+  )
+}
+const DiscordBar: React.FC = () => {
+  interface State extends discordJSON {
+    error?: string
+  }
```
ggwadera, line 23: Put types and interfaces outside the components scope.

components/DiscordBar.tsx
Ulisseus: I added new error case for no members data. Error during fetching will result in `Error fetching data...` message while successful response without members data is generic `Something went wrong ;(`

components/DiscordBar.tsx
Ulisseus: Done.

components/DiscordBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+  })
+  const getData = async () => {
+    try {
+      const data = await fetch(
+        'https://discord.com/api/guilds/828783458469675019/widget.json'
+      )
+      const json: discordJSON = await data.json()
+      json.members
+        ? setData(json)
+        : setData({ members: [], error: 'Something went wrong ;(' })
```
ggwadera, line 35: This way it's more confusing for anyone reading the code. 

On the interface you declare that the `members` field is always defined. But then you say `json` is of the type `discordJSON` and then check if `json.members` is defined or not, which from the interface you should expect that it is indeed defined. This is a bit counter intuitive.

What I suggested before is to check if the object conforms to the interface, and then set the state with the data, using the type assertion `as discordJSON` since you now know it's valid, otherwise show an error. This makes more sense because you're validating your assumptions about the data structure before assigning it.

Also, should the error message mean something different? Why not use the same one you already have in the `catch` block by throwing an error?

Another thing, prefer using `if/else` if you're not assigning the result of the ternary operator to a variable, it helps in making the code clearer. Shorter is not always better ;)

```ts
const json = await data.json()
if ('members' in json) setData(json as discordJSON)
else setData({ members: [], error: 'Something went wrong ;(' })
// or throw an error in the else
```

songz:
Screen Shot 2021-04-12 at 9 46 39 AM

components/DiscordBar.tsx
Ulisseus: You're right, I forgot to update interface again. I implemented type assertion as you suggested. 

As for different error messages, there are two ways fetching could fail. Actual network error 'error fetching data...' and response with incorrect schema (no members field). The latter gets generic error message. It's a pretty minor detail but different messages for different scenarios look a bit better.

Ulisseus:
'Join' button is less purple now. [Preview](https://c0d3-app-hp0h9jkhe-c0d3.vercel.app/curriculum). [Storybook](https://deploy-preview-645--youthful-thompson-448332.netlify.app/?path=/story/components-discord--twenty-users).

components/DiscordBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+            ? ' member online'
+            : ' members online'}
+        </div>
+      </div>
+      <div className="p-2">
+        <div className={`${styles['discord-list']}`}>{reshuffle(data)}</div>
+      </div>
+      <a
+        className={`${styles['discord_link']} font-weight-bold text-white`}
+        href="https://discord.com/invite/5KZ99KAs"
```
ggwadera, line 79: Oh I forgot about this, let's use our wonderful custom URL haha.

I'll approve the PR, just change this before merging please :)

```suggestion
        href="https://discord.gg/c0d3"
```

ggwadera (approved):

components/DiscordBar.tsx
songz: would have been better if this was a comment in your code... because when people read the code later, they may see your comment you left in this PR.

If you have to explain yourself for code reviewers, this explanation should be a comment in the code :)

Ulisseus: I added comment in this case but it's hard to strike balance between 'it should be obvious from the code' and 'too much comments'.

components/DiscordBar.tsx
Ulisseus: I updated URL.

songz (approved):

PR #646, opened 2021-04-10
[Migration] Change user `isAdmin` column type to boolean

ggwadera:
Part of #545

The time has finally come to fix this stupid thing! We had enough about this boolean being saved as a string! A real boolean should be `true` and not `'true'`!

Jokes aside, here's what this PR does:

- Added a migration to convert the `isAdmin` column to a boolean, also converting the values to the corresponding boolean ones, and set the column to `NOT NULL` too.
- Changed the Prisma schema and Sequelize model to reflect this change.
- Updated the `User` type-def for GraphQL too, and regenerated the `index.tsx` file.
- Fixed the type errors that came up after this.
- Fixed the admin/users page.
  - Filtering stopped working after the type change.
  - Made a kind of a ugly fix around this, the admin components definitively could use some refactoring.
  - Also simplified the filtering a little bit.
  - **The tests didn't catch that this page was broken**, gotta pay more attention to this.

### DIFF
--- a/__dummy__/sessionData.js
+++ b/__dummy__/sessionData.js
@@ -3,7 +3,7 @@ export default {
     id: 1,
     username: 'fakeusername',
     name: 'fake user',
-    isAdmin: 'true'
+    isAdmin: true
   },
   submissions: [],
   lessonStatus: []
--- a/__tests__/pages/admin/users.test.js
+++ b/__tests__/pages/admin/users.test.js
@@ -18,7 +18,7 @@ const dummyUsersData = [
     cliToken: 'WRyzWxDc_DqR3Avi7xcGD',
     email: 'some@mail.com',
     id: '5',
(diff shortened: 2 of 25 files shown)


ggwadera:
Some admin pages components are very broken after this, even with all tests passing. This isn't good :(

hwong0305 (approved):

hwong0305:
> Some admin pages components are very broken after this, even with all tests passing. This isn't good :(

You have to update the TypeScript types in the Admin page component like in `AdminLayout`

ggwadera:
Types are okay, what is broken is component behaviors like the users table. Filtering isn't working.

ggwadera:
That's why it's scary that the tests didn't catch anything. It would be broken on production if I haven't tested on my dev server.

kondanna (approved):

PR #647, opened 2021-04-11
update #622 - add server side redirection to index page

Ulisseus:
Currently we're using localStorage to check if user is authorized. It works but it's not ideal, user sees landing page for a second (Become a Software Engineer the Hard Way) then it's loading spinner of curriculum page. I'd like to find a way to redirect without loading index page first but nothing works (even adding a script to head).  

This PR changes the way landing page works. User sends request to `c0d3.com`, server checks if `connect.sid` cookie is present and sends either index or curriculum page back. No 'flickering'. But we're losing [Automatic Static Optimization](https://nextjs.org/docs/advanced-features/automatic-static-optimization) so it's some performance loss (it depends on number of vercel CNDs and servers close to you).


helpers/controllers/authController.ts
```diff
@@ -93,6 +92,10 @@ export const logout = async (_parent: void, _: void, ctx: Context) => {
           error: err.message
         })
       }
+      res.setHeader(
+        'Set-Cookie',
+        `connect.sid=false; expires=${new Date(Date.now())}; Path=/`
+      )
```
ggwadera, line 98: According to [this StackOverflow question](https://stackoverflow.com/q/5285940), the correct way to delete the cookie server-side is to set an expiration date in the past, like `Thu, 01 Jan 1970 00:00:00 GMT` to be 100% sure the cookie will be deleted. The cookie value can also be empty.

```suggestion
      res.setHeader(
        'Set-Cookie',
        `connect.sid=; expires=${new Date(0).toUTCString()}; Path=/`
      )
```

Also just FYI, `new Date()` already creates a date object with the current timestamp, so doing `new Date(Date.now())` is redundant ;)

pages/index.tsx
```diff
@@ -16,5 +12,17 @@ const IndexPage: React.FC<{}> = () => {
...
   )
 }
+export const getServerSideProps: GetServerSideProps = async context => {
+  return context.req.cookies['connect.sid']
+    ? {
+        redirect: {
+          destination: '/curriculum',
+          permanent: false
+        }
+      }
```
ggwadera, line 22: How about having another path for the landing page, like `/home` or `/landing`, and keep this `index` page only for redirection. This way we can keep the static page benefits while having the server-side redirection.

pages/index.tsx
```diff
@@ -16,5 +12,17 @@ const IndexPage: React.FC<{}> = () => {
     </>
   )
 }
+export const getServerSideProps: GetServerSideProps = async context => {
+  return context.req.cookies['connect.sid']
```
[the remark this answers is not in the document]
Ulisseus: Maybe I don't understand something but only graphql api route uses middleware now. Server side rendering doesn't have `req.user`.
Besides this cookie is created and destroyed on log in/log out. I doubt someone will edit it manually.

pages/index.tsx
Ulisseus: I could do it but I don't see the benefits. I'm not sure how exactly next handles server side redirects though. Does it automatically sends back redirected page or it sends back some redirect object first? 

Currently it's request to index->check for cookie->returning landing page. With redirect it would be request to index->check for cookie->returning redirect object->go to static landing page->returning landing page. Even if next automatically sends back curriculum page it would be about the same speed wise. Only people who would go straight to landing page would benefit from static optimization.

@types/helpers.d.ts
```diff
@@ -25,4 +26,5 @@ export interface UserSession extends Session {
 
 export interface Context {
   req: LoggedRequest
+  res: http.ServerResponse
```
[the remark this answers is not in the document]
Ulisseus: Done.

helpers/controllers/authController.ts
Ulisseus: Thanks. Fixed.

pages/index.tsx
ggwadera: I'm just concerned because we'll lose the static page generation benefits we have now, since using `getServerSideProps` forces the page to be server-side rendered. And making the landing page SSR seems wasteful because it is 100% static content (except for the profile query on the nav bar of course).

Unfortunately I don't know a good solution for this yet, but in my opinion we gotta think this through for some more time.

songz (approved):

hwong0305 (approved):

PR #648, opened 2021-04-11
update #622 - remove loading spinner from curriculum page

Ulisseus:
**What does this PR do?**
It removes loading spinner from curriculum page. But such simple change requires quite a lot of work.

Curriculum can finally use Next.js data fetching methods. Currently user needs to wait for getApp query which takes time. With these changes Next will add lessons and alerts data at build time. 

The bulk of work here is handling transitions between loading and loaded states. Previously we had unauthorized (lesson 0 is selected, progress is 0) and authorized states (your current lesson is selected, your current progress is shown), everything else was hidden behind loading spinner. I added new 'initial' state which doesn't have any selected lessons (and your progress is [loading](https://deploy-preview-648--youthful-thompson-448332.netlify.app/?path=/story/components-progresscard--loading)). So originally curriculum page is at 'initial' state, client sends new `session` query and rerenders the page after it gets the data. 

**I did not add any new logic in Curriculum page**. I just refactored it a bit because I need to run this code twice (initial render and rerender after page gets session data). TO-DO we need find a better way to store user data. So much effort for simply calculating current user lesson.

Go to [vercel](https://c0d3-app-11h15g138-c0d3.vercel.app/curriculum), disable javascript and reload the page. Data is already there, no need to wait for query. Enable js back and login as usual. You'll see initial->loaded transition. Refresh the page (transition is fast because it's loaded from persistent localStorage cache).


.storybook/webpack.config.js
```diff
@@ -16,5 +16,15 @@ module.exports = ({ config }) => {
   })
 
   config.resolve.extensions.push('.ts', '.tsx')
+  //exclude backend dependancies
+  config.externals = {
```
Ulisseus, line 20: This excludes backend node bundles from storybook.

graphql/queryResolvers/session.test.js
```diff
@@ -5,9 +5,9 @@ import db from '../../helpers/dbload'
 const { Submission, UserLesson, Star, User } = db
 
 describe('Session resolver', () => {
-  test('should return null if req.user does not exist', async () => {
+  test('should return empty session if req.user does not exist', async () => {
     const req = {}
-    expect(await session({}, {}, { req })).toEqual(null)
+    expect(await session({}, {}, { req })).toEqual({ lessonStatus: [] })
```
Ulisseus, line 10: Incorrect resolver was causing problems, so I fixed it. Session can't be null it's at least `{ lessonStatus: [] }` with our types.

helpers/apolloClient.ts
```diff
@@ -12,6 +12,25 @@ export const APOLLO_STATE_PROP_NAME = '__APOLLO_STATE__'
 let apolloClient: ApolloClient<NormalizedCacheObject>
 const cache = new InMemoryCache()
 
+import { persistCache } from 'apollo3-cache-persist'
+//restoring cache doesn't work with loading spinner, so I need to enable cache page by page
+const whiteList = ['/curriculum']
```
Ulisseus, line 17: Loading spinner doesn't work with persistent cache, hence white list. Eventually I'm going to remove loading spinner everywhere.

helpers/apolloClient.ts
```diff
@@ -12,6 +12,25 @@ export const APOLLO_STATE_PROP_NAME = '__APOLLO_STATE__'
...
+;(async function () {
```
Ulisseus, line 18: This strange self-invoking function is the only way to do it. There is no global await and I can't move this logic elsewhere because `MyApp` page can't be async (some Next.js restriction).

pages/login.tsx
```diff
@@ -83,7 +83,9 @@ const LoginPage: React.FC = () => {
   const router = useRouter()
   const [loginErrors, setLoginErrors] = useState<string[]>([])
   const [loginUser, { data, error }] = useMutation(LOGIN_USER, {
-    refetchQueries: [{ query: GET_APP }]
+    refetchQueries: [{ query: GET_APP }],
+    //prevents additonal render with unauthorized state on redirect
+    awaitRefetchQueries: true
```
Ulisseus, line 88: This is pretty important detail. By default refetching is async so without this flag you'll get additional render after redirect.

pages/curriculum.tsx
```diff
@@ -57,17 +55,58 @@ export const Curriculum: React.FC<{}> = () => {
...
+    className="mt-3 w-100"
+    src="https://discord.com/widget?id=828783458469675019&theme=light"
+    height="500"
+    allowTransparency
+    frameBorder="0"
+    sandbox="allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-scripts"
+  ></iframe>
+)
+export const Curriculum: React.FC<Props> = ({ lessons, alerts }) => {
+  const { data } = useGetSessionQuery({ fetchPolicy: 'cache-and-network' })
```
Ulisseus, line 86: I'm torn on adding this policy. It's not needed in 99% of cases but if user disabled localStorage refreshing the page will reset his status back to unauthorized. 
It won't cause any problems for normal users except small additional traffic.

ggwadera:
Login is not working on the preview.

Ulisseus:
Login works now (this PR was using old schema). [Preview](https://c0d3-app-11h15g138-c0d3.vercel.app/curriculum).

pages/curriculum.tsx
songz, line 86: this is a great comment, could you add this comment *in the code*? Because someone one day, another engineer will ask... "why is this necessary"?

songz (approved):

helpers/apolloClient.ts
songz: should be in code comment

pages/curriculum.tsx
```diff
@@ -111,5 +150,20 @@ export const Curriculum: React.FC<{}> = () => {
...
+export const getStaticProps: GetStaticProps = async () => {
+  const apolloClient = initializeApollo()
+  const query = await apolloClient.query<GetAppQuery>({
+    query: GetAppDocument
+  })
 
+  return {
+    props: {
+      lessons: query.data.lessons,
+      alerts: query.data.alerts
```
ggwadera, line 164: I think you can leave the alerts out of the SSG. They're not that common anyway. Might also cause some issues with alerts rendering on the page load making they show up and being dismissed after the page scripts run.

Ulisseus: Yes, I was thinking about it too. With this setup alerts will be updated every five minutes. Considering how often we use them it's not really a big deal. If it causes any problems with the rendering I'll include it in apollo query.

pages/curriculum.tsx
Ulisseus: I added a small comment explaining this policy.

Issue #677, opened 2021-04-15
Internal server error on landing page


Ulisseus:
It's very hard to reproduce but sometimes index page shows error screen. Refreshing always fixes it.

Vercel log:
`
GET] /
11:42:08:55
[...]

PR #680, opened 2021-04-15
update #677 - add separate aws config

Ulisseus:
Server side rendering uses AWS lambda which needs to be properly configured to work correctly. Lamda functions spawn/freeze/thaw containers according to their internal logic which in our case causes problems with db pooling. Most likely frozen containers don't release their connections properly.

I found a few articles about sequelize and aws. Both [recommend](https://github.com/sequelize/sequelize/blob/114c10eb09353cd32a630e42454a5d587ab0aed3/docs/manual/other-topics/aws-lambda.md) using lower connection count. 

**I can't properly test correctness of this PR**
I can't even reproduce landing page error on production server. Yes, it's there and if you refresh the page at random times you will get one eventually. But it depends on some internal AWS logic and I can't control it. Thankfully it shouldn't affect non-aws production config at all. New config is using `VERCEL_REGION` as a check for aws container. 
From vercel [docs](https://vercel.com/docs/environment-variables):
```
The ID of the Region where the app is running. Example: cdg1. Note: This Variable is only exposed during Runtime for Serverless Functions.
```
So current plan is to merge this PR, see if error is still there and revert it back (and probably revert back landing page redirect logic).

### DIFF
--- a/helpers/dbload.ts
+++ b/helpers/dbload.ts
@@ -7,6 +7,7 @@ import { ChallengeTypes, Challenge } from './models/Challenge'
 import { AlertTypes, Alert } from './models/Alert'
 import { StarTypes, Star } from './models/Star'
 
+/* istanbul ignore next */
 const sequelize = new Sequelize(
   process.env.DB_NAME as string,
   process.env.DB_USER as string,
@@ -17,10 +18,10 @@ const sequelize = new Sequelize(
     port: parseInt(process.env.DB_PORT as string),
     dialect: 'postgres',
     pool: {
-      max: 5,
+      max: process.env.VERCEL_REGION ? 1 : 5,
       min: 0,
-      acquire: 30000,
-      idle: 10000
+      acquire: process.env.VERCEL_REGION ? 3000 : 30000,
+      idle: process.env.VERCEL_REGION ? 0 : 10000
     }
   }
 )


helpers/dbload.ts
```diff
@@ -7,6 +7,7 @@ import { ChallengeTypes, Challenge } from './models/Challenge'
 import { AlertTypes, Alert } from './models/Alert'
 import { StarTypes, Star } from './models/Star'
 
+/* istanbul ignore next */
```
Ulisseus, line 10: I added jest ignore for this function because simple ternary operator doesn't warrant separate test IMO. And testing it will be rather ugly with a lot of mocks because Sequelize config isn't easily accessible.

ggwadera:
Just FYI, AWS means Amazon Web Services, but since we deploy on Vercel, we actually use Vercel lambdas (which is just another name for serverless functions). But if Vercel lambdas are equal to AWS I do not know the answer.

helpers/dbload.ts
```diff
@@ -17,10 +18,10 @@ const sequelize = new Sequelize(
     port: parseInt(process.env.DB_PORT as string),
     dialect: 'postgres',
     pool: {
-      max: 5,
+      max: process.env.VERCEL_REGION ? 1 : 5,
```
ggwadera, line 21: There's a more basic env var that vercel provides, its value will be 1 if it's running on a vercel server.

```suggestion
      max: process.env.VERCEL ? 1 : 5,
```

Ulisseus: I need to separate normal vercel environment from serverless. With `process.env.VERCEL` our production server will always use on connection which is bad and will cause problems.

ggwadera: What problems?

Ulisseus: Additional delay. If server gets two queries simultaneously one needs to wait.

Ulisseus:
>Re. Serverless vs AWS
I can't find it on vercel docs (there is mention of AWS API somewhere though) but here is [article](https://www.lambrospetrou.com/articles/battle-of-jamstack-platforms-netlify-vercel-aws/)
```
A big feature is once again their Serverless Functions offering, which is also using AWS Lambda under the covers. However, Vercel is a step up from Netlify, with more languages and more regions supported.
```

Ulisseus:
I just got Internal Server Error on preview. So this PR obviously doesn't work. Shame. 
landingError

PR #682, opened 2021-04-16
Refactor some things with Prisma

ggwadera:
- Replaced Sequelize with Prisma on the following files:
  - `helpers/middleware/user.ts`
  - `helpers/hasPassedLesson.ts`
  - `helpers/findUser.ts`
  - `helpers/controllers/passwordController.ts`
  - `helpers/controllers/authController.ts`
- Updated the `LoggedRequest` type to use the `User` type from Prisma
- Updated related tests
- Updated Prisma to the latest version


helpers/hasPassedLesson.ts
```diff
@@ -1,17 +1,15 @@
...
 export const hasPassedLesson = async (
   reviewerId: number,
   lessonId: string
 ): Promise<Boolean> => {
   // query userlesson that belongs to lesson and user
-  const userLesson = (await UserLesson.findOne({
+  const userLesson = await prisma.userLesson.findFirst({
     where: {
-      lessonId: lessonId,
+      lessonId: Number(lessonId),
```
Ulisseus, line 10: Are there plans to update grapql mutation as well? It's weird that Prisma already uses Int as type while resolver asks for string as argument.

helpers/controllers/passwordController.ts
```diff
@@ -25,21 +24,29 @@ export const reqPwReset = async (
...
+      where: {
+        id: user.id
+      },
+      data: {
+        forgotToken,
+        tokenExpiration
+      }
+    })
+
+    sendResetEmail(user.email!, user.forgotToken!)
```
Ulisseus, line 49: Why `user.email` is non-nullable in this case? `user` is just a record of `User` type and according to prisma schema email can be null.

ggwadera: Because we can be sure the user has an email if they're changing their password, since they receive an email for that. I didn't want to add useless checks just to keep TS happy.

I don't like this either, but the migration to make email and username not null is a bit harder. There's some entries with null or duplicated values in the database and I'm not sure how we should handle data loss in this case.

helpers/hasPassedLesson.ts
ggwadera: Yes, eventually. When I get to the submissions resolvers this will be fixed, since this is a helper function used there.

helpers/controllers/passwordController.ts
Ulisseus: Makes sense. I wonder how many users don't have email. Probably very few and they were created long ago, so we maybe we can simply delete them for inactivity.

Ulisseus (approved):

@types/helpers.d.ts
```diff
@@ -1,22 +1,15 @@
...
-  user: {
-    id: number
-    username: string
-    userLesson: UserLesson
-    email: string
-    name: string
-    isAdmin: boolean
-    cliToken: string
-  } | null
+  user: User | null
```
hwong0305, line 12: 👍

helpers/controllers/passwordController.ts
hwong0305: I think we should delete them if they're inactive and make `email` non-nullable in prisma schema. 

But we will need a way to confirm their last activity.

hwong0305 (approved):

PR #683, opened 2021-04-16
Fix package.json scripts not working on Windows

ggwadera:
Simple PR just to fix some scripts that fails if trying to develop on Windows.

Trying to run `db:init` or `db:seed` gives the following error.

image

Also included some docker helper commands so we don't have to write too much to get the db container started :P

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "scripts": {
     "db:init": "prisma migrate reset",
     "db:seed": "prisma db seed --preview-feature",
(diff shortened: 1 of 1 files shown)


Ulisseus:
Is it possible to add descriptions to every command? All these scripts are probably very confusing for new people. Something like short comment to wiki with full description.

Ulisseus (approved):

ggwadera:
Yeah we could do that, seems good.

PR #689, opened 2021-04-18
Update #499 - adds backend for showing star comments in user profile

kondanna:
- adds foreign key connecting Stars to Lessons
- updates foreign key types to Int and not null
- adds starGiven field to Star


Ulisseus:
What is this file `helpers/controllers/.userInfoController.ts.swp `?

prisma/schema.prisma
```diff
@@ -69,6 +70,7 @@ model Star {
   studentId Int?
   mentorId  Int?
   comment   String?  @db.VarChar(255)
+  lesson    Lesson?  @relation("starLesson", fields: [lessonId], references: [id])
```
ggwadera, line 73: Make this not optional. A star can't exist without a lesson.

```suggestion
  lesson    Lesson  @relation("starLesson", fields: [lessonId], references: [id])
```

prisma/migrations/20210404163725_added_star_lesson_relation/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AddForeignKey
+ALTER TABLE "stars" ADD FOREIGN KEY ("lessonId") REFERENCES "lessons"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```
ggwadera, line 2: Make the column not null, since every star needs a lesson. With that, deletion should also cascade instead of setting it to null.

Also give some stars on your dev db and then try applying the migration to see if doesn't cause any errors.

```suggestion
-- AddForeignKey
ALTER TABLE "stars" ADD FOREIGN KEY ("lessonId") REFERENCES "lessons"("id") ON DELETE CASCADE ON UPDATE CASCADE;

-- Make the column not nullable
ALTER TABLE "stars" ALTER COLUMN "lessonId" SET NOT NULL;
```

kondanna:
> What is this file `helpers/controllers/.userInfoController.ts.swp `?

vim buffer, will be deleted

helpers/controllers/userInfoController.ts
```diff
@@ -35,21 +35,60 @@ export const userInfo = async (_parent: void, args: UserInfoQueryVariables) => {
...
+        },
+        lesson: {
+          select: {
+            title: true,
+            order: true
+          }
+        }
+      }
+    }),
+    prisma.star.findMany({
```
ggwadera, line 54: I think this query to find the stars given by the user should remain separated from the profile query since this data is not used in the profile page.

What you could do is call this in the session resolver and make this query there where it's needed.

prisma/migrations/20210404163725_added_star_lesson_relation/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AddForeignKey
+ALTER TABLE "stars" ADD FOREIGN KEY ("lessonId") REFERENCES "lessons"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- Make the column not nullable
+ALTER TABLE "stars" ALTER COLUMN "lessonId" SET NOT NULL;
```
ggwadera, line 5: As I said on discord, there's some orphan star values that needs to be deleted before this migration.

Also make the `studentId` and `mentorId` not null. These FKs also need to have the ON DELETE action updated, but for that they need to be dropped and recreated, which means reindexing the entire table. But since the table is relatively small (only has 416 rows), this reindexing should not be any issue.

```suggestion
-- Clean orphan values
DELETE FROM "stars"
WHERE
	"lessonId" IS NULL
	OR "studentId" IS NULL
	OR "mentorId" IS NULL;

-- AddForeignKey
ALTER TABLE "stars" ADD FOREIGN KEY ("lessonId") REFERENCES "lessons"("id") ON DELETE CASCADE ON UPDATE CASCADE;

-- Make the columns not nullable
ALTER TABLE "stars" ALTER COLUMN "lessonId" SET NOT NULL;
ALTER TABLE "stars" ALTER COLUMN "studentId" SET NOT NULL;
ALTER TABLE "stars" ALTER COLUMN "mentorId" SET NOT NULL;

-- Drop and recreate student and mentor FK to update the ON DELETE action
ALTER TABLE "stars" DROP CONSTRAINT "stars_mentorId_fkey";
ALTER TABLE "stars" DROP CONSTRAINT "stars_studentId_fkey";

ALTER TABLE "stars" ADD CONSTRAINT "stars_mentorId_fkey" 
FOREIGN KEY ("mentorId") REFERENCES "users"("id") 
ON DELETE CASCADE
ON UPDATE CASCADE;

ALTER TABLE "stars" ADD CONSTRAINT "stars_studentId_fkey" 
FOREIGN KEY ("studentId") REFERENCES "users"("id") 
ON DELETE CASCADE
ON UPDATE CASCADE;
```

helpers/controllers/userInfoController.ts
```diff
@@ -35,21 +35,60 @@ export const userInfo = async (_parent: void, args: UserInfoQueryVariables) => {
...
+  }, {} as lessonMentorMapType)
+
   const starMap = stars.reduce((map: StarMap, star) => {
-    // TODO change star table so lessonId is not null
-    // the null check is needed so typescript doesn't complain
-    if (!star.lessonId) return map
-    map[star.lessonId] = map[star.lessonId] || []
-    map[star.lessonId].push(star)
+    map[star.lessonId!] = map[star.lessonId!] || []
+    map[star.lessonId!].push(star)
```
ggwadera, line 85: Remove the not null assertions since they will not be needed anymore.

```suggestion
    map[star.lessonId] = map[star.lessonId] || []
    map[star.lessonId].push(star)
```

helpers/controllers/userInfoController.ts
```diff
@@ -35,21 +35,60 @@ export const userInfo = async (_parent: void, args: UserInfoQueryVariables) => {
...
     })
   ])
 
+  interface lessonMentorMapType {
+    [lessonId: string]: string
+  }
+
+  const lessonMentorMap = starsGiven.reduce((map, starGiven) => {
+    const mentorUsername = _.get(starGiven, 'mentor.username', '')
+    map[starGiven.lessonId!] = mentorUsername
```
ggwadera, line 79:
```suggestion
    map[starGiven.lessonId] = mentorUsername
```

helpers/controllers/userInfoController.ts
```diff
@@ -35,21 +35,60 @@ export const userInfo = async (_parent: void, args: UserInfoQueryVariables) => {
...
+            username: true
+          }
+        }
       }
     })
   ])
 
+  interface lessonMentorMapType {
+    [lessonId: string]: string
+  }
```
ggwadera, line 75: Move interfaces and types outside the function scope.

prisma/schema.prisma
```diff
@@ -63,12 +64,13 @@ model Lesson {
...
-  lessonId  Int?
+  lessonId  Int
   createdAt DateTime @default(now()) @db.Timestamptz(6)
   updatedAt DateTime @updatedAt @db.Timestamptz(6)
   studentId Int?
   mentorId  Int?
   comment   String?  @db.VarChar(255)
+  lesson    Lesson   @relation("starLesson", fields: [lessonId], references: [id])
   mentor    User?    @relation("starMentor", fields: [mentorId], references: [id])
   student   User?    @relation("starStudent", fields: [studentId], references: [id])
```
ggwadera, line 75: As per the suggested migration changes, these fields will also be not null.

```suggestion
  studentId Int
  mentorId  Int
  comment   String?  @db.VarChar(255)
  lesson    Lesson   @relation("starLesson", fields: [lessonId], references: [id])
  mentor    User     @relation("starMentor", fields: [mentorId], references: [id])
  student   User     @relation("starStudent", fields: [studentId], references: [id])
```

prisma/migrations/20210404163725_added_star_lesson_relation/migration.sql
ggwadera: Hmm perhaps it's better to split these other things into another PR. I'll leave it up here for now, if you think it's good for this PR then commit it.

helpers/controllers/userInfoController.ts
kondanna: This definitely belongs on the session resolver. I was thinking since session resolver is just a userInfo resolver with a username passed in, could just put this here and then session resolver just calls userInfo. Would that be an antipattern?

ggwadera: Yeah I get you, but the way you did it will make this additional query even for the profile resolver, which does nothing with this data. The way I suggested would be to call this in the session resolver, and then just add this star given query there where it's needed. Something like this for example:

```js
const session = async (req) => {
  const { username } = req.user
  const [session, starsGiven] = await Promise.all([
    userInfo(username),
    prisma.star.findMany({ ... }) // get the stars given by this user
  ])
  // do the additional mapping
  return {
    user,
    lessonStatus,
    submissions
  }
}
```

kondanna: ah, yeah that's much better

graphql/typeDefs.ts
```diff
@@ -168,10 +168,12 @@ export default gql`
...
 
   type Star {
-    id: String!
-    studentId: Int
-    mentorId: Int
-    lessonId: Int
+    id: Int!
+    studentId: Int!
+    mentorId: Int!
+    lessonId: Int!
```
ggwadera, line 174: Are these ID values used on the app? Seems kinda unnecessary to me to have ids together with the nested value.

kondanna: just lessonId . removing the other two.

songz (approved):

PR #690, opened 2021-04-18
Updates #499 - Frontend for displaying star comments in user profile page

kondanna:
Included:
Stories for ProfileSubmissions and ProfileStarComments
added tests to [username].test.js 
adds dummyStarData.js to __dummy__ files

backend files from other PR


components/ProfileStarComments.tsx
```diff
@@ -0,0 +1,76 @@
...
+              {'@' + username}
+            </div>
+            <h6 className={`${styles['comment-lesson-title']} text-left`}>
+              {lessonTitle}
+            </h6>
+          </div>
+        </div>
+        <StarIcon
+          className={`${styles['comment-star-icon']}`}
+          strokeWidth={'1'}
```
Ulisseus, line 43: No need for curly brackets in this case. It's a string.

Ulisseus:
Many rules from `profileStarComments.module.scss` could be changed to bootsrap class names. 
` display: flex;`->`d-flex`
`align-items: center;`->`align-items-center`
`justify-content: center;`->`justify-content-center`
Also padding and margins.

scss/profileStarComments.module.scss
```diff
@@ -0,0 +1,65 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap');
```
Ulisseus, line 1: We should self-host it probably. See PTMono under public folder for example.

components/ProfileStarComments.tsx
```diff
@@ -0,0 +1,76 @@
...
+            className={`${styles['user-info-image']} text-uppercase bg-primary rounded-circle text-light`}
+          >
+            {firstName[0]}
+            {lastName[0]}
+          </div>
+          <div>
+            <div className={`${styles['comment-username']} text-left`}>
+              {'@' + username}
+            </div>
+            <h6 className={`${styles['comment-lesson-title']} text-left`}>
```
Ulisseus, line 36: It's a very minor thing since c0d3 is not really optimized for accessibility but skipping heading ranks should be avoided if possible. [See](https://www.w3.org/WAI/tutorials/page-structure/headings/)

scss/profileStarComments.module.scss
kondanna: Import removed.

components/ProfileStarComments.tsx
kondanna: I changed everything to h6 or p

songz (approved):

Issue #422, opened 2020-10-07
Change Progress Card status/description on completion


Cijin:
Right now, the description for progress card does not change even on completion of challenges.

![Progress Card](https://user-images.githubusercontent.com/1990966/95319126-b3039980-08c1-11eb-8720-08eff51eb97a.png)

PR #694, opened 2021-04-19
Updates #422 - Modifies progresscard based on modules completed

pmulard:
Also adds test coverage for progresscard.

### DIFF
--- a/components/ProgressCard.test.js
+++ b/components/ProgressCard.test.js
@@ -1,0 +1,22 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import ProgressCard from './ProgressCard'
+
+describe('ProgressCard component', () => {
+  it('should render correct lesson progress for 0% completion', () => {
+    const { container } = render(<ProgressCard progressCount={0} />)
+    expect(container).toMatchSnapshot()
+  })
+  it('should render correct lesson progress for 25% completion', () => {
+    const { container } = render(<ProgressCard progressCount={25} />)
+    expect(container).toMatchSnapshot()
+  })
+  it('should render correct lesson progress for 75% completion', () => {
+    const { container } = render(<ProgressCard progressCount={75} />)
+    expect(container).toMatchSnapshot()
+  })
+  it('should render correct lesson progress for 100% completion', () => {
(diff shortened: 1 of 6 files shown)


ggwadera (approved):
Nice work 👍

components/ProgressCard.tsx
```diff
@@ -57,14 +57,28 @@ const ProgressSVG: React.FC<Props> = ({ progressCount }) => {
 }
 
 const ProgressCard: React.FC<Props> = ({ progressCount }) => {
+  const setHeaderFromProgress = () => {
+    if (progressCount <= 13) {
+      // Has not finished 1st module yet
+      return `You're off to a great start!`
+    } else if (progressCount <= 50) {
```
songz, line 64: FYI, if you are returning, `else` is redundant.

songz (approved):

hwong0305 (approved):

components/ProgressCard.test.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import ProgressCard from './ProgressCard'
+
+describe('ProgressCard component', () => {
+  it('should render correct lesson progress for 0% completion', () => {
+    const { container } = render(<ProgressCard progressCount={0} />)
+    expect(container).toMatchSnapshot()
```
Ulisseus, line 8: No need to check against new snapshot every time. Testing library has more advanced [queries](https://testing-library.com/docs/queries/about/). For example, you can check if `You're off to a great start!` is present on the screen:

` expect(screen.getByText("You're off to a great start!")).toBeTruthy()`

Or if your component changes its state over time. 
` await waitFor(() => screen.getByText('You're off to a great start!'))`

It's just for your information. Your PR is fine as it is.

Ulisseus (approved):

components/ProgressCard.test.js
pmulard: Thank you for explaining this. Still very fresh using Jest.

components/ProgressCard.tsx
pmulard: Yes, good catch.

PR #713, opened 2021-04-21
Updates session.ts to use prisma

kondanna:
also makes use of userInfo queries


ggwadera (approved):

ggwadera:
You need to fix the tests to mock the prisma calls

PR #714, opened 2021-04-21
 update #555 - add Foundations of Javascript

Ulisseus:
* First self-hosted lesson
* Css fixes for mdx components
* A few (very simple) new mdx components

[Vercel](https://c0d3-app-2f7x6kygy-c0d3.vercel.app/curriculum/lessons/js0/primitive_data_and_operators).


ggwadera:
There is some formatting errors caused by backticks. [image](https://user-images.githubusercontent.com/16023489/115613990-7c242c80-a2c3-11eb-960c-b7a01e2d7231.png)

Also, what if we name the pages like `/lessons/js0` instead of `/curriculum/lessons/foundations`?

Ulisseus:
>There is some formatting errors caused by backticks. 

There are probably many typos as well. I copied text from notion (thankfully simple ctrl+c ctrl+v keeps markdown syntax) but every spoiler needs to be updated manually. We should proofread every lesson before migrating from notion. 

>Or even split by sections so we have like /lessons/js0/onesection and /lessons/js0/anothersection, this way the pages doesn't get too much lengthy too.

I was thinking about it too. Our self-hosting setup won't have any lazy loading so lesson pages could become really heavy. I also remember that working through full lesson was rather tiresome just because it's so big. Ideally it should be  - read section one, solve a few challenges, read section two, solve a few challenges and etc.

components/mdx/Layout.tsx
```diff
@@ -1,11 +1,22 @@
 import React from 'react'
```
hwong0305, line 1: I think the file should be named something else since there's already a file called Layout.

scss/index.scss
```diff
@@ -38,10 +38,39 @@ body {
 }
 
 p {
-  font-size: 15px;
+  font-size: 16px;
```
hwong0305, line 41: How does this impact other pages? Do we really want this across the app? This change isn't scoped.

Ulisseus: The better question is why it was needed in the first place? If you compare notion and self-hosted lesson you can notice that the latter looks a bit better. I suppose notion spent quite some time optimizing for the best reading experience, so I tried to copy it as much as I can. `16px` is the default font size anyways.

components/mdx/Layout.tsx
Ulisseus: Renamed.

Ulisseus:
There are a bunch of type errors at master right now, so there is no point in trying to fix vercel (and reviewing this pr).

__mocks__/next/router.js
```diff
@@ -1,14 +1,15 @@
...
     route: '/',
     pathname: '/',
     query,
     asPath: '/',
-    push: push,
+    push,
     replace: async () => true,
     reload: () => null,
-    back: () => null
+    back
```
Ulisseus, line 13: This way you write less boilercode if you need to check parameters with which `push` was called (simple object destruction instead of adding spies every time). Same as with `push`.

components/Layout.tsx
```diff
@@ -5,7 +5,7 @@ import Footer from './Footer'
 const Layout: React.FC = ({ children }) => (
   <>
     <AppNav />
-    <div className="container">{children}</div>
+    <div className="container-md">{children}</div>
```
Ulisseus, line 8: This change cascaded all these snapshot updates. It's for mobile layout, smaller screens with less than `768px` wide get more space.

Ulisseus:
Updated [vercel](https://c0d3-app-2f7x6kygy-c0d3.vercel.app/). Lesson 0 [Part1](https://c0d3-app-2f7x6kygy-c0d3.vercel.app/curriculum/lessons/js0/primitive_data_and_operators), [Part2](https://c0d3-app-2f7x6kygy-c0d3.vercel.app/curriculum/lessons/js0/functions_and_execution_context) (there are no links to these lessons from anywhere outside yet). 

* Added autogenerated links to lesson parts under main title
* There is `scroll to the top` arrow now
* Added `Go Back` button for non-lesson documents
* Small css fixes (LightBulb and TwoColumn components transform into column-flex on smaller screens)

This PR got rather big so unless there is something really broken I'd prefer to fix it in later PRs. Students won't see these lessons yet anyway.

ggwadera:
Mobile TOC looks kinda broken.

scss/index.scss
ggwadera: Shouldn't we be using `em` units instead of absolute `px` units anyways to keep size scaling pixel perfect?

Ulisseus:
>Mobile TOC looks kinda broken.

What browser? And what are your phone dimensions? Mozilla and chromium both looks fine for me. 

Screen Shot 2021-04-25 at 17 57 56

scss/index.scss
Ulisseus: `px` are not really pixels, they are some sort of pseudo units which are tied to physical screen size, so it's not that bad. For example, hero image on the landing page uses hardcoded width and height in pixels (just like most next/images) and it will look good every screen from small old phones to modern 4k monitors. 

I wouldn't use `em`s for fonts because they depend on parent component, so `em` in on place won't be the same as `em` somewhere else. 

`rem` is tied to root font size and it's the safest way to define font-size. It will scale if user zooms the page and it will be the same everywhere. 

At least that how I understand it, don't quote me on that.

ggwadera:
Chrome on my phone

Ulisseus:
[Vercel](https://c0d3-app-mlsd1cm32-c0d3.vercel.app/curriculum/lessons/js0/primitive_data_and_operators). Does this work for you? It resizes correctly in my Chrome.

ggwadera:
Yeah it's working now 👍

[Screenshot_20210424-152432978.jpg](https://user-images.githubusercontent.com/16023489/115969035-542a0880-a511-11eb-8538-64b046fa3bf6.jpg)

ggwadera (approved):

Ulisseus:
I refactored `Spoiler` component with `details` tag:
* Spoiler arrow doesn't change its position on click.
* No need to ignore `.mdx` in eslint config. 
* It's also much more simpler compared to previous iteration. No need for js or custom css anymore. I could even remove `Spoiler` altogether, but providing default name is more convenient because majority of spoilers use 'Answer' title. 

New [vercel](https://c0d3-app-178sgmqfp-c0d3.vercel.app/curriculum/lessons/js0/primitive_data_and_operators) (not much of a difference here).

PR #716, opened 2021-04-22
Add submission status DB constraint and GraphQL enum

ggwadera:
Part of #545

- Added a migration to create a constraint for the `status` column in the `submissions` table
  - Now `status` is not null and can only be `passed`, `needMoreWork` or `open`, defaulting to the last one.
  - The migration will also delete some legacy entries from the db which have a `underReview` status which is not used in the app. There's 13 rows that match this condition and will be deleted. [These are the entries.](https://user-images.githubusercontent.com/16023489/115643766-cc1ae780-a2f3-11eb-9651-dea082b0914f.png)

- Added a `SubmissionStatus` GraphQL enum type in our typedef file.
  - Changed the type of the `status` field in the `Submission` type to this enum.
  - The auto-generated `graphql/index.tsx` also contains this enum type that can be used for development.
  - Also updated the `submissions` resolver to use Prisma.
- Replaced the hardcoded submission status strings with the new enum on the query resolvers and React components.
  - The `ProfileSubmissions` component had a unused `pending` status code switch case, I replaced it with the `open` status.
  - Also updated all tests, snapshots and stories.
- Fixed the `codegen.yml` file to generate the HOCs that we still use, and to also run `eslint` after generating.
- Added more seed data to include some open submissions for the first lesson from the `noob` user.

### DIFF
--- a/codegen.yml
+++ b/codegen.yml
@@ -13,4 +13,4 @@ generates:
       withHOC: true
 hooks:
   afterAllFileWrite:
-    - prettier --write
+    - eslint --fix
--- a/codegen.yml
+++ b/codegen.yml
@@ -10,3 +10,7 @@ generates:
       - "typescript-react-apollo"
     config:
       withHooks: true
+      withHOC: true
+hooks:
+  afterAllFileWrite:
+    - prettier --write
--- a/components/LessonCard.tsx
+++ b/components/LessonCard.tsx
@@ -10,6 +10,7 @@ import {
 import NavLink from './NavLink'
 import Image from 'next/image'
 import styles from '../scss/lessonCard.module.scss'
+import { SubmissionStatus } from '../graphql'
 
 const ReviewCount: React.FC<ReviewCountProps> = props => {
   const { loading, data } = useQuery(GET_SUBMISSIONS, {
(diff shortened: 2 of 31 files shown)


kondanna (approved):

codegen.yml
```diff
@@ -9,4 +9,8 @@ generates:
       - "typescript-resolvers"
       - "typescript-react-apollo"
     config:
-      withHooks: true
\ No newline at end of file
+      withHooks: true
+      withHOC: true
```
Ulisseus, line 13: Why are we still using apollo HOCs? I remember right I needed to add this line every time I recreated graphql types.

ggwadera: Some pages still use the `withGetApp` HOC instead of `useGetAppQuery` hook.

components/LessonCard.tsx
```diff
@@ -10,6 +10,7 @@ import {
 import NavLink from './NavLink'
 import Image from 'next/image'
 import styles from '../scss/lessonCard.module.scss'
+import { SubmissionStatus } from '../graphql'
 
 const ReviewCount: React.FC<ReviewCountProps> = props => {
   const { loading, data } = useQuery(GET_SUBMISSIONS, {
```
Ulisseus, line 16: Should be changed to codegen generated `useSubmissionsQuery`. Or at least it should be properly typed, currently data returned from this query in untyped. 

It's not related to prisma migration but still.

ggwadera: Yeah it's unrelated. Don't wanna touch any more stuff as this PR is already big enough.

Ulisseus (approved):

Issue #733, opened 2021-04-23
Progress Card still redirects to Mattermost


hwong0305:
![image](https://user-images.githubusercontent.com/7990856/115938435-807d5080-a44f-11eb-9f0b-729222057a44.png)

hwong0305:
Additionally, if you're not logged in, it should not show as 0%. Ask them to join.

PR #734, opened 2021-04-25
Update progress card

JasirZaeem:
For #733 

# Changes made

## components/ProgressCard.tsx

- Change mattermost link to a discord link
- add Different prop types for svg progress and progress card
- Accept a logged in prop
- Show progress only if a user is logged in
- removed loading spinner from progress svg, as svg progress is not rendered until a user is in state

## components/ProgressCard.test.tsx

- Added a new test to check ProgressCard is no user is logged in
- Added loggedin prop to existing tests

## pages/curriculum.js

- Pass logged in prop true to ProgressCard is a user is in state

## stories/components/ProgressCard.stories.tsx

- Added logged in component to progress card comopnent
- Added story for logged out progress card

## Updated snapshots

### DIFF
--- a/components/ProgressCard.tsx
+++ b/components/ProgressCard.tsx
@@ -83,17 +83,18 @@ const ProgressCard: React.FC<CardProps> = ({ progressCount, loggedIn }) => {
           </>
         ) : (
           <h4 className={`${styles['progress-card__title']} mt-3`}>
-            Join and start learning!
+            Join C0D3 now and start your software engineering journey!
           </h4>
         )}
 
         <div className="mt-3">
+          Join us on
           <NavLink
             className={`${styles['progress-newuser__chatlink']} text-white`}
             path="https://discord.gg/c0d3"
             external
           >
-            Join us on Discord
+            Discord
           </NavLink>
           to ask your questions.
           <p>
--- a/components/ProgressCard.tsx
+++ b/components/ProgressCard.tsx
@@ -1,11 +1,16 @@
 import React from 'react'
 import NavLink from './NavLink'
 import styles from '../scss/progressCard.module.scss'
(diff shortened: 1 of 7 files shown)


components/ProgressCard.tsx
```diff
@@ -1,93 +1,99 @@
...
+            Join and start learning!
+          </h4>
+        )}
+
         <div className="mt-3">
-          Login to
           <NavLink
             className={`${styles['progress-newuser__chatlink']} text-white`}
-            path="https://chat.c0d3.com/login"
+            path="https://discord.gg/c0d3"
```
Ulisseus, line 93: Maybe you should update css for styling discord link. Previously link to mattermost was surrounded by text from each side so additional padding worked but right now it looks off. Or simply add link to `Discord` word instead of `Join us on Discord`. 

margins

JasirZaeem: Thanks, I hadn't noticed it. I'll reduce the link to be just on `Discord` since there is a Discord card above the progress card anyway

Ulisseus:
One thing you could improve upon is transition from `we don't know anything about the user` to authorized or unauthorized state. If you go to vercel and refresh (maybe with throttling and disabled cache) you'll notice that sidebar 'jumps` breaking the layout. It's because `ProgressBar` component has different height depending on its content. Ideally it should be the same height while text and progress percentages appear inside it. 

But I'm totally nitpicking, your PR is great and it could be merged as it is.

JasirZaeem:
Since we did not find a satisfactory way to ensure the same card height across different contents I've increased the text of the title to minimise the amount of shifting that happens on content change. And have fixed the discord link.

ggwadera (approved):

PR #736, opened 2021-04-26
close #735  - add usernames to reviews

### LINKED ISSUE #735: No usernames in review comments
You can see the comment but you don't know who wrote it. It's rather confusing. 
reviewComment

Ulisseus:
**Changes**:

* `getSubmission` query and `submissionController` include reviewer data
* `ReviewCard` has simple username next to comment

Updated [ReviewCard](https://deploy-preview-736--youthful-thompson-448332.netlify.app/?path=/story/components-reviewcard--with-long-comment).


ggwadera:
Splitting the comment section leaves a very weird whitespace imo ([image](https://user-images.githubusercontent.com/16023489/116116177-3e435180-a691-11eb-8d15-e37af70f2a0d.png)). That's a bunch of unused space the comment could use.

I think the reviewer name/username can be just a line on the top before the comment. Don't even need flex for that.
[Something like this looks better to me](https://user-images.githubusercontent.com/16023489/116116589-b14cc800-a691-11eb-83e1-85662c433e4b.png). The `@username` should also link to the reviewer profile.

pages/review/[lesson].tsx
```diff
@@ -19,14 +19,14 @@ import _ from 'lodash'
...
-  submissions: SubmissionData[]
+  submissions: Submission[]
 }
 
 const SubmissionDisplay: React.FC<SubmissionDisplayProps> = ({
   submissions
 }) => (
   <div className="submissions-container container p-0">
-    {submissions.map((submission: SubmissionData) => (
+    {submissions.map((submission: Submission) => (
```
Comment (ggwadera) [inline pages/review/[lesson].tsx:29]: Don't need the type here, it gets inferred from the array type.

pages/review/[lesson].tsx
```diff
@@ -60,9 +60,9 @@ const Review: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
     router.push('/curriculum')
     return <LoadingSpinner />
   }
-  const lessonSubmissions: SubmissionData[] = data
+  const lessonSubmissions: Submission[] = data
```
Comment (ggwadera) [inline pages/review/[lesson].tsx:63]: This type should also get inferred from the query type.

Also the `Submission` type might not always be the same as the one from the query, as the query may not include all the fields in the type. For this there's the autogenerated query result type: `SubmissionsQuery['submissions']`

pages/review/[lesson].tsx
```diff
@@ -60,9 +60,9 @@ const Review: React.FC<QueryDataProps<AppData>> = ({ queryData }) => {
...
     ? data.submissions.filter(
-        (submission: SubmissionData) =>
+        (submission: Submission) =>
```
Comment (ggwadera) [inline pages/review/[lesson].tsx:65]: Again unnecessary, inferred type

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx]: Removed.

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx:63]: I tried to type incoming data but it became much less readable because you need either check for every optional field or assert non-nulls. We should fix our type definitions first. 

And I'm not quite sure if you need to type every query result like this `SubmissionsQuery['submissions']`. Codegen generates nullable fields like this `<Maybe T> | undefined` so you can safely use general `Submission` type for everything.

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx:65]: It's `any` without this type.

Ulisseus:
I changed layout for [ReviewCard](https://deploy-preview-736--youthful-thompson-448332.netlify.app/?path=/story/components-reviewcard--active-card), simple inline name indeed looks better than flex. It's also links to reviewer profile.

kondanna (approved):

pages/review/[lesson].tsx
Comment (ggwadera) [reply pages/review/[lesson].tsx:63]: That's not the issue that may arise. The generated types for the query still comes from the ones on the typedef, so the optional fields still gonna be there. That's only gonna get fixed by correcting the typedef.

Let me give you an example of an issue that this may cause. Imagine if now you needed the reviewer email from a submission for some reason, to do something in one of the components for the page, like to put a button to email the reviewer. Since the component uses the [`Submission`](https://github.com/garageScript/c0d3-app/blob/1dab2746ba0409f3f87b21930babec3894547398/graphql/typeDefs.ts#L94) type, which has a `reviewer` field of type [`User`](https://github.com/garageScript/c0d3-app/blob/1dab2746ba0409f3f87b21930babec3894547398/graphql/typeDefs.ts#L119), all the fields from `User` are available for you from Typescript. So you just get the email from there and that's it, it should work as expected. So you commit, open the PR and then it get's merged.

Then you push it to production and it doesn't work. You check the logs and there's a bunch of undefined value errors. But why though? You coded the new feature and all the tests and type checks were ok. What happened is that you didn't check the GraphQL query, and didn't see that the reviewer field only includes the `id` and `username`. Even though Typescript was telling you that everything was ok, in reality that wasn't true. Unfortunately this is a weakness of duck-typed languages like TS.

Of course this doesn't cause an immediate problem for this PR. What I'm trying to emphasize is that it's a code smell, that may lead someone not familiar with the code in the future to create a new bug.

Now, since the `submissions` query return type is still optional, until that's fixed you need to do a workaround to use the generated query type. Something like this:

```ts
type submissionData = NonNullable<SubmissionsQuery['submissions']>[number]
```

In the case of the code that this comment is referencing, the type of `data` is `any` because the component is not using the generated `useSubmissionsQuery` hook. If you use the correct typed hook, the types can be inferred very easily and typing will be much easier.

pages/review/[lesson].tsx
Comment (ggwadera) [reply pages/review/[lesson].tsx:65]: Type is `any` because component doesn't use generate query hook. A little personal tip from me, always be cautious of variables with an `any` type, 99% of the time it needs to be fixed. Inferred types can take you long way without needing explicit assertions.

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx:63]: I agree with the importance of proper typing but I still can't see how using general codegen types could cause any issues. Codegen is smart it doesn't use general types, it picks only fields which are defined in query. So with your reviewer email example, this is SubmissionQuery type:
```js
export type SubmissionsQuery = { __typename?: 'Query' } & {
  submissions?: Maybe<
    Array<
      Maybe<
        { __typename?: 'Submission' } & Pick<
          Submission,
          | 'id'
          | 'status'
          | 'diff'
          | 'comment'
          | 'challengeId'
          | 'lessonId'
          | 'createdAt'
          | 'updatedAt'
        > & {
            challenge?: Maybe<
              { __typename?: 'Challenge' } & Pick<Challenge, 'title'>
            >
            user?: Maybe<
              { __typename?: 'User' } & Pick<User, 'id' | 'username'>
            >
            reviewer?: Maybe<
              { __typename?: 'User' } & Pick<User, 'id' | 'username' | 'name'>
            >
          }
      >
    >
  >
}
```
You can't just type `submission.reviewer.email`, typescript will complain (a lot). First `submission` itself could be undefined or null, second `reviewer` can be undefined or null and according to this type `reviewer` doesn't even has `email` field. While general `User` type has it SubmissionQuery type is generated according to graphql query so it's `Pick<GeneralType, 'certain fields'>`.

pages/review/[lesson].tsx
Comment (Ulisseus) [reply pages/review/[lesson].tsx:65]: The problem is codegen types are not strict enough. Almost every field is optional so it's either a lot of unnecessary null/undefined checks or type assertion again.

Ulisseus:
I fixed conflicts but I won't change `submission` types in this PR, it has nothing to do with adding profiles to reviews. 
 
To properly refactor types we need to start with incorrect typedefs. Remove `withQuery` wrapper and use single query instead of GET_APP and GET_SUBMISSIONS. Maybe even add SSG for this page and enable persistent cache.

pages/review/[lesson].tsx
Comment (ggwadera) [reply pages/review/[lesson].tsx]: > SubmissionQuery type is more specific

Exactly, it's only a subset of the fields from `Submission` that are included on the query. The example I gave you was assuming you were using the more general `Submission` type.

As I said before, the optional fields are still gonna be the same, as they are inherited from the typedef. So null checks won't be affected by this.

pages/review/[lesson].tsx
Comment (ggwadera) [reply pages/review/[lesson].tsx]: > it's either a lot of unnecessary null/undefined checks or type assertion again

Yes that's the reason to use typescript, so you don't assume everything is there and get into an undefined value error on production. It's better to do useless null/undefined checks with a safely typed value than having to use an `any` type that you don't know what it is and doesn't work as expected.

components/ReviewCard.tsx
```diff
@@ -74,16 +76,38 @@ export const DiffView: React.FC<DiffViewProps> = ({ diff = '' }) => {
...
+    <a
+      className={`${styles['comment_author']} mt-2 d-block`}
+      href={`/profile/${username}`}
+    >
+      <div className="d-inline">{`${firstName} ${
+        lastName ? lastName : ''
+      }`}</div>
+      <div className="d-inline text-muted">
+        {username ? '@' + username : 'Anonymous user'}
+      </div>
```
ggwadera, line 97: Why the check on `lastName`? You already default it to an empty string.

Also put a space before the username otherwise it renders right next to the name.

```suggestion
      <div className="d-inline">{`${firstName} ${lastName}`}</div>
      {' '}
      <div className="d-inline text-muted">
        {username ? '@' + username : 'Anonymous user'}
      </div>
```

components/ReviewCard.tsx
```diff
@@ -74,16 +76,38 @@ export const DiffView: React.FC<DiffViewProps> = ({ diff = '' }) => {
 }
 
 const MemoDiffView = memo(DiffView)
-
+type UserProfileProps = {
+  username: string | undefined | null
+  name: string | undefined | null
+}
+const UserProfile: React.FC<UserProfileProps> = ({ username, name }) => {
```
ggwadera, line 83: Rename this to something else to make it clear that it renders the reviewer user.

`UserProfile` makes me think that it's something from the profile page.

Ulisseus: Renamed to `ReviewerProfile`.

components/ReviewCard.tsx
Ulisseus: I need the second check to `lastName`, otherwise it can end up undefined.
` const lastName = name ? name.split(' ')[1] : ''` defaults it to empty string is there is no `name`.
```lastName ? `${lastName} ` : ''``` defaults it to empty string if `name.split(' ')[1]` is undefined

ggwadera: You can simplify the default assignement then.

`const lastName = name?.split(' ')[1] || ''`

or 

`const lastName = name?.split(' ')[1] ?? ''`

Ulisseus: Done.

ggwadera (approved):

PR #740, opened 2021-04-27
Replace `connect-session-sequelize` with `prisma-session-store`

ggwadera:
- Removed `connect-session-sequelize`.
- Added [`@quixo3/prisma-session-store`](https://github.com/kleydon/prisma-session-store).
- Added migration to recreate the `Session` table to use with new session store.
  - Updated `schema.prisma` with new Session model.
  - All users will be logged out after this is merged.
- Replaced sequelize with prisma session store on `pages/api/graphql.ts` middleware.
- Fixed Sequelize altering tables.
- Added `.nvmrc` file to set the Node version for Netlify. (Build was failing because it used Node 10)


songz (approved):

Ulisseus (approved):

helpers/dbload.ts
```diff
@@ -96,7 +97,8 @@ UserLesson.belongsTo(User, {
   as: 'user'
 })
 
-sequelize.sync({ alter: !!process.env.ALTER_DB })
+// Calling this method causes Sequelize to alter the DB, undoing changes from Prisma migrations
```
Ulisseus, line 100: So this will prevent sequelize from altering DB according to its models in the future.  Should we reapplyl prisma migrations again to fix current state of production db?

ggwadera: Not necessarily all the migrations, from what I've seen it's just some foreign keys constraints that we had changed, to set `ON DELETE CASCADE` for example, that sequelize messed up.

I need to look at the production db to see what has been changed. I think writing a new migration is easier too, since prisma already keeps track of applied migrations it might make things too complicated to re-apply.

PR #748, opened 2021-04-28
update#545 updates User type on prisma and graphql

kondanna:
Graphql - User typeDef now requires username, email, and name

Prisma - User table now requires username, email, name, password, with username and email unique constraints added.

Only 'real' change besides that is the addition of password in authController (line 153)

    let newUser = await prisma.user.create({
      data: {
        name,
        username,
        email,
        password <-- now required by schema
      }
    })

### DIFF
--- a/prisma/migrations/20210428060853_user_username_email_required_unique/migration.sql
+++ b/prisma/migrations/20210428060853_user_username_email_required_unique/migration.sql
@@ -1,0 +1,22 @@
+/*
+  Warnings:
+
+  - A unique constraint covering the columns `[username]` on the table `users` will be added. If there are existing duplicate values, this will fail.
+  - A unique constraint covering the columns `[email]` on the table `users` will be added. If there are existing duplicate values, this will fail.
+  - Made the column `name` on table `users` required. This step will fail if there are existing NULL values in that column.
+  - Made the column `username` on table `users` required. This step will fail if there are existing NULL values in that column.
+  - Made the column `password` on table `users` required. This step will fail if there are existing NULL values in that column.
+  - Made the column `email` on table `users` required. This step will fail if there are existing NULL values in that column.
+
+*/
+-- AlterTable
+ALTER TABLE "users" ALTER COLUMN "name" SET NOT NULL,
+ALTER COLUMN "username" SET NOT NULL,
+ALTER COLUMN "password" SET NOT NULL,
+ALTER COLUMN "email" SET NOT NULL;
+
+-- CreateIndex
+CREATE UNIQUE INDEX "users.username_unique" ON "users"("username");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "users.email_unique" ON "users"("email");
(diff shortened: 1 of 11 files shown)


prisma/migrations/20210428060853_user_username_email_required_unique/migration.sql
```diff
@@ -0,0 +1,22 @@
...
+  - Made the column `name` on table `users` required. This step will fail if there are existing NULL values in that column.
+  - Made the column `username` on table `users` required. This step will fail if there are existing NULL values in that column.
+  - Made the column `password` on table `users` required. This step will fail if there are existing NULL values in that column.
+  - Made the column `email` on table `users` required. This step will fail if there are existing NULL values in that column.
+
+*/
+-- AlterTable
+ALTER TABLE "users" ALTER COLUMN "name" SET NOT NULL,
+ALTER COLUMN "username" SET NOT NULL,
+ALTER COLUMN "password" SET NOT NULL,
```
ggwadera, line 15: This migration will probably fail. Currently password is set only after the user has signed up, through the email you receive. Unless the sign up process is changed to include the password, this column should be kept optional.

ggwadera: Oh another thing, there's a bunch of users with duplicated emails and usernames, so the `unique` constraint will fail if we try to apply it. See the links I will send on discord for the sql queries I used to check.

prisma/migrations/20210428060853_user_username_email_required_unique/migration.sql
```diff
@@ -0,0 +1,20 @@
...
+-- AlterTable
+ALTER TABLE "users" ALTER COLUMN "name" SET NOT NULL,
+ALTER COLUMN "username" SET NOT NULL,
+ALTER COLUMN "email" SET NOT NULL;
+
+-- CreateIndex
+CREATE UNIQUE INDEX "users.username_unique" ON "users"("username");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "users.email_unique" ON "users"("email");
```
ggwadera, line 20: I think it's better if we create these unique indexes on a later migration until we figure out a way to do it safely.

For now, let's just make the columns not null to solve the typescript complaints.

```suggestion
```

prisma/schema.prisma
```diff
@@ -114,10 +114,10 @@ model UserLesson {
 
 model User {
   id                     Int          @id @default(autoincrement())
-  name                   String?      @db.VarChar(255)
-  username               String?      @db.VarChar(255)
+  name                   String       @db.VarChar(255)
+  username               String       @db.VarChar(255) @unique
```
ggwadera, line 118: See previous comment.

```suggestion
  username               String       @db.VarChar(255)
```

prisma/schema.prisma
```diff
@@ -114,10 +114,10 @@ model UserLesson {
...
   password               String?      @db.VarChar(255)
-  email                  String?      @db.VarChar(255)
+  email                  String       @db.VarChar(255) @unique
```
ggwadera, line 120: See previous comment.

```suggestion
  email                  String       @db.VarChar(255)
```

ggwadera (approved):

PR #750, opened 2021-04-28
fix image stretching

Ulisseus:
#744 fixed test warnings but it broke image stretching (check 500 error screen or compare J4 and J5 lesson images on curriculum page). This PR fixes mismatch between `Image` and `img` props.

### DIFF
--- a/__mocks__/next/image.tsx
+++ b/__mocks__/next/image.tsx
@@ -1,6 +1,11 @@
 import React from 'react'
 import type { ImageProps } from 'next/image'
+import _ from 'lodash'
 
-const Image = (props: ImageProps) => <img {...props} />
+const Image = (props: ImageProps) => {
+  //not all Image properties are allowed in <img/> tag
+  const copy = _.omit(props, 'objectFit')
+  return <img {...copy} />
+}
 
 export default Image
--- a/__mocks__/next/image.tsx
+++ b/__mocks__/next/image.tsx
@@ -3,7 +3,9 @@ import type { ImageProps } from 'next/image'
 import _ from 'lodash'
 
 const Image = (props: ImageProps) => {
-  //not all Image properties are allowed in <img/> tag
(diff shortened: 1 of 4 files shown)


__mocks__/next/image.tsx
```diff
@@ -1,6 +1,11 @@
 import React from 'react'
 import type { ImageProps } from 'next/image'
+import _ from 'lodash'
 
-const Image = (props: ImageProps) => <img {...props} />
+const Image = (props: ImageProps) => {
+  //not all Image properties are allowed in <img/> tag
```
songz, line 6: could you add comment on why `objectFit` is important here so we don't accidentally remove it?

Example:

```
// ObjectFit causes warning, but we need to add it because curriculum page images won't fill up its allocated space and looks weird without it.
```

This way we won't regress anymore

__mocks__/next/image.tsx
```diff
@@ -1,6 +1,11 @@
...
+  const copy = _.omit(props, 'objectFit')
```
ggwadera, line 7: Why not simply delete that property? Making a copy and using lodash seems kinda overkill.

```suggestion
  delete props.objectFit
```

Or if the copy is really needed you can do it with vanilla JS.

```suggestion
  const { objectFit, ...copy } = props
```

Ulisseus: `delete` doesn't work, maybe it's inherited from the parent or protected/frozen. Simple spread is probably more readable I agree.

__mocks__/next/image.tsx
Ulisseus: I updated comment. Just to clarify `objectFit` doesn't cause any errors in production, it's a completely valid [property](https://nextjs.org/docs/api-reference/next/image#objectfit) for `next/Image`.

ggwadera (approved):

PR #757, opened 2021-04-29
Fix Nav bar on Review Page

### LINKED ISSUE #756: Non-Functional nav button in mobile review page LessonTitleCard
The `Show Challenges` button does nothing on the Mobile version of the review page.

image

**Proposed Fix**: Replace `Show Challenges` tab with `Challenge` navlink on the review page regardless of device width

JasirZaeem:
Fixes #756 

## Changes

- ### components/LessonTitleCard.tsx
  - Show `CHALLENGE` NavLink instead of `SHOW CHALLENGE` if setShow is not passed regardless of screen width
- ### components/LessonTitleCard.test.js
  - Separate tests to test rendering LessonTitleCard on Curriculum page and Review page
- ### Update Snapshots


Ulisseus:
Clicking on `CHALLENGES` from `https://c0d3-app-git-fork-jasirzaeem-fix-mobile-nav-review-page-c0d3.vercel.app/review/2` redirects you to `/curriculum/2` while ideally it should open challenges modal.

JasirZaeem:
> Clicking on `CHALLENGES` from [mobile review page](https://c0d3-app-git-fork-jasirzaeem-fix-mobile-nav-review-page-c0d3.vercel.app/review/2) redirects you to `/curriculum/2` while ideally it should open challenges modal.

That would require fetching challenges on the review page, then adding the challenge selection modal to review page too. After that, we can't link to curriculum page since that router only goes upto the lesson, can't link to a specific challenge since that is decided by that component. Alternative would be then to render challenge cards on the review page too somewhere. Do we want to do that?
I have replicated the functionality from desktop version to mobile version.
- Before:
  - Desktop
    - Review page links to curriculum
    - Curriculum page shows challenge
  - Mobile
    - Review page's link does not work
    - Curriculum page open challenge selection modal
- Now:
  - Desktop: Same as before
  - Mobile
    - Review page links to curriculum
    - Curriculum page open challenge selection modal

ggwadera:
I think it should be only a link to the challenges page. Good point about not being able to link directly to a challenge too, so the modal doesn't make too much sense on the review page.

Ulisseus:
Then we need to change `Go Back` button to actually redirect you back to previous page and not to `/curriculum`. Otherwise it's very confusing.

ggwadera:
> Then we need to change `Go Back` button to actually redirect you back to previous page and not to `/curriculum`. Otherwise it's very confusing.

lol why does it do that? hate when website have a back button linked to a specific page.

the back button should just do `router.pop()` everywhere in the app

Ulisseus:
>lol why does it do that? hate when websites have a back button linked to a specific page.

It's not really a button, it's a `NavLink` component which requires hard coded route to redirect. I guess that's why.

JasirZaeem:
I've made go back, go back in history with router

components/LessonTitleCard.tsx
```diff
@@ -24,7 +27,18 @@ const LessonTitleCard: React.FC<LessonTitleProps> = props => {
           />
           <div>
             <p className="m-0">
-              <NavLink path="/curriculum">Go Back</NavLink>
+              <a
+                href=""
```
ggwadera, line 31: Use `href="#"` so it doesn't actually link to anything.

```suggestion
                href="#"
```

components/LessonTitleCard.test.js
```diff
@@ -25,3 +51,23 @@ describe('LessonTitleCard component', () => {
...
+describe('LessonTitleCard component on Review Page', () => {
+  const props = {
+    lessonCoverUrl: 'coverUrl',
+    lessonUrl: 'lessonUrl',
+    lessonTitle: 'Test Lesson',
+    lessonId: '0',
+    isPassed: true
+  }
+
+  test('Should render default layout for all screens', async () => {
```
Ulisseus, line 64: Unfortunately, snapshot testing fails short when it comes for media query based rules. You can check `LessonTitleCard.test.js.snap` file:
`LessonTitleCard component on Review Page Should render default layout for all screens 2` and `LessonTitleCard component on Review Page Should render default layout for all screens 2` snapshots are the same. 
Jest compares only classnames and they don't change.

JasirZaeem: I haven't changed this, this is how tests were before. They are just testing whether `show challenges` is shown on mobile or not which snapshot can detect. I made two describe blocks for review and curriculum page, tests that differ for the two pages can be put in these. But for now, on curriculum page `Show challenge` should be shown and on review page no change should be made on mobile (it is handled by bootstrap's responsive css)
So since this describe is checking that the markup remains unchanged regardless of window width, the snapshots should be same right ?

components/LessonTitleCard.test.js
```diff
@@ -1,7 +1,33 @@
...
 import { render, screen, fireEvent, waitFor } from '@testing-library/react'
-describe('LessonTitleCard component', () => {
+
+const mockBack = jest.fn()
+
+jest.mock('next/router', () => ({
+  useRouter: () => ({
+    back: mockBack
+  })
+}))
```
ggwadera, line 11: We already have a manual mock for the router, don't need to mock it again.

https://github.com/garageScript/c0d3-app/blob/master/__mocks__/next/router.js

You can just use the back method directly and jest will mock it automatically.

components/LessonTitleCard.test.js
```diff
@@ -25,3 +51,23 @@ describe('LessonTitleCard component', () => {
...
+  }
+
+  test('Should render default layout for all screens', async () => {
+    global.window.innerWidth = 1080
+    let { container } = render(<LessonTitleCard {...props} />)
+    expect(container).toMatchSnapshot()
+
+    global.window.innerWidth = 500
+    ;({ container } = render(<LessonTitleCard {...props} />))
+    expect(container).toMatchSnapshot()
```
ggwadera, line 71: Split this into two tests for wide and mobile layout. Tests should be testing a single thing only.

components/LessonTitleCard.test.js
Ulisseus: You're right, layouts should be the same in this case. But in general it's better to avoid relying on snapshots when you need to test different layouts.

components/LessonTitleCard.test.js
JasirZaeem: Removed custom mock, now using the existing named mock

components/LessonTitleCard.test.js
JasirZaeem: Since markup does not change in this case and there is no specific functionality to test as it is static component, I've reduced the duplicate snapshot test and made it just check that the component gets rendered

ggwadera (approved):

PR #767, opened 2021-05-01
Refactored submissions resolver with Prisma, removed Sequelize

### LINKED ISSUE #766: GraphQL API returning 500 status even though it's working
The API is not retuning the sucess status correctly, even though people are saying that the process was completed, it still shows an error message.

CLI submission example:

Review page bug report:

image

ggwadera:
Part of #545
Closes #766

- Submissions table migration
  - Deletes legacy orphan data with null `userId`, `lessonId` or `challengeId`
  - Sets `userId`, `lessonId` and `challengeId` to NOT NULL
  - Creates unique index on (`userId`, `lessonId`, `challengeId`)
  - Updates the foreign keys to add `ON DELETE CASCADE`

- Refactored `submissionsController` with Prisma
  - Simplified the queries by using a single Prisma query to get the lesson and challenge along with the submission
  - Updated some types
  - Refactored the `getReviewer` function to use coalescing operator, removing lodash
  - Updated and simpliflied many of the tests

- Refactored `updateSubmission` with Prisma
  - Simplified some of the logic
  - Optimized queries
  - Updated and simpliflied many of the tests

- Updated submissions GraphQL queries
  - Added the `createSubmission` the cli uses to have proper return types for it
  - Added `lessonId` argument to `acceptSubmission` and `acceptSubmission` as it was already being sent and is used on the resolver.

- Fixed some leftover TO-DOs that were waiting for some types to be fixed.

- **REMOVED SEQUELIZE FROM THE PROJECT** (F* FINALLY 🚀🚀🚀🚀🚀🚀)

I tested submissions on my end with the CLI and reviews, everything was working correctly, no error messages.

P.S.: the submissions queries might not work on the preview as it may rely on the migration being run to update the table.


helpers/controllers/submissionController.ts
```diff
@@ -1,55 +1,60 @@
...
+          challengeId
+        }
+      },
+      create: { ...submissionData, challengeId, lessonId, userId: Number(id) },
+      update: submissionData,
+      include: {
+        user: true,
+        lesson: true,
+        challenge: true
+      }
```
kondanna, line 45: wow.

helpers/updateSubmission.ts
```diff
@@ -1,120 +1,125 @@
...
+    })
 
     const [
-      { id: userChatId, username: chatUsername },
-      lessonSubmissions
+      lessonChallengeCount,
+      passedLessonSubmissions,
+      userLesson,
+      { id: userChatId, username: chatUsername }
     ] = await Promise.all([
```
Ulisseus, line 54: Do we still need `Promise.all`? According to wiki it was a weird [quirk](https://github.com/garageScript/c0d3-app/wiki/Sequelize-Query-Performance) of sequelize.

ggwadera: Dunno, what's bad about it?

I don't think it's really a "quirk", it's more of a way to start multiple processes in parallel.

Ulisseus: With sequelize `Promise.all` was much faster than one big query. 
We should delete this wiki page once sequelize is refactored.

Ulisseus (approved):

PR #775, opened 2021-05-04
update #555 - add Learn page

Ulisseus:
This PR:

* adds new self-hosted page  - Learn (previously Journey). I've omitted `Training Camp` and everything below, lesson list is available on curriculum page anyways while git videos are mentioned in some early lesson. I need to find a place for `Number bases page` though. Note: AppNav link still redirects to notion.  
* changes development setup link from notion to self-hosted page. I think it's good enough at this point.
* updates `ModalImage` component to handle very tall or very long images (custom css based on manually passed props). If you open current [development setup page](https://www.c0d3.com/docs/setup) on mobile device shell image is too small even if you click on it, this pr fixes it. 

[Vercel](https://c0d3-app-o25ihmp1c-c0d3.vercel.app/). Check development setup button, set responsive layout, click on shell image, you will see bigger scrollable pop up.

[Learn page](https://c0d3-app-o25ihmp1c-c0d3.vercel.app/docs/learn).


pages/curriculum/lessons/js0/functions_and_execution_context.mdx
```diff
@@ -65,6 +65,7 @@ Let's break down the above example.
     src="/assets/curriculum/lessons/foundations/params.png"
     width="1000"
     height="400"
+    long='true'
```
Ulisseus, line 68: In case you're wondering why not simply `long`, react prints warnings about boolean passed as props. I'm not sure why. 
```
Warning: Received `true` for a non-boolean attribute `tall`.                       
                                                                                   
If you want to write it to the DOM, pass a string instead: tall="true" or tall={val
ue.toString()}.     
```

components/mdx/ModalImage.tsx
```diff
@@ -2,7 +2,9 @@ import React, { useState } from 'react'
 import Modal from 'react-bootstrap/Modal'
 import Image, { ImageProps } from 'next/image'
 
-const ModalImage: React.FC<ImageProps> = (props: ImageProps) => {
+const ModalImage: React.FC<
+  ImageProps & { long?: string | boolean; tall?: string | boolean }
```
ggwadera, line 6: Why `string | boolean`? With only boolean you don't need to pass a value to it like you needed.

Ulisseus: That React warning is not related to typescript, it comes from passing boolean as an attribute to element. Maybe I'm blind but I can't figure out why it's happening.

Ulisseus:
I changed `ModalImage` types a bit, there is one general `proportion` type which can be either `long` or `tall`. `proportions='tall'` looks less awkward that `tall='true'` and etc.

ggwadera (approved):

PR #778, opened 2021-05-07
update #555 - add Arrays (JS2)

Ulisseus:
Another self-hosted lesson.

I removed `Owner` paragraph (sorry, noob101) and meme pictures for consistency because other lessons don't have them.

Since there is a lot of info about `innerText`, `innerHTML` we should add html challenges too. Optional exercises at the end of first section could moved to challenges. Also, we should add new testing challenges to JS2 and every lesson afterwards.  

[Vercel](https://c0d3-app-7e1pdkyzq-c0d3.vercel.app/curriculum/lessons/js2/introduction_to_testing_inner_properties)


ggwadera:
Not directly related to this, but I just noticed that the code highlighting theme has a little issue, some symbols are getting a white background making it look a little weird on top of that sepia background.

image

Ulisseus:
>a white background making it look a little weird

It's be design (at least notion does this too). It looks kind of weird indeed, so maybe I should disable it.

Ulisseus:
I disabled token highlighting. It looks like this now:

token

ggwadera (approved):

PR #779, opened 2021-05-07
close #508 - add reminder to enable javascript 

### LINKED ISSUE #508: c0d3 doesn't work without javascript

Ulisseus:
Making `C0D3` fully usable without javascript won't be possible unless we drop certain Next.js features like using router instead of `window.location` and etc. 

So I chose to add a small remainder and close this issue. 

[Vercel](https://c0d3-app-4lekum2qd-c0d3.vercel.app/). Disable javascript and you will get a bunny with a sign urging you to enable it back.

### DIFF
--- a/__tests__/pages/__snapshots__/_app.test.js.snap
+++ b/__tests__/pages/__snapshots__/_app.test.js.snap
@@ -2,6 +2,7 @@
 
 exports[`MyApp component should render Login component passed in as prop 1`] = `
 <div>
+  <noscript />
   <nav
     class="navbar navbar-expand-lg navbar-light justify-content-between bg-white"
   >
--- a/__tests__/pages/__snapshots__/_app.test.js.snap
+++ b/__tests__/pages/__snapshots__/_app.test.js.snap
@@ -2,7 +2,6 @@
 
 exports[`MyApp component should render Login component passed in as prop 1`] = `
 <div>
-  <noscript />
   <nav
     class="navbar navbar-expand-lg navbar-light justify-content-between bg-white"
   >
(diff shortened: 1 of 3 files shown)


ggwadera (approved):
By the way I think you were trying to say "reminder" (def.: One who or that which reminds; anything which serves to awaken remembrance.), instead of remainder (def.: Something left over after other parts have been taken away.).

Ulisseus:
On the other hand, maybe it's apollo HOC wrapper was causing problems instead of next router. `C0D3` is kind of half-usable without js right now. Landing won't load (it waits for local storage), login/login won't work, curriculum works but clicking on any lesson results in eternal loading spinner. Self-hosted dev setup page looks good, so any self-hosted lessons should be usable with javascript disabled too. 

I need to change a few things (I still would like to add this bunny though). 

>By the way I think you were trying to say "reminder" 

Thanks. It was autocorrected and I'm too lazy to proofread anything.

ggwadera:
Every page that is statically generated can be used without JS, the landing page only displays the nav bar because the page content is not being statically rendered due to the local storage check which is done client-side with JS.

Ulisseus:
>due to the local storage check which is done client-side with JS

I was thinking about moving the whole `LandingPage` component inside `noscript` tag but it wasn't designed for such things. First react renders it twice with javascript enabled, second it breaks styling (even in prod build). And to make challenge page work without js we need to enable static generation, challenges data should be included at build time.

Anyways, I moved bunny to `index` page and centered it a bit. It looks like this now. 
bunny

PR #789, opened 2021-05-11
update #583 - add Comment model and graphql queries

Ulisseus:
This is for github-like line comments. 

Line comment is identified by submission id, file name (in case of multiply files in one submission) and line number. Multiply comments from the same conversation are sorted according to creation date. 

To add comment you need all of the  above plus content of new comment and id of the author.

To get all comments for certain file you need submission id and file name.


graphql/queryResolvers/getComments.test.js
```diff
@@ -0,0 +1,12 @@
+import { prisma } from '../../prisma'
+import { getComments } from './getComments'
+
+//should be added to coverageIngorePatters
```
Ulisseus, line 4: There are many tests like this under resolvers folder and this is a bad practice IMO. These test will never fail and we shouldn't retest third party modules. 

I've created two such pseudo-tests to pass jest coverage requirements but I'd rather add all resolvers to ignore patterns.

prisma/schema.prisma
```diff
@@ -136,6 +149,7 @@ model User {
   submissionsReviewed    Submission[] @relation("userReviewedSubmissions")
   submissions            Submission[] @relation("userSubmissions")
   userLessons            UserLesson[]
+  comments               Comment[]
```
ggwadera, line 152: You might have forgotten this here. There's no `UserLesson <--> Comment` relation on the model.

graphql/typeDefs.ts
```diff
@@ -110,6 +118,18 @@ export default gql`
...
+  type Comment {
+    id: Int!
+    fileName: String!
+    line: Int!
+    content: String!
+    authorId: Int!
+    submissionId: Int!
+    createdAt: String!
+    author: User
+    submission: Submission
```
ggwadera, line 130: Shouldn't these two be not optional too?

graphql/queries/getComments.ts
```diff
@@ -0,0 +1,18 @@
+import { gql } from '@apollo/client'
+
+const GET_COMMENTS = gql`
+  query getComments($submissionId: Int!, $fileName: String!) {
```
ggwadera, line 4: Don't know how you're thinking of doing this, but are you planning on doing a query for each file on each submission?

Why not include the comments in the submissions query to have prisma join the tables, so you get everything in a single query?

graphql/queries/addComment.ts
```diff
@@ -0,0 +1,23 @@
+import { gql } from '@apollo/client'
+
+const ADD_COMMENT = gql`
+  mutation addComment(
+    $line: Int!
+    $submissionId: Int!
+    $content: String!
+    $fileName: String!
+    $authorId: Int!
```
ggwadera, line 9: `authorId` shouldn't be a parameter, you need to get it from the session in the resolver from the user that sent the request. Otherwise someone could impersonate another person when sending a comment.

graphql/queryResolvers/addComment.ts
```diff
@@ -0,0 +1,21 @@
+import { prisma } from '../../prisma'
+import { MutationAddCommentArgs } from '../../graphql'
+import { Context } from '../../@types/helpers'
+
+export const addComment = async (
+  _parent: void,
+  arg: MutationAddCommentArgs,
+  _ctx: Context
+) => {
+  const { line, submissionId, authorId, fileName, content } = arg
```
ggwadera, line 10: Get `authorId` from `req.session.user`, don't send it as a parameter for the query.

ggwadera (changes requested):

prisma/schema.prisma
Ulisseus: There is one to many User-Comment relation. Line 152 is from `User` model and not from `UserLesson`.

graphql/queries/addComment.ts
Ulisseus: Makes sense. I will change it.

graphql/queries/getComments.ts
Ulisseus: I need to recheck it but almost every question "why does it look so weird?' can be answered by 'because of the way our diff rendering library works'. 

Getting all comments in one submission query should be better indeed.

prisma/schema.prisma
ggwadera: Oh right, sorry.

graphql/queries/getComments.ts
Ulisseus: Come to think about it I'm not sure if getting everything in one query would result in better user experience. 

Case in point: when you're reviewing submission you start with one big GET_APP query which is hidden behind loading spinner. And there is GET_SUBMISSION query which gets all submissions for the given lesson (it's using spinner too). The bigger these queries are the longer end user stares at loading screen. Meanwhile, querying for comments case by case is non-blocking. Git diff renders as it is, when comment query gets back you will get a comment box. 

More importantly, to make comments interactive I'm using polling. So there won't be one big query vs many small queries. It's repeating the same query every second, in this case the one with simple select wins over joins.

graphql/typeDefs.ts
Ulisseus: You're right I was thinking that it's about `schema.prisma` and not graphql `typedefs`. I made these fields non nullable as well.

graphql/queryResolvers/addComment.ts
Ulisseus: Fixed.

graphql/queries/getComments.ts
ggwadera: Biggest issue with that would be content shift as you load the comments, what if you had a slow connection and open a submission with a lot of comments? You'd had a page with content jumping around all over the place.

In my opinion, keep this simple for now, query everything together and don't do polling, update the comments by having the user refresh the page. You can tackle these issues separately **IF** needed in the future. Premature optimization only leads to more complexity.

Ulisseus: I agree with keeping it simple but in this case simple is most likely means using separate queries. I was trying to make top down passing of comments work but ChallengeMaterial needs to be refactored before anything could be done. It's a mess.

ggwadera: Yeah, just change the [`GET_SUBMISSIONS`](https://github.com/garageScript/c0d3-app/blob/master/graphql/queries/getSubmissions.ts) query to include the comments list instead of the single comment field that it has now.

Ulisseus: OK, everything is coming top down from GetApp and GetSubmissions queries. I'm still not sure if I won't need to revert it back to fetching it submission by submission in the future.

ggwadera: You need to change [`GET_SUBMISSIONS`](https://github.com/garageScript/c0d3-app/blob/master/graphql/queries/getSubmissions.ts) to include comments too, `GET_APP` only fetches the logged in user submissions. `GET_SUBMISSIONS` fetches the submissions awaiting review.

You already changed the resolver, only the query that needs to be fixed.

helpers/controllers/submissionController.ts
```diff
@@ -103,25 +103,25 @@ export const rejectSubmission = async (
 export const submissions = async (
   _parent: void,
   arg: QuerySubmissionsArgs,
-  ctx: Context
+  _ctx: Context
 ) => {
-  try {
-    const { lessonId } = arg
-    await getReviewer(ctx.req.user, lessonId)
```
Ulisseus, line 110: Submissions can be viewed by reviewer (in this case they are coming from GetSubmissions query which checks if user has passed this lessons). And they can be viewed by student in this case they are coming from GetApp query which could be run by anyone. 

To update cache I need to refetch both of these. So I removed this check. It shouldn't cause problems, because review page has another check while accept/reject mutations still use getReviewer.

graphql/queries/getComments.ts
Ulisseus: Yes, I forgot to update the query. Fixed.

ggwadera (commented):
What about existing comments, will they still work?

helpers/controllers/submissionController.ts
Ulisseus: Update. I found a way to make it work without removing `getReviewer` check. So student will refetch only GET_APP query while reviewer will refetch only GET_SUBMISSIONS query.

Ulisseus:
>What about existing comments, will they still work?

These changes should not touch existing comments in any way. Rejection comment is a field in Submission model while new inline comments are completely new model. I rejected submission locally and resubmitted it, it works as expected.

ggwadera (approved):

Issue #585, opened 2021-03-26
Feature (from #483) - give students ability to comment on review


kondanna:
Maybe each submission can have multiple comments, from one or more reviewers and the student. More like a conversation.

ggwadera:
Another suggestion is to be able to comment on individual lines similar to Github PR reviews.

One issue that may arise for this one is that the submissions table only has a single comment field for each entry. To be able to have multiple comments we'll probably need to change some things on the database and have another table for the comments with a many-to-one relation.

Also the mock-ups on Zeplin display a very nice review screen, we could take them as inspiration.

![28  Review Challenges - View Challenge](https://user-images.githubusercontent.com/16023489/112635131-338b6780-8e1a-11eb-8746-93326a0cf260.png)
![29  Review Challenges - View Accepted Challenge](https://user-images.githubusercontent.com/16023489/112635136-34bc9480-8e1a-11eb-94c2-10a96317af8e.png)

Ulisseus:
I tried to implement github-like comments but after a few unsuccessful attempts I switched to other things.
 
If I remember correctly we're using two libraries - one for parsing submission string and another for highlighting differences. 

One approach is to store all lines in a state. When you click on a line it inserts new component discussion component. Cons. You need to implement highlighting by yourself and it's not an easy task at all. Otherwise you need to modify highlighting library so it can include state. Again not easy. 

I agree that we should improve review form. Maybe even integrate it into github somehow.

kondanna:
> Another suggestion is to be able to comment on individual lines similar to Github PR reviews.
> 
> One issue that may arise for this one is that the submissions table only has a single comment field for each entry. To be able to have multiple comments we'll probably need to change some things on the database and have another table for the comments with a many-to-one relation.
> 
> Also the mock-ups on Zeplin display a very nice review screen, we could take them as inspiration.
> 
> ![28 Review Challenges - View Challenge](https://user-images.githubusercontent.com/16023489/112635131-338b6780-8e1a-11eb-8746-93326a0cf260.png)
> ![29 Review Challenges - View Accepted Challenge](https://user-images.githubusercontent.com/16023489/112635136-34bc9480-8e1a-11eb-94c2-10a96317af8e.png)

Perhaps add a Review_Comments table that's a many-to-one for each submission? 
And drop the comment column from submissions table?

Initial thought was to just implement the flat comments structure seen in the Zeplin mockups but I'll look into this GH-style commenting (with syntax highlighting?!)

[...]

PR #814, opened 2021-05-21
update #585 - add DiffView component

Ulisseus:
The last new component for inline comments. 

DiffView will be used to render diff for student and reviewer both. If you check current `ReviewCard` and `ChallengMaterial` components you will find some duplicated logic. 

This components gets an array of comments along with other props. Since submission can consists of multiply files it creates state indexed by "id:fileName" ("1:1.js"). Every file has an array of line numbers and array of comments. Every time syntaxHyghlight runs it checks if current line number is included in the state and renders CommentBox if true. If you click on lineNumber you add it to the state.

Note: DiffView will render only new lines, just like current review page. So instead of green/red/white for new lines/deleted lines/unaffected lines everything will be green. It sounds like a big deal but it is not. Currently only the first three lessons are using full diff highlighting and in a very boring way (against the initial template file). 

This PR should not affect production in any way. 
[
Storybook](https://deploy-preview-814--youthful-thompson-448332.netlify.app/?path=/story/components-diffview--default) (real component will clear input after adding comment, it's a storybook bug).

### DIFF
--- a/components/DiffView.tsx
+++ b/components/DiffView.tsx
@@ -1,12 +1,29 @@
 import React, { useEffect, memo } from 'react'
 import gitDiffParser, { File } from 'gitdiff-parser'
 import Prism from 'prismjs'
-import ReactDiffViewer from 'c0d3-diff'
+import ReactDiffViewer, { ReactDiffViewerStylesOverride } from 'c0d3-diff'
 import _ from 'lodash'
 import CommentBox from './CommentBox'
 import { Comment } from '../graphql'
 
 const prismLanguages = ['js', 'javascript', 'html', 'css', 'json', 'jsx']
+
+const styles: ReactDiffViewerStylesOverride = {
+  lineNumber: {
+    fontWeight: 'bold',
+    color: '#364d3b'
+  },
+  gutter: {
+    background: '#cdffd8',
+    '&:hover': {
+      fontSize: '1.2rem',
+      opacity: '0.75'
+    }
+  },
+  emptyGutter: {
+    pointerEvents: 'none'
+  }
+}
 
 const DiffView: React.FC<{
   diff?: string
@@ -23,24 +40,25 @@ const DiffView: React.FC<{
   const [commentsState, setCommentsState] = React.useState<fileComments>({})
 
   useEffect(() => {
-    const commentMap: fileComments = {}
-    comments &&
-      comments.forEach(c => {
-        if (!commentMap[`${c.submissionId}:${c.fileName}`])
-          commentMap[`${c.submissionId}:${c.fileName}`] = {
+    const commentsMap: fileComments =
+      comments &&
+      comments.reduce((acc: fileComments, comment) => {
+        const index = `${comment.submissionId}:${comment.fileName}`
+        if (!acc[index])
+          acc[index] = {
             lines: [],
             comments: []
           }
-        commentMap[`${c.submissionId}:${c.fileName}`].lines!.push(c.line)
-        commentMap[`${c.submissionId}:${c.fileName}`].comments!.push(c)
-      })
-    setCommentsState(commentMap)
+        acc[index].lines?.push(comment.line)
+        acc[index].comments?.push(comment)
+        return acc
+      }, {})
+    setCommentsState(commentsMap)
     //rerunning useEffect on id rerenders submission when student clicks on another challenge
   }, [id, comments])
 
   const renderFile = ({ hunks, newPath }: File) => {
     const newValue: string[] = []
-
     if (!hunks.length || !newPath) return
     let extension = newPath.split('.').pop()!
     if (!prismLanguages.includes(extension)) extension = 'javascript'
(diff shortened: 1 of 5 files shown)


anthonykhoa:
Can you explain the context of this PR more?
Is this PR the only one that implements the `inline comments` feature, or is this just 1 of many planned PRs to accomplish the feature? I checked out the storybook and am unable to add a new comment, unless a comment(`conversation`) was already previously made

components/DiffView.tsx
```diff
@@ -0,0 +1,104 @@
...
+  lessonId?: number
+  status?: string
+}> = ({ diff = '', id, name, username, comments, lessonId, status }) => {
+  const files = gitDiffParser.parse(diff)
+  type fileComments = Record<string, { lines?: number[]; comments?: Comment[] }>
+  //every file gets unique index in format of submissionId:fileName
+  const [commentsState, setCommentsState] = React.useState<fileComments>({})
+
+  useEffect(() => {
+    const commentMap: fileComments = {}
```
anthonykhoa, line 26: This should be refactored to use `reduce` instead of `forEach`, because you are using `forEach` to do what `reduce` already does. This change will make the code more understandable,

```
const commentMap = comments && comments.reduce(...)
```

Ulisseus:
It's one of many PRs to implement this feature. Hopefully, with next PR I will be able to enable it in production. 

>I checked out the storybook and am unable to add a new comment, unless a comment(conversation) was already previously made

Weird. When I click on line number in storybook a commentBox appears. Did you click on number or on a plus sign?
comment

Since it's a storybook and not a real component it does not send queries, so comments are not persistent. If you refresh the page everything is gone.

stories/components/DiffView.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+    submissionId: 1
+  },
+  {
+    content: 'Second Comment',
+    createdAt: '1620762275096',
+    authorId: 3,
+    line: 4,
+    author: {
+      username: 'newbie',
+      name: 'Noob Newbie',
```
anthonykhoa, line 37: This is awesome

anthonykhoa:
> It's one of many PRs to implement this feature. Hopefully, with next PR I will be able to enable it in production.
> 
> > I checked out the storybook and am unable to add a new comment, unless a comment(conversation) was already previously made
> 
> Weird. When I click on line number in storybook a commentBox appears. Did you click on number or on a plus sign?
> comment
> 
> Since it's a storybook and not a real component it does not send queries, so comments are not persistent. If you refresh the page everything is gone.

Ah yeah I deleted my comment before you replied xD
I clicked around more and saw that I had to click on a line number to add a comment.
What made me confused was that clicking on the `+` did not work to add a comment -- I had to click on the number. Can you change this, so that clicking on the `+` works to add a comment? I think the user experience would be best if it mirrors that of github. Mirroring `github user experience on adding comments` makes the process more simple for c0d3 users

Instead of adding comment by `clicking` on line number, I think clicking on line number should do what github does -- it alters the URL, and when you go to that URL, the line of the line number you clicked on will be focused upon.

Ulisseus:
>Instead of adding comment by clicking on line number, I think clicking on line number should do what github does -- it alters the URL, and when you go to that URL, the line of the line number you clicked on will be focused upon.

I'd like to do it but diff rendering is done by third party module which adds onClick event only to numbers. I already forked it to pass line numbers to rendering function but since I'm not really familiar with its inner workings I have no idea how long it will take me to modify it further. 

Maybe I could edit css to make line numbers stand out more.

ggwadera:
The `+` is from the diff, it means that line was added to the file. I think the line number should be used to add a comment since it's not part of the diff, just need a better visual indication that it does something, like the blue plus sign that appears on Github when reviewing.

anthonykhoa:
> The `+` is from the diff, it means that line was added to the file. I think the line number should be used to add a comment since it's not part of the diff, just need a better visual indication that it does something, like the blue plus sign that appears on Github when reviewing.

Yeah i agree I think there should be a visual indication like the blue plus sign, just like how github does it

components/DiffView.tsx
Ulisseus: I changed `forEach` to reduce.

Ulisseus:
I've updated this component a bit. 

* Line numbers change their background color and font-size on hover
* Clicking on line number in closed submission does nothing
* Clicking on line number with empty conversation removes commentBox from this line 

[Storybook](https://deploy-preview-814--youthful-thompson-448332.netlify.app/?path=/story/components-diffview--default)

components/DiffView.tsx
```diff
@@ -0,0 +1,138 @@
...
+    setCommentsState(commentsMap)
+    //rerunning useEffect on id rerenders submission when student clicks on another challenge
+  }, [id, comments])
+
+  const renderFile = ({ hunks, newPath }: File) => {
+    const newValue: string[] = []
+    if (!hunks.length || !newPath) return
+    let extension = newPath.split('.').pop()!
+    if (!prismLanguages.includes(extension)) extension = 'javascript'
+    hunks.forEach(hunk => {
```
hwong0305, line 65: We should start moving away from `forEach`. There's no need to use it.

hwong0305 (approved):

kondanna (approved):

songz (approved):

PR #823, opened 2021-05-26
update #585 - add line comments feature

Ulisseus:
This PR looks big but changes are pretty straightforward:

* I updated types definitions. Previously it was possible to get array of nulls for some types which doesn't make much sense.  It should be either null or valid data (`[Comment!]!` means there is always array of comments, `[Comment!]` means that it's either null or array of comments `[Comment]` allows null, array of nulls and array of comments).
*  I refactored out duplicated logic in `ReviewCard` and `ChallengeMaterial`. Both are using new `DiffView` component. 
~~A bit of props drilling to make it all work together.~~ I create global context which can store session to avoid props drilling.

I added a test submission (js0 first challenge), feel free to add comments (you need to do it trough vercel [deployment](https://c0d3-app-9qpt4uaz3-c0d3.vercel.app/curriculum)). Do not comment on normal student submissions yet, please.


ggwadera:
What is need for `session` in the props all over the place?

Ulisseus:
>What is need for session in the props all over the place?

`DiffView` needs username and name of current user. I could make two different props instead of session but it's probably a case for global context.

Ulisseus:
I added global context which can store session (and other properties, but there is only session now). You can check the latest vercel deployment.

I also found that resubmitting challenges does not clear previous comments in some cases. I assumed that every submission gets new id, but it looks like resubmitting simply overrides the diff. Since comments are using submissionId as foreign key it's possible to get old conversations in new submissions. I need to find some workaround.

ggwadera:
> I also found that resubmitting challenges does not clear previous comments in some cases. I assumed that every submission gets new id, but it looks like resubmitting simply overrides the diff. Since comments are using submissionId as foreign key it's possible to get old conversations in new submissions. I need to find some workaround. 

It would be nice to have previous comments displayed on a updated submission so if a different person reviews they can get more context on the changes.

But I understand that this may be too complicated to do with how we handle submissions currently, and because you save the comments by the line number.

Don't really know how we could handle that unless we change it to have multiple submissions for a single challenge.

ggwadera:
> It's possible to show previous submissions as separated diffs. So suppose user had submitted three versions for the same challenge - A, B, C (the latest). By default reviewer sees C but he could select previous submissions (like we can view previous commits on github) and he will get older conversations for each submission.

Yeah it's what I meant by multiple submissions. This is very similar to what [exercism.io](https://exercism.io) does where you have iterations for one problem where you can send more than one solution, except they send the entire file contents instead of doing diffs.

helpers/controllers/submissionController.ts
```diff
@@ -38,19 +38,19 @@ export const createSubmission = async (
...
           userId: Number(id)
         },
-        update: submissionData,
+        update: {
+          diff: submissionData.diff,
+          status: submissionData.status,
+          comments: {
+            deleteMany: {}
+          }
+        },
```
ggwadera, line 47: wow nice
did this work?

Ulisseus:
Updating submission deletes previous comments now. So every submission starts from clean state.

It was possible to remove unique constants on submissions allowing users to have multiply submissions for the same challenge but this way I need to handle a scenario where user resubmits without waiting for rejection. You could end up with multiply submission by the same user waiting for review. It's possible to solve it (show only the most recent one or automatically close all previous on resubmission) but the result would be the same and this pr is big enough already.

ggwadera (approved):

PR #825, opened 2021-05-27
Fix Curriculum lessons container overflow

### LINKED ISSUE #824: Overflow on the Curriculum lessons container
There's an overflow in the Curriculum lessons container that's caused by the CSS property `flex: 0 0 100%;` on the element with the class of `curriculum_child-scroll__3g0eX`

image

Removing the property `overflow: auto!important;` and replace it with `overflow: hidden;` should fix it.

flacial:
Fixes the overflow by applying `overflow: hidden;`

Resolves #824


Ulisseus:
Setting `overflow: hidden` breaks scroll on mobile (or any device with small screen size). 

It's supposed to be like this: initially you get lesson list with arrow prompting you to scroll right (or swipe right  on mobile). If you scroll you see discord bar, announcement and additional resources. 
Screen Shot 2021-05-29 at 11 41 00
Screen Shot 2021-05-29 at 11 46 37
If you hide overflow it won't be possible to get the second screen. Check vercel preview and compare it to live version.

flacial:
@Ulisseus You're right. It'll break the mobile swipe feature.

It seems the scrollbar only appears on Brave. I've tried Chrome and Firefox.

Issue #843, opened 2021-06-07
Discord Integration


songz:
# Problem
Ever since [migration to Discord from Mattermost](https://github.com/garageScript/c0d3-app/wiki/Process-Changes), when students make submissions, our Discord server doesn’t get notifications.

# Goal
Our goal is to bring back the feature to keep mentors / students more engaged and be aware of student activity. This way, students could get faster code review feedback. 

1. When users submits a challenge for a lesson X, we should update our discord channel for lesson X + 1 to students who have completed the lesson know that there has been a new submission. 
2. When a submission has been approved, we check to see if the user has completed all the challenges for the given lesson. If so, we progress them to the next lesson and send a congratulatory message to the lesson X channel. And then welcome the user into the lesson X+1 channel.
3. We may also want to display a user's discord information on their c0d3 profile page. 

In the future, we may just use Discord login instead of c0d3.com login. Also, possibly notify student when a comment is made on a submission.

To accomplish these features, we need to solve the following high level problems:

1. How to correlate users’ Discord account to their c0d3.com account? (OAUTH after login)
2. How to notify the lesson’s corresponding channel when users make a submission?

# Architecture

# Research
Since all of the unknowns are within the Discord API, our research focuses mainly around answering questions about the Discord API.

**Correlating User’s Discord Account to c0d3.com account**
* Best way to do this is using [discord OAUTH2 API](https://discord.com/developers/docs/topics/oauth2). This gives us the refresh token, auth token, which we can then use to fetch user’s discord info.
* [test demo](https://demo.songz.dev/discord) and [result](https://demo.songz.dev/discord/profile)
* [next-auth](https://next-auth.js.org/adapters/models) has an adapter for prisma and built in sign in functions and session handling, but we would need to add 2 or 3 more models to our schema and they provide no way of retrieving another user's discord info in real time, which is what we want

**How to notify the lesson’s corresponding channel when users make a submission**
* Create a bot and invite it to our guild/server. [Sample link](https://discord.com/oauth2/authorize?client_id=845470281283665970&scope=bot+applications.commands)
* [Official bot nodejs module](https://github.com/discordjs/discord.js#readme)

**How to get channel ID**
* Go to User Settings > Advanced > Activate Developer Mode
* Now when you right click on channel, there's an option to copy channel ID

**Terminologies**
[...]

PR #859, opened 2021-06-11
Update #843 - Adds backend for discord integration 

kondanna:
- adds utility functions for fetching data
- adds refresh token to User table 

Todo:
- add refresh token to userInfo query


helpers/discordAuth.ts
```diff
@@ -0,0 +1,76 @@
+import { prisma } from '../prisma'
+const qs = require('qs')
```
Ulisseus, line 2: Why not simple import? Also, do we really need a new library for this? As far as I understand `qs` helps with deeply nested string parsing.

kondanna: ok, i replaced it with URLSearchParams like they did here:
https://discordjs.guide/oauth2/#authorization-code-grant-flow

helpers/discordAuth.ts
```diff
@@ -0,0 +1,77 @@
...
+const client_id = process.env.DISCORD_KEY
+const client_secret = process.env.DISCORD_SECRET
+
+type DiscordUserInfo = {
+  discordUsername: string
+  discordAvatarUrl: string
+  discordRefreshToken: string
+}
+
+export const getTokenFromAuthCode = (code: string) => {
```
ggwadera, line 13: Please add the return type for these JSONs. Don't wanna have to deal with `any` types specially with authentication.

https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-access-token-response

helpers/discordAuth.ts
```diff
@@ -0,0 +1,108 @@
...
+  email: string
+  verified: boolean
+}
+
+type DiscordUserInfo = {
+  discordUsername: string
+  discordAvatarUrl: string
+  discordRefreshToken: string
+}
+
```
kondanna, line 33: how do I test these functions? seems pointless to make mock inputs and outputs.

helpers/discordAuth.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (!tokenResponse.refresh_token) throw new Error('refresh token invalid')
+
+  const { username, avatar } = await getUserInfo(tokenResponse.access_token)
+
+  return {
+    discordUsername: username,
+    discordAvatarUrl: avatar,
+    discordRefreshToken: tokenResponse.refresh_token
+  }
+}
```
JasirZaeem, line 109: Can you return the Discord user's id as well, it will be needed to integrate the bot and let it dm or mention a user.
`const {username, avatar, id} = ...`
`return { 
  discordId: id`
And add it to the `DiscordUserInfo` type

graphql/typeDefs.ts
```diff
@@ -143,6 +143,7 @@ export default gql`
     name: String!
     isAdmin: Boolean!
     cliToken: String
+    discordRefreshToken: String
```
songz, line 146: we shouldn't expose this to the front end.

helpers/discordAuth.ts
```diff
@@ -0,0 +1,111 @@
...
+  return fetch(`${discordAPI}/users/@me`, {
+    method: 'GET',
+    headers: {
+      'content-type': 'application/x-www-form-urlencoded;charset=utf-8',
+      Authorization: `Bearer ${accessToken}`
+    }
+  }).then(r => r.json())
+}
+
+export const updateUserRefreshToken = async (
```
songz, line 82: do we want to export this function? I was under the impression that only `getUserInfoFromRefreshToken` is required, and therefore only that function should be exported?

graphql/typeDefs.ts
kondanna: oh true, got carried away

helpers/discordAuth.ts
kondanna: true, i removed the two functions from exports. the others may be needed.

songz: oh I remembered... you must always call `updateUserRefreshToken` after getting the accessToken. This is because after getting user's accessToken, the refresh token becomes invalid and you need to get a new one. so this function is a helper function that shouldn't be exported.

kondanna: I see now. I have a case during login where I just need to check if refreshToken exists and is valid, and no userInfo needs to be fetched. Therefore I just store the new refreshToken in that case, using the functions getTokenFromRefreshToken and updateUserRefreshToken

121316637-3292ae00-c912-11eb-93d9-12d7dc675ab9

helpers/discordAuth.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  if (!updatedRefreshToken) throw new Error('refresh token invalid')
+
+  const { id, username, avatar } = await getUserInfo(tokenResponse.access_token)
+
+  return {
+    discordUserId: id,
+    discordUsername: username,
+    discordAvatarUrl: avatar,
+    discordRefreshToken: updatedRefreshToken
```
ggwadera, line 113: I think these names could be less verbose. From the type name we know that it's related to the Discord account, don't need that `discord` prefix in every variable.

graphql/typeDefs.ts
kondanna: hmm I think I want it in both the AuthResponse (upon login) and userInfoController. is a separate prisma query for the refreshToken more appropriate?

ggwadera:
By the way, what's the motive in writing this from scratch instead of using something like [next-auth](https://next-auth.js.org/)? Seems like it could save a lot of work, would only need to focus getting the bot working.

kondanna:
> By the way, what's the motive in writing this from scratch instead of using something like [next-auth](https://next-auth.js.org/)? Seems like it could save a lot of work, would only need to focus getting the bot working.

No reason. The only pages I can find on it: https://next-auth.js.org/providers/discord
and
https://github.com/nextauthjs/next-auth/blob/main/src/providers/discord.js
so it still looks like I still have to do the fetches myself.

ggwadera:
> No reason. The only pages I can find on it: [next-auth.js.org/providers/discord](https://next-auth.js.org/providers/discord)
> and
> [https://github.com/nextauthjs/next-auth/blob/main/src/providers/discord.js](https://github.com/nextauthjs/next-auth/blob/main/src/providers/discord.js?rgh-link-date=2021-06-23T14%3A57%3A27Z)
> so it still looks like I still have to do the fetches myself.

Look at the general providers configuration, the discord one just show the options it supports.

https://next-auth.js.org/configuration/providers

There's also an example app https://github.com/nextauthjs/next-auth-example

kondanna:
https://next-auth.js.org/adapters/prisma

helpers/discordAuth.ts
```diff
@@ -0,0 +1,115 @@
...
+  updateUserRefreshToken(userId, updatedRefreshToken)
+
+  if (!updatedRefreshToken) throw new Error('refresh token invalid')
+
+  const { id, username, avatar } = await getUserInfo(tokenResponse.access_token)
+
+  return {
+    userId: id,
+    username,
+    avatarUrl: avatar,
```
songz, line 112: this `avatar` is not a url, you have to construct this and I think you should construct it here

jest.config.ts
```diff
@@ -15,7 +15,8 @@ const config: Config.InitialOptions = {
     'graphql/index.tsx',
     '__tests__/utils/',
     'prisma/',
-    'scripts/'
+    'scripts/',
+    'discordAuth.ts'
```
songz, line 19: could we write tests for this instead of excluding it from tests? I see a few places where it could go wrong.

For example, when you call `getUserInfoFromRefreshToken`, you want to make sure:

1. `if (!updatedRefreshToken) throw new Error('refresh token invalid')`. 
2. `updateUserRefreshToken` is called

I think if you tested these 2 you should have 100% coverage on this file because that's the only branching.

kondanna:
I spent some time looking through next-auth docs and some examples. They do have a prisma adapter, but it looks like we would have to add a few additional tables, and if we want to look up another user's most updated discord info (username and avatar), we'd still have to roll out our own functions. 

It's debatable whether someone's discord info changes much at all, but for now I think it's less complex to do it this way.

helpers/discordAuth.ts
kondanna: fixed

jest.config.ts
kondanna: removed

helpers/discordAuth.ts
songz: yeah so from what you are saying, only `getUserInfoFromRefreshToken` needs to be exported right?

ggwadera:
> it looks like we would have to add a few additional tables

That's true, but we'll eventually need to do that anyway in the future to support more advanced features. Having all these things in the users table could make it too bloated in the future. It'd better if we could have separate tables for better division of concerns.

> if we want to look up another user's most updated discord info (username and avatar), we'd still have to roll out our own functions

That's also true. I don't think we really need to fetch that too frequently though, just updating that when signing in is fine in my opinion. With `next-auth` we could that with a [Sign in callback](https://next-auth.js.org/configuration/callbacks#sign-in-callback), something like this:

```js
callbacks: {
  async signIn(user, account, profile) {
    if (user.id) {
      await prisma.user.update({
        where: { id: user.id },
        data: { image: profile.image_url }
      });
    }
    return true
  }
}
```

You're right that we'd still have to roll out our functions, but it's a lot less code than writing a whole authentication system.

All in all I'd still prefer a production ready module for this feature that we can work on integrating it with the project, specially as it's dealing with sensitive stuff that is authentication and authorization. I'll leave this up for discussion if you have any opinions on this.

If we choose to use `next-auth` our biggest concern would be how to make it work with our current tables and data, and also move the session handling to it. From the prisma [schema](https://next-auth.js.org/adapters/prisma#setup) they have on the docs it doesn't look to different from ours, so that shouldn't be too hard. And all this can be done gradually, like we've done with the prisma migration. The discord integration could be prioritized and worked on so we can roll out that more quickly.

I've made some changes to the example app they have on Github to see how it works, and added a Github and Discord oauth provider with the prisma database adapter, and also that callback to update the user image. It worked without any issue, I'm pretty impressed with that actually haha. If you wanna see it I can upload to a repo here.

kondanna:
Sure, let's see that code. I actually tried it with our c0d3 repo but couldn't figure out how to get the redirect_uri to work on the login() page, so I gave up.

I definitely wouldn't mind getting next-auth into the project. In fact, months ago I helped the woman integrate it into her personal project (without the database part though) and it looked pretty nice. I could then remove the getTokenFromAuthCode function here.

ggwadera:
Here you go: https://github.com/ggwadera/next-auth-example-prisma

kondanna:
I deleted forked repo and branch. Resubmitting PR.

Issue #874, opened 2021-06-18
Add a way for a reviewer to see previous submissions [feature]


Ulisseus:
**Why is this needed?**
Being able to compare previous attempts would allow reviewer to provide better feedback. Currently, only the most recent submission is visible and all comments are reset on re-submission. 

Every diff should get a new options menu (simple gear icon with dropdown menu). When you click on it you get submission dates (we don't store hashes or commit messages). Clicking on date should rerender the diff and its line comments (comments under submission should be shared across all attempts).  

**Potential problems**
Right now our db enforces unique index `@@unique([userId, lessonId, challengeId])` meaning that one user can have only one submission per challenge. If I remove this restriction it would be possible to submit multiply files for the same challenge (say submission A, B and C being the latest). When reviewer opens the page he will get all three submissions. If reviewer accepts A what should happen with B and C and vice versa?

One possible solution is to close all other submissions by the same user on clicking accept/reject (reviewer rejects B, A and C are also being rejected). Another way is to close previous attempt on re submission, meaning that B closes A, and C closes B, so only one submission is active at any time.

ggwadera:
If possible I'd like to keep the state of previous submissions unchanged, just so we can keep a history of accepted or rejected submissions for a given challenge, meaning not accepting/rejecting all previous submissions. This way we can also get better statistics on Metabase.

One issue that may appear is someone resubmitting when the submission hasn't been reviewed yet, which would leave 2 `OPEN` submissions from the same user. For this case I suggest adding a new `OUTDATED` status for the submission table so the previous submission is not considered for review. Or just replace the newest open one if you don't wanna have any more trouble.

PR #914, opened 2021-06-28
update #874 - remove index from submission table

Ulisseus:
This PR removes unique requirement for submission, meaning that our db will be able to store multiply submissions for the same challenge by the same user. 

Submission controller checks if there is an existing open submission and changes its status to 'overwritten' before creating new row in db. 

Reviewing process won't be affected by this change. Students and reviewers will see only the last open submission.

### DIFF
--- a/helpers/controllers/submissionController.ts
+++ b/helpers/controllers/submissionController.ts
@@ -22,34 +22,39 @@ export const createSubmission = async (
     const { challengeId, cliToken, diff, lessonId } = args
     const { id } = decode(cliToken)
     const submissionData = { diff, status: SubmissionStatus.Open }
-    const { lesson, challenge, user, ...submission } =
-      await prisma.submission.upsert({
+    const previousSubmission = await prisma.submission.findFirst({
+      where: {
+        challengeId,
+        lessonId,
+        user: {
+          id
+        },
+        status: SubmissionStatus.Open
+      }
+    })
+    if (previousSubmission) {
+      await prisma.submission.update({
         where: {
-          userId_lessonId_challengeId: {
-            userId: Number(id),
-            lessonId,
-            challengeId
-          }
+          id: previousSubmission.id
         },
-        create: {
-          ...submissionData,
-          challengeId,
-          lessonId,
-          userId: Number(id)
-        },
-        update: {
-          ...submissionData,
-          comments: {
-            deleteMany: {}
-          }
-        },
-        include: {
-          user: true,
-          lesson: true,
-          challenge: true
+        data: {
+          status: SubmissionStatus.Overwritten
         }
       })
-
+    }
+    const submission = await prisma.submission.create({
+      data: {
+        ...submissionData,
+        challengeId,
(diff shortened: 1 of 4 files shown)


helpers/controllers/submissionController.ts
```diff
@@ -22,34 +22,39 @@ export const createSubmission = async (
...
         }
       })
-
+    }
+    const submission = await prisma.submission.create({
+      data: {
+        ...submissionData,
+        challengeId,
+        lessonId,
+        userId: Number(id)
```
ggwadera, line 50: You can remove the `Number()` conversion here, that id should already be a number now I think.

helpers/controllers/submissionController.ts
```diff
@@ -22,34 +22,39 @@ export const createSubmission = async (
...
-          challenge: true
+        data: {
+          status: SubmissionStatus.Overwritten
         }
       })
-
+    }
+    const submission = await prisma.submission.create({
+      data: {
+        ...submissionData,
```
ggwadera, line 47: If you want you can inline these values here. I only created this `submissionData` to avoid duplicated values in the upsert create and update.

helpers/controllers/submissionController.ts
```diff
@@ -22,34 +22,39 @@ export const createSubmission = async (
...
-          user: true,
-          lesson: true,
-          challenge: true
+        data: {
+          status: SubmissionStatus.Overwritten
         }
       })
-
+    }
+    const submission = await prisma.submission.create({
```
ggwadera, line 45: You can move the return here now and the await won't be needed anymore. This function returns a promise either way.

helpers/controllers/submissionController.ts
```diff
@@ -22,34 +22,39 @@ export const createSubmission = async (
...
+        ...submissionData,
+        challengeId,
+        lessonId,
+        userId: Number(id)
+      },
+      include: {
+        user: true,
+        lesson: true,
+        challenge: true
+      }
```
ggwadera, line 56: These includes were used only for the chat notification, because this query only expects to return the id and diff of the submission, so they can be removed until we get that back. I should've removed them on my previous mattermost deprecation PR but forgot that.

Or just leave that commented out for when someone implements the chat notification back. At least we avoid some unneded table joins for now.

helpers/controllers/submissionController.ts
Ulisseus: It's not a number, decoder returns any, but prisma doesn't enforce types here, so I removed this conversion.

helpers/controllers/submissionController.ts
Ulisseus: No need to keep these unused fields, so I removed them too.

helpers/controllers/submissionController.ts
ggwadera: Oh right it's decoded from the cli token. Should probably type that function correctly someday.

ggwadera (approved):

Issue #916, opened 2021-06-28
Should we remove all env variables from next.config.js


tomrule007:
We currently have our environment variables defined in our [next.config.js](https://github.com/garageScript/c0d3-app/blob/master/next.config.js) but it appears to be an artifact of an old version of Next.

(as @ggwadera  pointed out here https://github.com/garageScript/c0d3-app/issues/906#issuecomment-869694620 )

Benefits
- Removes an extra location to hide default values (all env variables should be in the `.env` file
- Extra Security preventing keys from accidentally being leaked to the frontend code. (Next will only expose keys with `NEXT_PUBLIC_*` prefix to the frontend)
-
cons
- ??

I'm still getting up to speed on typescript / next / c0d3-app so I might be missing something but It seems to be working on my end and maybe we can deploy a demo branch on vercel to confirm its working there as well.

hwong0305:
Since this is no longer needed, I think the logging PR is a good time to make this refactor!

tomrule007:
rolled into issue #906

PR #917, opened 2021-06-28
WIP: Remove env from config

tomrule007:
Suggested changes to remove env variables from `next.config.js` https://github.com/garageScript/c0d3-app/issues/916

*Note as is this will break the sentry frontend log reporting since it isn't updated to take the `NEXT_PUBLIC_SENTRY_DSN` variable


tomrule007:
Ok so the security concern was a bit of a reality. 

The first 1 build failed because we didnt have the `dp_port` variable set. 

I dont think exposing that is to bad but,

the db still wasn't working because we also didnt have our `session_secret` and were defaulting to our publicly set key that is visible to anyone.

I'm guessing there is a way to hijack sessions with that key out in the public, probably not terrible but should be fixed.

ggwadera:
Included in #926

PR #923, opened 2021-06-30
close #874 - implement selection of previous iterations for reviewer and user

### LINKED ISSUE #874: Add a way for a reviewer to see previous submissions [feature]

Ulisseus:
This PR is very big, but unfortunately I cannot break it down into smaller pieces, everything is tightly coupled together. I could try to separate changes for reviewer and student pages but it would lead to weird temporary components and helpers which would be overwritten by the very next pr. The initial pr would be about the same size minus a few files. 

Main changes are in ReviewCard and ChallengeMaterial components. I'd suggest starting review with ReviewCard, it's shorter and a bit easier to read. 

### ReviewCard

Currently we're using GetSubmissions query to fetch all open submission for the specified lesson. This PR adds additional query GetPreviousSubmissions which is triggered for every ReviewCard component once it mounts. It's non-blocking data fetching, so render speed will be exactly the same as before. 

Is it possible to fetch everything in one query? I don't know, I guess some sql wizard could have done it, but Prisma doesn't support subqueries yet, and the raw sql would be really big and clunky to maintain.

There are two new states - index and submission. Submission is the current selected submission and index is its index in getPreviousSubmissionQuery data.

The initial component got too big, so I split it in two. Review buttons are now in their own ReviewButtons component, their functionality is the same.

### ChallengeMaterial

ChallengeQuestionCard component also got too big. Diff rendering is now done in a separate component -  ChallengeQuestionCardDisplay. 

Again, there are two new states - index and submission. But unlike ReviewCard there are two useEffects too. One is for updating data after receiving getPreviousSubmissions query and changing current index just like in ReviewCard. Another one is for resetting index when user clicks on different challenge.

### updateCache

Both components use the same updateCache helper function which updates getPreviousQuery client side. 

For people who are not familiar with updating client cache in apollo. Only server side is updated when you use a mutation, so you need to manually override cache in order to reflect current state of component. Once you do this all updated queries-hooks are triggered which results in rerendering. It's a data loop of sorts. 

### DiffView

This component had a lot of different props but they all were parts of submissions type. So istead of adding another one I simply passed the whole submission. GeneralStatus is total status of submission (for example, you could have five iterations from getPreviousSubmissions query - 'needMoreWork', 'overwritten', 'overwritten', 'needMoreWork', 'Open') with generalStatus 'Open'. 

### CommentBox

Again, the same generalStatus. Also, I choose to show comments even on rejected submissions, hiding them makes no sense when you can select previous submissions. Adding comments is disabled when generalStatus isn't 'Open' though.

### Test file changes

I added new dummy data file - getPreviousSubmissionsData. We should unify our test data someday, currently most tests use their own data. It makes test files harder to read but more importantly if you update any grapql type you need to update every test mock (grapqh enforces types even in tests).

---

[Vercel](https://c0d3-app-qj6qpxosg-c0d3.vercel.app/) (I created test submission under js0, do not comment on students previous iterations yet, users won't be able to see your messages).


components/ReviewCard.tsx
```diff
@@ -106,6 +111,110 @@ export const ReviewCard: React.FC<ReviewCardProps> = ({ submissionData }) => {
...
+    challengeId,
+    user.id
+  )
+  const underComments = submissionState.comments?.filter(
+    comment => !comment.line
+  )
+  useEffect(() => {
+    if (data?.getPreviousSubmissions) {
+      setPreviousSubmissions(data)
+      /*istanbul ignore else*/
```
Ulisseus, line 208: I spent quite a lot of time trying to trigger this else case but to no avail. I couldn't get the final rerender from updating the cache in tests no matter what.

components/ChallengeMaterial.tsx
```diff
@@ -14,6 +19,9 @@ import { MdInput } from './MdInput'
 import { updateCache } from '../helpers/updateCache'
 import { GlobalContext } from '../helpers/globalContext'
 import { SubmissionComments } from './SubmissionComments'
+import { useEffect } from 'react'
```
tomrule007, line 22: Should be moved into the other react imports on line 1

components/ChallengeMaterial.tsx
```diff
@@ -139,35 +147,137 @@ export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = ({
...
+  const userId = context.session?.user?.id
+
+  const [index, setIndex] = useState(-1)
+
+  const [submission, setSubmission] = useState(currentChallenge.submission)
+  const reviewerUserName = submission?.reviewer?.username || null
   const comments =
-    currentChallenge.submission &&
-    currentChallenge.submission.comments?.filter(comment => !comment.line)
+    submission && submission.comments?.filter(comment => !comment.line)
```
tomrule007, line 165: Could you remove the `submission && ....`  and just do `submission?.comments?.filter(comment => !comment.line)`

components/ChallengeMaterial.tsx
```diff
@@ -139,35 +147,137 @@ export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = ({
...
+  })
+  useEffect(() => {
+    if (data?.getPreviousSubmissions) {
+      if (data.getPreviousSubmissions[index])
+        setSubmission(data.getPreviousSubmissions[index] as Submission)
+      else
+        setSubmission(
+          data.getPreviousSubmissions[
+            data.getPreviousSubmissions.length - 1
+          ] as Submission
```
tomrule007, line 185: could use lodash here and just do `_.last(data.getPreviousSubmissions) as Submission`

components/ReviewCard.test.js
```diff
@@ -1,16 +1,17 @@
...
 import { MockedProvider } from '@apollo/client/testing'
-import _ from 'lodash'
-import '@testing-library/jest-dom'
 import { SubmissionStatus } from '../graphql'
+import { ContextProvider, GlobalContext } from '../helpers/globalContext'
 import dummySessionData from '../__dummy__/sessionData'
-// correct javascript submission
-const JsDiff =
-  'diff --git a/js7/1.js b/js7/1.js\nindex 9c96b34..853bddf 100644\n--- a/js7/1.js\n+++ b/js7/1.js\n@@ -1,8 +1,19 @@\n-// write your code here!\n const solution = () => {\n-  // global clear all timeout:\n+  const allT = [];\n+  const old = setTimeout;\n+  window.setTimeout = (func, delay) => {\n+    const realTimeout = old(func, delay);\n+    allT.push(realTimeout);\n+    return realTimeout;\n+  };\n+  window.clearAllTimouts = () => {\n+    while (allT.length) {\n+      clearTimeout(allT.pop());\n+    }\n+  };\n   cat = () => {\n-  }\n+    window.clearAllTimouts();\n+  };\n };\n \n module.exports = solution;'
+import previosSubmissionsData from '../__dummy__/getPreviousSubmissionsData'
```
tomrule007, line 13: typo on import name, missing `u` in : `previous`

components/ChallengeMaterial.tsx
anthonykhoa: Since we are using lodash, I think using this is better to `_.get` lodash function to maintain codebase consistency. With lodash, I'm not sure this repo needs to use the `?.` operator.
```
_.get(submission, 'comments').filter(...)
```

Ulisseus: Tom is right, looking at this part I am not sure why I needed `submissions&&` check in the first place.

Chaining `?` feels more natural than using lodash in this case. If anything we should start replacing `_.get` with `?` operators everywhere.

components/ChallengeMaterial.tsx
Ulisseus: Done. I wish javascript supported negative indexes. `array[ -1 ]` would be so much better her.

Ulisseus:
There is something off with rejection comment on vercel preview, the last open submission should not have rejection message. I'll check it tomorrow, this is not the intended behavior.

components/ChallengeMaterial.tsx
ggwadera: Agreed on using the coalescing chain. It is more natural and works better with typescript. Lodash uses a string for the property which may cause bugs if the name changes for some reason.

ggwadera:
As a small suggestion, please add a screenshot of the new or changed component when you're working on front-end code. It's easier to grasp what the PR is about this way. I know that there is the storybook preview but it's easier if you have a little screenshot on your description.

That said, I think the current styling has way too much whitespace, and that weird box around it.

image

Maybe the submissions selector could go on the top right, after the challenge name, and make it a little smaller and the boxes more close together. Also that `Iteration 3 of 4` text is kinda redundant since it's the same information as the boxes display.

Ulisseus:
>please add a screenshot of the new or changed component when you're working on front-end code

Maybe I should have added general layout with this component. But storybook enforces types, so it's rather cumbersome creating all these dummy files according to proper type definitions. 

Anyways, selection counter is now in the top right corner. And accept/reject buttons are hidden for previous submissions (check updated vercel). 
counter

components/ChallengeMaterial.tsx
```diff
@@ -139,35 +152,131 @@ export const ChallengeTitleCard: React.FC<ChallengeTitleCardProps> = ({
...
   const [commentValue, setCommentValue] = React.useState('')
   const [addComment] = useAddCommentMutation()
-  const update = updateCache(
-    currentChallenge.submission?.id!,
-    commentValue,
-    name!,
-    username!
-  )
+
+  const { data, loading, error } = useGetPreviousSubmissionsQuery({
```
ggwadera, line 174: Why not reduce this query to only get the past submissions ids and then fetch them on demand by its id? Querying them all here with the diff will increase the loading time when opening the review page (meaning the full screen loading spinner).

Ulisseus: I am not sure how it should work. GetPreviousSubmisisons query doesn't affect fullscreen loading spinner at all. 

Current fetching order is:
* getApp query (it should be in client's cache, and if this is not the case user is being redirected) 
* getSubmission query fetches all open submissions for this lesson (this is what actually hidden behind fullscreen loading spinner)
* Review page is rendered. getPreviousSubmissions query starts, select iteration in the top right corner is showing "Loading submissions...".
* getPreviousSubmissions returns and submission selection is fully rendered. So it doesn't affect loading spinner duration at all, it only increases time to final render but most users probably won't mind it (nobody notices a small component in the top right corner). 

Alternatives:
* Fetch everything in one query. First, I don't know how to do it (it's not possible in prisma and raw sql would be really big). Second, and more important this would actually increase loading spinner duration while reducing time to final render. This tradeoff would lead to bad UX in my opinion, why wait for full data when you can render it bit by bit and making page look more responsive?

* Use lazy loading. This again would be a worse UX. When you click on iteration you need to wait for query to return. Why not fetch it on first render while user is focused on reading initial diff?

Come to think about it, we need to get rid of fullscreen spinners someday. Nothing here prevents rendering of NavBar, Footer and Lesson components. Maybe even add a template loading submission just like youtube does it with gray rectangles.

components/ChallengeMaterial.tsx
tomrule007: Ya I would vote moving to the coalescing operator when possible. A quick global search on the project for `_.get(` returns 40 hits in 13 files, seems like we should be able to refactor it out of the codebase at some point in the future. <br> \* This could be a very jr. opinion, as I haven't seen how bad projects can get when the consistency slowly gets destroyed with promises of future refactoring.

components/ReviewCard.tsx
Ulisseus: I've managed to finally trigger that branch by hardcoding comment for every cache update. I guess it's better than completely ignoring it.

components/ChallengeMaterial.tsx
ggwadera: Yeah you're right, my mistake. We'll have to see how it will work out with big files like on JS6, the submission selector could take a longer time to load since it will have to fetch the entire submissions.

> Come to think about it, we need to get rid of fullscreen spinners someday. Nothing here prevents rendering of NavBar, Footer and Lesson components. Maybe even add a template loading submission just like youtube does it with gray rectangles.

Yes having a skeleton placeholder would be ideal.

ggwadera:
Accepted submissions without comment are not showing who was the reviewer.

image

Should look like this one but without the approval comment text.

image

Ulisseus:
>Accepted submissions without comment are not showing who was the reviewer.
It should work correctly even without a comment. These submissions were overwritten by user without waiting for approval/rejection.

I need to work on these review cards more though. Maybe add another one for resubmission. And refactor them into separate component to use in challenge page.

But this PR is too big already, so I'd rather make another one for these changes.

components/ChallengeMaterial.tsx
ggwadera: Agreed on sprint Honeydew to use and replace `_.get` with the coalescing chain operator `?.`.

kondanna (approved):

ggwadera (approved):

Issue #924, opened 2021-06-30
Bug: MaxListenersExceededWarning: Possible EventEmitter memory leak detected


tomrule007:
## Bug

This maybe only a local issue to my machine but I thought I would document it just encase others are having it.

In my server console when running either: `yarn start` or `yarn dev` shortly after first load of the page I am getting this  `MaxListenersExceededWarning: Possible EventEmitter memory leak detected` warning.

running `NODE_OPTIONS='--trace-warnings' yarn dev` i was able to get the stack trace and included it below.

I'm not sure if the logger files are causing the issue or they are just reporting the issue?

## Setup

**commit:**  17e6818350b0d90369f83c58d8b79ccea7348c77 (current master)
**environment**: Local Dev (debian 9 linux)

## Details

```
(node:14074) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:451:17)
    at process.addListener (events.js:467:10)
    at ExceptionHandler.handle (/home/thomas/JS/c0d3-app/node_modules/winston/lib/winston/exception-handler.js:51:15)
    at DerivedLogger.configure (/home/thomas/JS/c0d3-app/node_modules/winston/lib/winston/logger.js:139:23)
    at new Logger (/home/thomas/JS/c0d3-app/node_modules/winston/lib/winston/logger.js:42:10)
    at new DerivedLogger (/home/thomas/JS/c0d3-app/node_modules/winston/lib/winston/create-logger.js:44:7)
    at Object.module.exports [as createLogger] (/home/thomas/JS/c0d3-app/node_modules/winston/lib/winston/create-logger.js:48:18)
    at winstonLogger (webpack-internal:///./helpers/middleware/logger.ts:40:57)
    at Array.__webpack_exports__.default (webpack-internal:///./helpers/middleware/logger.ts:60:18)
    at loop (/home/thomas/JS/c0d3-app/node_modules/next-connect/dist/index.cjs:69:47)
    at next (/home/thomas/JS/c0d3-app/node_modules/next-connect/dist/index.cjs:74:13)
    at Function.handle (/home/thomas/JS/c0d3-app/node_modules/next-connect/dist/index.cjs:77:5)
    at /home/thomas/JS/c0d3-app/node_modules/next-connect/dist/index.cjs:57:12
    at new Promise (<anonymous>)
[...]

PR #926, opened 2021-07-01
close #906 - Feature/sentry logging revamp

### LINKED ISSUE #906: Improve production error logging
We currently use Sentry to capture errors on the production deployment but it seems it's not doing a very good job because we're mostly clueless whenever a bug happens to show up. I don't know if it's a configuration issue with Sentry or we're just not doing things correctly to send the logs.

Either way, it's worth to take some time to read the docs to better understand this and try to fix, so we can have more reliable logging for bug fixing. Alternatives to Sentry are also welcome if it's reasonable enough to warrant a migration.

Another thing is that logs from dev environments are also being sent, which leads to some confusion when going over the errors on Sentry. Might have to add a condition to only init Sentry when running on production. https://github.com/garageScript/c0d3-app/blob/d35f563149901dbcde67a7e3619ca23d6b58eaa9/pages/_app.tsx#L20-L22

So basically, this is what is desired to be improved:

- Disable error reporting outside of production environment
- Fix all the logs that aren't being sent from the production deployment
- Include the serverless function errors logs in the report, or setup a log drain on Vercel

Some useful resources:

- Vercel and Sentry integration https://docs.sentry.io/product/integrations/vercel/
- Sentry guide for Next.js https://docs.sentry.io/platforms/javascript/guides/nextjs/
- Sentry docs https://docs.sentry.io/
- How to store logs on Vercel https://vercel.com/support/articles/how-do-i-store-logs-on-vercel

tomrule007:
Checking vercel preview functionality before finishing:
- [x] apolloSentryErrorPlugin tests (remove api/graphql from coverage ignore list)
- [x] apolloSentryErrorPlugin Types (replace any with correct type)

It currently should catch all errors and can be tested with invalid username or password login attempts

See Issue #906 for details.


ggwadera (commented):
Looks good. Only need to remove the console.logs and resolve conflicts by merging upstream master into your branch.

next.config.js
```diff
@@ -8,19 +9,22 @@ const withMDX = require('@next/mdx')({
...
+  pageExtensions: ['tsx', 'js', 'jsx', 'mdx', 'ts']
 })
+
+const SentryWebpackPluginOptions = {
+  // Additional config options for the Sentry Webpack plugin. Keep in mind that
+  // the following options are set automatically, and overriding them is not
+  // recommended:
+  //   release, url, org, project, authToken, configFile, stripPrefix,
+  //   urlPrefix, include, ignore
+  debug: false,
```
Ulisseus, line 22: According to docs debug is false by default. This line is unnecessary.

next.config.js
```diff
@@ -8,19 +9,22 @@ const withMDX = require('@next/mdx')({
...
+  silent: true // Suppresses all logs
```
Ulisseus, line 23: Could you add a short comment explaining why do we need to supress logs?

helpers/middleware/logger.ts
```diff
@@ -37,29 +34,16 @@ export const winstonLogger = (sessionId: string) => {
...
+    transports: [new winston.transports.Console()]
   })
 }
 
 export default (req: LoggedRequest, res: NextApiResponse, next: () => void) => {
   const uid = nanoid()
   const logger = winstonLogger(uid)
+
   req.info = (val: any) => {
+    console.log('capture info')
```
Ulisseus, line 46: Did you forget to remove this console.log?

helpers/apolloLogPlugin.ts
```diff
@@ -0,0 +1,64 @@
...
+
+            const transactionId =
+              ctx.request.http?.headers.get('x-transaction-id')
+            if (transactionId) {
+              scope.setTransactionName(transactionId)
+            }
+            Sentry.captureException(err)
+          })
+          // Log error so logflare can also receive a copy of the error
+          console.error(err)
```
Ulisseus, line 54: Shouldn't we use winston-logger for printing errors?

pages/api/graphql.ts
```diff
@@ -7,6 +8,11 @@ import typeDefs from '../../graphql/typeDefs'
 import resolvers from '../../graphql/resolvers'
 import { PrismaSessionStore } from '@quixo3/prisma-session-store'
 import { prisma } from '../../prisma'
+import { apolloLogPlugin } from '../../helpers/apolloLogPlugin'
+
+// Temporary fix to broken sentry package.
```
anthonykhoa, line 13: Since this is temporary, can you write a `TODO: ` comment as well? For example `TODO: remove import when permanent fixed is implemented`

pages/api/graphql.ts
```diff
@@ -27,7 +34,7 @@ handler
   .use(loggingMiddleware)
   .use(
     session({
-      secret: process.env.SESSION_SECRET || '',
+      secret: process.env.SESSION_SECRET as string,
```
anthonykhoa, line 37: Would this create breaking changes if we passed in a value of undefined to this `secret` field? From what I understand, the `as` assertion doesn't do type coercion during runtime, it only tells the compiler to treat some value as a some specific type. So if `process.env.SESSION_SECRET` has an undefined value, then during runtime it will still be undefined as its value.

next.config.js
tomrule007: Correct, line is removed

next.config.js
tomrule007: how does this sound 
```
  // This plugin is used to send source map files to sentry.io during build time.
  // Silencing the log hides all the information about each individual file being
  // sent keeping the build logs cleaner and still logging any error that occurs.
  silent: true
```

helpers/middleware/logger.ts
tomrule007: I did, ty

helpers/apolloLogPlugin.ts
tomrule007: The main reason I opted to just use a simple console.error is because it plays nicely with our vercel hosted serverless setup. The winston setup was using an old sentry transporter that will not work with our vercel hosting, which is why we currently see no logs.

I also ran into issues with trying to get our custom `LoggedRequest` type to work with the new sentry wrapper `withSentry` in the api/lessons route and then also integrating it into the apollo plugin `GraphQLRequestContext` type. 

It's really about getting our logging to play nicely with AWS Lambda functions that are wrapped in some vercel function that isnt that well documented (they just say use our integrations).

I also believe our winston current setup is responsible for this error (which we are currently seeing in production) https://github.com/garageScript/c0d3-app/issues/924 and would like to phase it out completely in another pr if that is the case or at least fix the issue. With our serverless hosting winston just seems to add more headache than it's worth.

pages/api/graphql.ts
tomrule007: So I still dont know if this is the right way to do it but I believe we want to move away from this `env || default` pattern we are using because it prevents us from catching errors


If the env var isnt set at runtime it produces this warning 

`express-session deprecated req.secret; provide secret option webpack-internal:/pages/api/graphql.ts, <anonymous>:60:136`

If it is set to empty string we get no warning, which is bad as it will slip by unnoticed.

I documented it here that https://github.com/garageScript/c0d3-app/issues/916 (and the associated pr) that our current live build is using a default value that is publically available in our source code because no one set the env var on our production server. (so we basically don't have a session secret right now which is bad)

I should have have patched this on a seperate pr but agreed to roll it into this one (not realizing how long it would take me to complete)

pages/api/graphql.ts
tomrule007: added the todo comment, thanks.

pages/api/graphql.ts
anthonykhoa: @tomrule007 Thanks for the context, that makes sense!

Ulisseus (approved):

songz (approved):

songz (approved):

hwong0305 (approved):

PR #934, opened 2021-07-04
update #993 - fix rendering bug on challenge page

Ulisseus:
What happened? I accidentally passed initial value instead of state to query, so when user clicked on different challenge diff render wasn't updated (props were the same). And query failed (one of arguments was undefined).  

Also, with recent dayjs update we got a bunch of failing snapshots. I could mock date but I use it as unique id in one of test cases to break user's cache. I leave it as it is for now.

### DIFF
--- a/components/__snapshots__/ChallengeMaterial.test.js.snap
+++ b/components/__snapshots__/ChallengeMaterial.test.js.snap
@@ -647,6 +647,74 @@ exports[`Curriculum challenge page Should hide mobile modal on click 1`] = `
 </div>
 `;
 
+exports[`Curriculum challenge page Should not render diff if submission was not send 1`] = `
+<div>
+  <div
+    class="row challenge-display mt-3"
+  >
+    <div
+      class="challenge-display_challenges undefined col-md-4"
+    >
+      <div
+        class="card mb-2 challenge-title-card shadow-sm border-0"
+        data-testid="challenge-title"
+      >
+        <div
+          class="card-body d-flex justify-content-between"
+        >
+          <div>
+            0. Greater than 5
+          </div>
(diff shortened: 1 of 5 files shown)


components/__snapshots__/ChallengeMaterial.test.js.snap
```diff
@@ -86,26 +86,12 @@ exports[`Curriculum challenge page Should be able to select another challenge 1`
             <p
               class="selection__title text-muted mx-2"
             >
-              Select submission
+              Loading previous submissions...
```
ggwadera, line 89: Looks like this snapshot is incorrect.

Ulisseus: Fixed.

Ulisseus:
I changed dummy data to avoid shifting dates in snapshots. Ideally dates should be mocked with `jest.useFakeTimers('modern')` but using fake timers will cause problems with breaking cache.

ggwadera:
Why `jest.useFakeTimers`? That's used for testing code with timed functions like `setTimeout` and not dates. You'll probably need to mock `dayjs` to have some controlled way of outputting dates or try to change the tests to not use snapshots.

ggwadera (approved):

Ulisseus:
>You'll probably need to mock dayjs to have some controlled way of outputting dates or try to change the tests to not use snapshots.
I tried it but mocking dayjs is not as easy as it sounds (days js uses another library for mocking in its own tests). 

I updated this pr again, so dummy data uses fixed dates `createdAt: dayjs().subtract(16, 'day').valueOf()`. It's 16 days ago no matter what.

Issue #945, opened 2021-07-08
Update mobile curriculum layout


Ulisseus:
#883 could been fixed in a few different ways. By hiding images and thus creating more space for lesson title and "completed" check or by using completely different layout for mobile devices. 

Initially I choose the former approach, but it looks rather boring on the second glance. I also noticed that progress bar component has too much empty space so it needs to be changed as well.
![Screen Shot 2021-07-08 at 16 51 39](https://user-images.githubusercontent.com/3845012/124980992-b89f3300-e03d-11eb-84fc-f65b7f15d2b9.png)

- [x] refactor lesson card for mobile devices (while keeping lesson images)
- [x] refactor progress bar for mobile devices
- [x] make announcement component the same width as lesson cards (and change it's appearance so it won't be confused with progress bar)
- [x] refactor navigation bar (currently everything except login/logout is hidden)

PR #946, opened 2021-07-08
update #945 - refactor lessonCard component

Ulisseus:
This PR adds new layout for lessonCard component. Depending on your viewport width (small bootsrap breakpoint) either mobile or tabletAndDesktop part is hidden. Desktop and tablet layouts are untouched. 
Screen Shot 2021-07-08 at 16 44 23

Lesson images were recentered. You won't notice the difference but they were off by a few pixels which resulted in straight edges when using `border-radius`. 

Review button got additional prop to make it more customizable (previously it was impossible to change its margins or padding).  

[Vercel](https://c0d3-app-9vrklcotj-c0d3.vercel.app/curriculum). Use responsive mode with small viewport. Log in in to see "completed" check near lesson title.


@types/lessonCard.d.ts
```diff
@@ -14,6 +14,7 @@ export type ReviewButtonProps = {
   isCompleted: boolean
   reviewUrl: string
   lessonId: number
+  style?: string
```
songz, line 17: `style` should be renamed to `additionalClass` or something. `style` is a misnomer.

songz (approved):

@types/lessonCard.d.ts
Ulisseus: I renamed it to className.

PR #949, opened 2021-07-10
Connect app to the c0d3r (Discord Bot)

JasirZaeem:
Part of #843 

As a part of the effort to integrate the discord server with c0d3.com I had created a [Discord bot, c0d3r, here](https://github.com/garageScript/c0d3r) which exposes a [REST Api](https://github.com/garageScript/c0d3r/wiki/C0D3R-Api-Documentation) to interact with Discord.

This PR adds utility functions to the app that allows interaction with this bot, right now this lets us send messages to Discord from the app and more features can be added later as required.

Full integration to achieve the same that we had mattermost can be done once the OAuth integration, under process, is completed.

Until then the following notifications have been added in this PR which the app will send to Discord.

- ## On challenge submission

When a new challenge is submitted the app will call the bot to send a message to the channel of the next lesson to notify reviewers. e.g. to the JS2 channel for submissions in lesson JS1 (submissions for the last lesson will be notified in its channel itself, currently this is JS9)

### Example
image

- ## On lesson completion

When a student completes a lesson two notifications will be sent, one to the completed lesson to notify that the student has completed the lesson, and one to the next lesson's channel to notify that a new student is beginning the lesson, as was the case with mattermost. In case of the last lesson (currently JS9) the second notification will be skipped.

### Example
**1st notification**
image

**2nd notification**
image

- ## On receiving a star
When a reviewer receives a star a notification is sent to the channel for the lesson they received the star for.
### Example
image

### TODO

Once the OAuth integration is complete and the app has knowledge of the user's discord id more ways to communicate with discord through the bot can be implemented. For example challenge submission notifications can mention the discord user with their correct username/server nickname (if set) once we have their id like this
image

## Other changes
- After discussion with @songz , added `discord.js` to be able to use it's types for better type safety.
- c0d3r now immediately responds to requests after a firing a request to discord (for valid requests) by default, an additional parameter `includeDetails` can be passed in json bodies of requests, requests with these will be responded to after the message is sent to discord with the id of the message. [Relevant PR](https://github.com/garageScript/c0d3r/pull/11)
- added `.idea` to `.gitignore` to ignore files created by the jetbrains ides like webstorm.


helpers/discordBot.ts
```diff
@@ -0,0 +1,86 @@
...
+  embed
+) =>
+  fetch(`${messagesEndpoint}/channel/${channelId}`, {
+    method: 'POST',
+    headers,
+    body: JSON.stringify({
+      message,
+      embed
+    })
+  }).then(r => r.json())
```
ggwadera, line 39: Should probably add a `.catch()` on these fetches to handle unexpected errors that could interrupt the API call.

Could also maybe extract a helper function here with the `fetch` call since the parameters are the same for all the functions.

tomrule007: Well there is actually another issue were we need to rememeber we are running serverless functions so unless you await every async call there will be no work finished after the user gets there response. 

But I don't think we really want our users waiting for a discord msg to go through so not awaiting is good in this case.

I think the real proper pattern we need to implement is if we want to catch all the errors is either auto respond with a msg received update and await that or just send it and know all error reporting is responsible via the receiver

ggwadera: > Well there is actually another issue were we need to remember we are running serverless functions so unless you await every async call there will be no work finished after the user gets there response.

Yeah, when solving the issue with emails not being sent I believe part of the problem was that the promise was not being awaited. In this case the await should be put on the function that will be calling this, because the return of these functions are going to be 'Promise' even with the await.

For the error catching, in the case of these notifications I think they should be handled gracefully and not propagate up the call stack. Since these notifications are not crucial for the app they can be ignored and just be logged so us on the engineering team can look into it, but the response for the client should still be successful.

JasirZaeem: My understanding is all these discord functions should be fire-and-forget. They send a command to the bot and that's it, this is to ensure that requests, of which these are called a part of, don't wait on response from discord as that is not necessary.

And so, keeping this in mind, like @tomrule007 pointed out, the serverless invocation that was handling a request may have already ended before discord responds.

Considering this, I'm not sure what a catch function would do here. The intended use is to call these without awaiting them or adding `.then`, if the caller needs the returned response then it would be their responsibility to take care of the error handling by either adding `.catch` then or if using async/await then surrounding it in a try catch block.

Does this make sense? if not then I'd need some suggestion as to what to do inside a `.catch` if a attach it to these.

JasirZaeem: @ggwadera So is your suggestion then to await where they are called, and add a catch to these functions which silently logs the error on failure and lets the rest of the caller function run like usual ?

ggwadera: Yes, having an await here won't make a difference, the function would still be asynchronous.

And because we don't really need the response here, `fetch` will throw an error on response statuses different from 2xx or 3xx. So to avoid having a try/catch everywhere we`re gonna call these functions, we can handle the exceptions here so the caller can function normally.

If an error is thrown here it would interrupt the caller and send an error response to the client. So if we had some more db operations to do after this for example, they would never be executed and we could end up with an invalid state on our database.

tomrule007: It is my understanding as long as the promise propagates it's way back to the graphql server, the server will wait for resolve before sending a response to the user.

So the await would have to be on the 
`await sendSubmissionNotification(...)` in the `helpers/controllers/submissionController.ts  (line 69)` file

Then the catch statement could be anywere before or on this call. I talked with @JasirZaeem and told him I believe `.catch(console.error)` is the safest way to insure that error makes it to our log drain.

And since this response isn't important and we don't want to slow our users down unnecessarily I think we should follow the example from `js5/problem 7 image text extraction challenge` and have the `c0d3r` api always return a [Status 202 Accepted](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) as soon as possible to prevent user slow down. We could also add a timeout to the fetch to reduce slow downs when the server is down.


Either way I am excited to get this discord bot going because I think it will be great for community engagement :)

ggwadera: Hmmm good idea. The bot could reply immediately since we don't to need to wait for the Discord API call to finish it here. After the bot sends the response back it could the process the notification normally.

Detailed error logs could be done on the bot side too, here we could just leave a generic catch to avoid halting the execution.

@types/discordBot.d.ts
```diff
@@ -0,0 +1,70 @@
+import { BotErrorType, IdType } from '../helpers/discordBot'
+
+export type snowflake = string
```
Ulisseus, line 3: I don't understand this `snowflake` type. Why not use string instead?

helpers/discordBot.ts
JasirZaeem: Yeah I am working on that, I will change the default behaviour of the bot to respond immediately without waiting on discord, while providing an option to get details of the message if needed, as these details can be used to perform tasks like replying to these messages, editing them reacting to them and so on.
So the default behaviour of the bot will be to respond without waiting on discord, but if requested it will wait for the message sending to complete and respond with the details of the message, this can then be used in the future to build more features.
Adding logging in the bot itself is a task in progress.

@types/discordBot.d.ts
JasirZaeem: It's just for convenience, to make it clear at a glance that it'll be something like '828783458469675019' and not an arbitrary string.

ggwadera: It's just an type alias. This way we can know that this string type actually has a snowflake id format.

Ulisseus (commented):
TIL about snowflake ids. This wiki link should be in comment IMO.

@types/discordBot.d.ts
Ulisseus: TIL about snowflake ids. This wiki link should be in comment IMO.

helpers/discordBot.ts
JasirZaeem: That's because in most cases you would just fire and forget it, but, it does return a json body with any errors and (optionally) details of the message the was created at discord, so if a function wants to use any of these it can just acquire an object.

@types/discordBot.d.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export interface BotApiError extends BotErrorBase {
+  type: BotErrorType.api_error
+}
+
+export type BotError = BotAuthError | BotValidationError | BotApiError
+
+export type MessageResponse =
+  | {
```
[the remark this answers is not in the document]
JasirZaeem: That was just the linter, Though this works too, I'll change it

@types/discordBot.d.ts
```diff
@@ -0,0 +1,70 @@
...
+  url?: string
+  timestamp?: number
+  color?: any
+  fields?: any[]
+  files?: any[]
+  author?: any
+  thumbnail?: any
+  image?: any
+  video?: any
+  footer?: any
```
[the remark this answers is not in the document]
JasirZaeem: https://github.com/discordjs/discord.js/blob/master/typings/index.d.ts#L3546-L3558 it depends on many other types I'm not sure how deep to go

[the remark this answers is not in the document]
songz: could you add a comment with the link? 

couldn't we import discord.js's types instead of recreating this? Of no, then perhaps instead of `any`, we just type the specific resource that we need. If in the future someone needs more, they can add it in by following your comment

helpers/discordBot.ts
JasirZaeem: - Calls to dicord functions are now awaited.
- Common fetch calls in discord functions have been moved into a separate function
- c0d3r will immediately respond to requests by default after https://github.com/garageScript/c0d3r/pull/11

@types/discordBot.d.ts
JasirZaeem: discord.js has been added as dependency, snowflake type alias is now imported from it. A comment describing it has been added in discordBot.d.ts

@types/discordBot.d.ts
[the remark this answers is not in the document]
JasirZaeem: A third type has been added to the union, linter now only allows
```typescript
export type MessageResponse =
  | { status: 'sending' }
  | { id: snowflake }
  | BotError
  ```

File: @types/discordBot.d.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (JasirZaeem) [reply @types/discordBot.d.ts]: As discussed, `discord.js` has been added as a dependency and relevant types are now imported from it rather than been manually recreated.

**Review by songz — APPROVED** [review]

File: helpers/discordBot.test.js [hunk 5]
```diff
@@ -0,0 +1,122 @@
...
+
+const channelId = '1234'
+const message = 'hello'
+const embed = { description: 'an embed' }
+const lessonId = 1
+const notificationLessonId = 2
+const idType = IdType.C0D3
+const userName = 'test_user'
+const discordUserId = '9898'
+const challengeTitle = 'test challenge'
```
hwong0305, line 36: Can you please make this into an object instead? I think it's neater than having separate variables for everything.

ggwadera (approved):

songz:
@JasirZaeem I assigned this issue to you perhaps you could take a look afterward? Its sort of related to this feature: 
https://github.com/garageScript/c0d3-app/issues/954

helpers/discordBot.test.js
tomrule007: Might not look the neatest but it does make all the code below it very easy to read IMO

package.json
```diff
@@ -41,6 +41,7 @@
     "bootstrap": "^4.6.0",
     "c0d3-diff": "^0.0.1",
     "dayjs": "^1.10.5",
+    "discord.js": "^12.5.3",
```
tomrule007, line 44: I might have missed it but if this is just added for type defs can it be moved to dev dependency ? or does that break our ci?

helpers/controllers/submissionController.ts
```diff
@@ -41,15 +42,39 @@ export const createSubmission = async (
...
     })
+
+    const { challenge, user, lesson } = submission
+
+    // TODO: Add support for discord ids when oauth implementation is complete
+
+    // Get next lesson
+    const nextLesson = await prisma.lesson.findFirst({
+      where: { order: lesson.order + 1 }
+    })
```
tomrule007, line 67: Seems like we could just have used lesson.order in the `sendSubmissionNotification(...)` and then moved this id lookup to the c0d3r side.

Micro optimization but it might be a little easier to use api

package.json
tomrule007, line 44: Could this be a dev dependency if it is only used for types?

tomrule007 (approved):
Looks good, just got the one question about `discord.js` possibly being a dev dependency if it is only used for Types? (I'm still a ts noob so i dont know how that all works)

And lastly I thought you could remove some extra database calls by doing some if the id look ups on the c0d3r side. Not really sure how much time an extra database call roundtrip adds.

.gitignore
```diff
@@ -116,3 +116,6 @@ dist
 
 #removing package-lock as the master branch has yarn.lock
 package-lock.json
+
+# Jetbrains ide config files
+.idea
```
hwong0305, line 121: This is the first time I've seen someone use Jetbrains as an IDE for JS/TS.

hwong0305 (approved):

helpers/discordBot.test.js
JasirZaeem:
```javascript
const mockData = {
  channelId: '1234',
  message: 'hello',
  embed: { description: 'an embed' },
  lessonId: 1,
  notificationLessonId: 2,
  idType: IdType.C0D3,
  userName: 'test_user',
  discordUserId: '9898',
  challengeTitle: 'test challenge'
}
```
Is this fine? variable names will be longer where used but should be readable

package.json
JasirZaeem: Should work in devDependencies too, it was @songz 's suggestion that main dependencies is a better place to put it in to prevent future issues.

helpers/controllers/submissionController.ts
JasirZaeem: `sendLessonChannelMessage` is used here. I feel id is the better property to refer to channel's by, it is more self-explanatory in code and docs to see lessonId than lessonOrder. Are you suggesting to make a different endpoint in c0d3r api for this specific case?
This codepath is only executed when someone completes a lesson so this isn't executing too often, I think for now it is probably better to have this here than abstracting it away inside the bot.

.gitignore
JasirZaeem: I find WebStorm very convenient :D

JasirZaeem:
> @JasirZaeem I assigned this issue to you perhaps you could take a look afterward? Its sort of related to this feature:
> #954

Yeah, sure, I'll work on it.

Ulisseus (approved):

PR #952, opened 2021-07-13
Update the local setup instructions

SlyBouhafs:
The step for installing yarn was missing although it was required for setting up the curriculum repo, so i updated the local setup instructions section to include the step for installing yarn globally.


ggwadera (approved):
Thanks @SlyBouhafs!

PR #955, opened 2021-07-14
update #945 - add hamburger menu on mobile devices

Ulisseus:
Current navBar functionality is rather limited on phones and tablets, you can only log in or sign up. Which makes sense considering that there isn't much space on smaller screens.

This PR adds hamburger menu which makes it possible to have nice mobile layout while keeping full functionality from desktop component.

### AppNav

I switched from using bootsrap to react-boostrap for navBar. Normal bootsrap is using jquery for updating the dom (which is obviously bad for react project), so there is no bootrap js in our bundle and more complex components like navBar don't work. 

* Username is now capitalized.
* Login and Logout buttons have the same width. 
* Admin dropdown uses the same styling as other buttons.
* Small fix to route highlighting. It still doesn't work with admin components but it's a feature at this point.
* Bazillion updated snapshots 

[Vercel](https://c0d3-app-nr3h0rk2s-c0d3.vercel.app).

### DIFF
--- a/components/DropdownMenu.tsx
+++ b/components/DropdownMenu.tsx
@@ -24,6 +24,7 @@ type DropDownMenuProps = {
     | 'warning'
     | 'danger'
     | 'none'
+  bsPrefix?: string
 }
 
 export const DropdownMenu: React.FC<DropDownMenuProps> = ({
@@ -31,7 +32,8 @@ export const DropdownMenu: React.FC<DropDownMenuProps> = ({
   variant = 'none',
   title,
   size,
-  items
+  items,
+  bsPrefix
 }) => {
   const menuItems = items.map((item: Item, itemsIndex: number) =>
     !item ? (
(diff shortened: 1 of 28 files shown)


components/DropdownMenu.tsx
```diff
@@ -23,33 +24,49 @@ type DropDownMenuProps = {
     | 'warning'
     | 'danger'
     | 'none'
+  bsPrefix?: string
```
hwong0305, line 27: what is `bsPrefix`? Can you please give this a better name?

scss/appNav.module.scss
```diff
@@ -7,3 +7,21 @@
...
+  & > * {
+    min-width: 5em;
+    @media (min-width: 992px) {
+      background-color: white !important;
+    }
+  }
+}
+
+.dropdown-item {
+  color: #838383;
```
hwong0305, line 22: This color `#838383` is used really often. Should this be setup in our theme instead?

songz (approved):

components/DropdownMenu.tsx
Ulisseus: I added comment describing how bsPrefix works in bootsrap.

scss/appNav.module.scss
Ulisseus: I created new file colors.module.scss from which we can import default colors. It wasn't possible to import from index (it uses global styles). 

TO-DO. Change all colors in scss files to import from this new file.

Issue #976, opened 2021-07-24
Bug Hydration Mismatches


tomrule007:
## Bug: Hydration Mismatch
  server render does not match client render on first load
![hydrationMismatch](https://user-images.githubusercontent.com/17365077/126851837-9b3fb76d-6866-4fab-b01b-85e516720ac8.png)

**Environment:** local dev machine, current master: f7ef11b67de6b48be943600b7259a0fa3d4726e2

**Reproduce:** first load into `/curriculum`

## Details
NextJs Provides a lot of cool features but it can also produce some sneaky gotchas like this one.

When NextJS builds our pages it runs the page components server side and generates the first html file with the intial nodes already loaded. When a user request a page it sends that prebuilt html page which loads really fast and can be painted on the browser before any javascript has been parsed. (ideally this first view should be as close to the end result as possible). Then once the javascript loads react runs `reactDOM.hydrate()` instead of the normal `reactDOM.render()`.
 
Hydrate is really fast because it doesn't do any node diffs and just expects the trees to be the same as the server while it attaches all the event listeners. So if the trees don't match we can run into issues with nodes being misidentified by react

_\*\* My explination might be partial wrong but I think it's close enough and i've included more reading material on the topic below_

## How to prevent
Don't use any code outside of a useEffect block that may return different results on the server and client IE any code that depends on `window` or `document` will likely return different results as those nodes don't exist server side. Also and issue with graphql queries as those can run server side but can be blocked with the `{ssr: false}` 

**Further reading material**
- https://www.joshwcomeau.com/react/the-perils-of-rehydration/
- https://aboutmonica.com/blog/server-side-rendering-react-hydration-best-practices
- https://www.apollographql.com/blog/apollo-client/next-js/next-js-getting-started/
- https://tech.willhaben.at/how-to-shoot-yourself-in-the-foot-in-react-401e7dbba720

flacial:
Has been fixed in #976

PR #977, opened 2021-07-24
Updates #976 Bugfix/hydration mismatch

tomrule007:
This PR fixes hydration mismatches on the `/curriculum` page

See issue #976 for more details

### DIFF
--- a/__tests__/__snapshots__/storyshots.test.js.snap
+++ b/__tests__/__snapshots__/storyshots.test.js.snap
@@ -14568,11 +14568,6 @@ Array [
   <div
     className="container-md"
   >
-    <img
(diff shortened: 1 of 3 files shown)


ggwadera (approved):

songz (approved):

hwong0305:
This is a very nice find. I was actually listening to Syntax and Scott mentioned how it's super painful to roll your own server-side rendering in React which is why everyone relies on a framework like Next.JS to handle it instead. 

Unfortunately, it makes gotcha like this not very documented but this is a good start. I think this is a good candidate to be added to our wiki :)

hwong0305 (approved):

Issue #950, opened 2021-07-10
Handling 3rd party services with serverless hosting


tomrule007:
# Problem

Serverless hosting setup does not behave exactly like our dev environments and we run into these issues
1) logging system not  reporting errors to sentry.io
2) Then discussing how to handle our new discord bot integration here https://github.com/garageScript/c0d3-app/pull/949#discussion_r667349580 that would also run into the same issues.
3) which also brought up that the email signup issues experienced with the old provider were probably also serverless related

After realizing there are a lot of quirks  I think it is best to handle the issue out in the open with a solid design doc on our best practices moving forward.

The root of the problem is when the lambda receives the `res.send(..)` command responding to the server any unresolved promises get killed (or really frozen? still don't understand all the lambda details)  like fetchs or database writes. If something must be done we have to await it before sending the response. We also only have 10 seconds before the lambda times out

# Goal 

With c0d3-app being a nextjs app hosted on vercel our express backend gets converted into serverless lambda functions and I think it will be best if we have a central understanding of how that effects the application. Ideally we can come up with a set of best practices and use the same pattern throughout the codebase with this design doc moved to the wiki when finalized.

# Architecture
TODO: Add diagram of basic lambda spin up -> loop(freeze -> thaw) -> kill process
# Research

My initial thought was to make sure 3rd party services always return a status 202, like js5/problem 7 but the reality is that still adds extra round trip time and we are not incontrol of all 3rd party apis.

I just found this [blog post](https://www.sensedeep.com/blog/posts/stories/lambda-fast-http.html) that goes into great detail of how lambdas work and suggests using `res.end(null,()=> action to wait on)` pattern or something like that still not quite sure on how that callback works or if it will be easy to apply that pattern when our graphql server is handling the response.

Lets start the discussion and get on the same page here :)

Ulisseus:
I'd add [this](https://github.com/garageScript/c0d3-app/pull/681) to issues with serverless. I resorted to reverting my pr and using local storage check because I couldn't make it work with sequelize (considering that someone wrote a very lengthy guide about setting up sequelize for aws I wasn't the only one). 
And while we have migrated from sequelize (yay!) the same problem could happen with prisma. One container uses all connections, aws freezes it and thaws another container but max limit for connections is already exhausted.

ggwadera:
Not really related to this issue, but this blog post goes over some technical details on how AWS Lambdas works, saw it on HN today. Might get something useful out of it though.

[...]

PR #985, opened 2021-07-30
Update/sentry

tomrule007:
- Bumps sentry from 6.8.0 -> 6.10.0
- Removed the temporary work around code that was used to manually init and wait for flush #926 
- update dependant tests
  - apolloLogPlugin no longer called sentry flush (sentry package handles this now)
  - lessons api needed a non null `req` object as it is wrapped by sentry that expects a req to exist

**Extra note**
There solution to get sentry flush working on lambda serverless functions is very clever. Since the lambda functions call `res.end()` to finish the transaction and shut down, they just make a copy of that function and then overwrite it with their own. The new function waits until there tasks are finished before calling the real `res.end()` themselves. https://github.com/getsentry/sentry-javascript/pull/3811/commits/62189efca1e3fe371ae158acd4d8e6b7bf99d59a

As I documented here, #950, this server shutting down before everything finishes is a real problem and we could probably use there fix to make a utility function that waits for all provided promises to resolve before calling end


ggwadera (approved):
Awesome, I'm glad that this bug didn't take too long for them to fix.

songz (approved):

hwong0305 (approved):

PR #988, opened 2021-08-01
[Migration] Change `isPassed` column to `passedAt`

### LINKED ISSUE #545: Fix Sequelize model and database types

ggwadera:
Finally closes #545

- Changed the `isPassed` column in the User Lessons table to `passedAt` and with a Date type
- Removed unused `isTeaching` and `isEnrolled` columns
- Added migration to do the above, while also converting existing text values into timestamps
- Updated GraphQL schema
- Updated pages, components, resolvers, etc, to use the new field name
- Removed some unused type declaration files from `@types`

**P.S.:** preview deployment will not work correctly because the DB migration only runs during a production deployment.


__dummy__/lessonData.ts
```diff
@@ -854,7 +854,7 @@ const dummyLessonsData: Lesson[] = [
     id: 27,
     title: 'General Algorithms',
     description: 'General Algorithm from interviews',
-    docUrl: '',
+    docUrl: null,
```
ggwadera, line 857: Changed this to get 100% branch coverage due to changes in the curriculum page.

tomrule007 (commented):
Great change! Really helps new devs to not have any dead code or data laying around

songz (approved):

Issue #670, opened 2021-04-14
Write integration tests


ggwadera:
Since a lot of production breaking bugs have been happening recently, writing integration tests to assure full API/client functionality can help in reduce the number of these bugs.

Basic implementation steps:

1. Start Server
2. Send API request
3. Save response as a snapshot

This will probably be better implemented using a in-memory embedded database like SQLite to avoid connecting to an external database, the recent Prisma seed PR #659  can help in seeding this db with fake data for the tests.

tomrule007:
I'm glad to see progress is getting made on this issue. Last week I had mentioned how switching to the `getLayout` pattern interfered with our current tests and I thought [next-page-tester](https://www.npmjs.com/package/next-page-tester) might be a good solution. After reading this little blog post https://dev.to/toomuchdesign/dom-testing-next-js-applications-46ke  it sounds like the real solution is to switch to intergration tests like this and possible impliment some e2e tests as well.

ggwadera:
Yeah some e2e tests running the entire Next.js stack might be good too, so we can test the SSG and SSR rendering with Next's data fetching and also the API layer. But for the front-end components I believe most cases can be covered by the unit tests.

With E2E tests we could test an entire flow of our app's logic, for example: account creation -> login -> send submissions -> receive review -> complete lesson. That might help in detecting some regressions like we already had before.

songz:
Sample PR for reference: https://github.com/garageScript/c0d3-app/pull/992

PR #992, opened 2021-08-02
[PoC] Integration tests with Apollo and PostgreSQL database

ggwadera:
This PR is a proof-of-concept for a possible way that we could do integration tests with Apollo Server and Prisma, connected with a live PostgreSQL database instance through Docker, as part of #670.

Included are 2 example tests files:
- `__tests__/integration/lessons.ts` - simple query for fetching the lessons and comparing to a snapshot
- `__tests__/integration/signup.ts` - more complex example that goes through the entire user creation process, verifying that the user was actually created on the database

References:
- https://www.apollographql.com/docs/apollo-server/testing/testing/
- https://www.prisma.io/docs/guides/testing/integration-testing#integration-testing


helpers/apolloClient.test.js
```diff
@@ -17,7 +17,6 @@ jest.mock('@apollo/client', () => ({
     cache: { restore: jest.fn() }
   }))
 }))
-jest.mock('nodemailer')
```
hwong0305, line 20: I don't think we should remove mocking mail for unit tests. 

If we want to test the mail, we should only do it explicitly in the integration tests and e2e tests.

helpers/controllers/authController.test.js
```diff
@@ -1,5 +1,4 @@
 jest.mock('bcrypt')
-jest.mock('nodemailer')
```
hwong0305, line 2: same as above

helpers/controllers/passwordController.test.js
```diff
@@ -4,7 +4,6 @@
 
 jest.mock('bcrypt')
 jest.mock('nanoid')
-jest.mock('nodemailer')
```
hwong0305, line 7: same as above

helpers/mail.test.js
```diff
@@ -6,12 +6,6 @@ import { getResetTemplate, getSignupTemplate } from './mailTemplate'
 import { sendResetEmail, sendSignupEmail } from './mail'
 import nodemailer from 'nodemailer'
 
-jest.mock('nodemailer', () => ({
-  createTransport: jest.fn().mockReturnValue({
-    sendMail: jest.fn().mockReturnValue((mailoptions, callback) => {})
-  })
-}))
-
```
hwong0305, line 14: same as above

helpers/apolloClient.test.js
ggwadera: I didn't remove nodemailer mocking, I made a manual mock under `__mocks__` that is automatically applied.

hwong0305 (approved):

songz:
Closing because nobody has picked this up. Will be picked up with this issue: https://github.com/garageScript/c0d3-app/issues/670

Issue #1010, opened 2021-08-05
Enhanced MdInput


tomrule007:
# Enhanced MdInput

Fix issue with input box size not saving size (Issue #869) and add additional features using Github's comment box as a target spec.

## Complete 
![autoresize-demo](https://user-images.githubusercontent.com/17365077/128428111-8720c746-efc0-41fd-a791-567b71b0078a.gif)

- [x] min-height 100px / max-height 500px
- [x] Autosize to fit content
- [x] Disables autosize after user resize event
- [x] Saves all state when switching back and forth to preview mode
  - [x] size
  - [x] scroll location
  - [x] focus and selected text
  - [x] preserve undo old text (ctrl+z)

## TODO
 
New Markdown buttons that insert/delete markdown styles around selected text
(with hotkeys that only work while textbox is focused)

- [ ] Header `### <text>`
- [ ] **bold** (hotkey: ctrl+b) `**<text>**`
- [ ] _italic_ (hotkey: ctrl+i) `_<text>_`

- [ ] quote (hotkey: ctrl+b) `\n\n> <text>\n\n`
- [ ] code (hotkey: ctrl+e) `**<text>**`
- [ ] link (hotkey: ctrl+k) `[<text>](url)` \*Updates selected text to url

- [ ] bullet list `- <text>`
- [ ] ordered list `1. <text>`
- [ ] task list `\n\n- [ ] <text>\n\n`
- [ ] Auto continue lists on return

- [ ] Make responsive - move first six buttons to drop down menu on small screens

Other extras

- [ ] prop to enable local store persist

ggwadera:
Great to have someone working on this issue. The reviewing experience could a lot more pleasant by having a little more smarter editor.

> New Markdown buttons that insert/delete markdown styles around selected text
(with hotkeys that only work while textbox is focused)

For this one, besides hotkeys I think that adding surrounding parenthesis, brackets, quotes, etc, when you have some text select could also be a good feature. Similar to what VS Code or any other code editor does.

And for a bit more difficult task, do you think that it's possible to have live syntax highlighting like Discord has when you're editing a code block for a message? I think that's an awesome feature that they have, very cool haha.

tomrule007:
Ya I orginally had the bracket closing feature on my list! I removed it because I thought it might be a bit tricky as the buttons will just call the on change handler with my new changes but for the closing brackets I might want to listen to the key down event, I'll figure it out.

[...]

PR #1013, opened 2021-08-08
WIP Feature/markdown toolbar

tomrule007:
WIP Adding markdown toolbar to MdInput, See issue  #1010


JasirZaeem:
Looks great! We already have `react-feather`/feathericons.com as a dependency, maybe you can use those instead of the GitHub icons if they have all the ones you need?

tomrule007:
> Looks great! We already have `react-feather`/feathericons.com as a dependency, maybe you can use those instead of the GitHub icons if they have all the ones you need?

Both packages are tree shaken single svg imports so have both really isnt adding any extra weight to the project physically but does at extra mental burden to the developers so I get not wanting to have both.


Feather is missing: Header icon, quote icon and numbered list icon.


We are currently only using: Star, CheckCircle, File, Code, GitHub, Linkedin icons from the feather package right now. The GitHub packages has all of those besides the Linkedin icon so we could potential remove feather and just use GitHub icons.

JasirZaeem:
Yeah bundle size isn't a problem with this, I suggested a single package mostly for a cohesive design, mixing different icon libraries might not look good, so I think a single style of icons across the board when possible would be good regardless of the actual icon library used.

PR #1039, opened 2021-08-17
[Migration] Add 'slug' field to Lesson

tomrule007:
## Changes
- Add `slug` field to Lesson database table
- update admin lesson creating / update components to handle new slug field
  - Add error message in response to submit errors (used to catch is unique error)
slug-field-with-error
(Screenshot of http://localhost:3000/admin/lessons)
## Purpose
Allow us to easily link self hosted lesson material to correct lesson via [folders names](https://github.com/garageScript/c0d3-app/tree/master/pages/curriculum/lessons)

## Test locally 

Since this pr includes a migration the vercel preview branch will not work as our preview branches currently run on the production db.

You can test it locally by pulling down the branch and then switching to it and running `yarn prima migrate dev` \* When you switch back to another branch you will have to run `yarn db:init` to reset your dev db

## Potential Migration issues

- Duplicate Lesson.orders will cause the migration to fail as it uses `js` + `Lesson.order` as the default slug value which must be unique, but order isn't required to be unique

-  ???? Please point out any other potential issues if you see them

### DIFF
--- a/components/FormCard.tsx
+++ b/components/FormCard.tsx
@@ -26,6 +26,7 @@ type Btn = {
 type FormCardProps = {
   values: Option[]
   onSubmit: Btn
+  submitError?: string
   capitalizeTitle?: boolean
   onChange: Function
   title?: string
@@ -95,6 +96,7 @@ const OptionInfo: React.FC<OptionInfoProps> = ({
 export const FormCard: React.FC<FormCardProps> = ({
   values,
   onSubmit,
+  submitError,
   capitalizeTitle = true,
   title,
   onChange,
@@ -121,6 +123,7 @@ export const FormCard: React.FC<FormCardProps> = ({
           )}
         </div>
         <div className="text-left">{optionsList}</div>
+        {submitError && <h6 className="text-danger">{submitError}</h6>}
         <div className="text-center mb-4">
           <Button onClick={btnOnClick} type="primary" color="white">
             {onSubmit.title}
--- a/components/admin/lessons/AdminLessonInfo.tsx
+++ b/components/admin/lessons/AdminLessonInfo.tsx
@@ -90,6 +90,7 @@ const newLessonAttributes = {
   githubUrl: '',
   videoUrl: '',
   order: '',
+  slug: '',
   chatUrl: ''
 }
 
--- a/components/admin/lessons/AdminLessonInfo.tsx
+++ b/components/admin/lessons/AdminLessonInfo.tsx
@@ -32,7 +32,7 @@ type NewLessonProps = {
 
 // Creates card for a lessons's information to update
 const EditLesson: React.FC<EditLessonProps> = ({ setLessons, lesson }) => {
-  const [alterLesson, { loading, data }] = useMutation(updateLesson)
+  const [alterLesson, { loading, data, error }] = useMutation(updateLesson)
   const [lessonProperties, setLessonProperties] = useState(
     getPropertyArr(lesson, ['challenges', '__typename'])
   )
@@ -75,6 +75,7 @@ const EditLesson: React.FC<EditLessonProps> = ({ setLessons, lesson }) => {
         <FormCard
           onChange={handleChange}
           values={lessonProperties}
+          submitError={error?.message}
           onSubmit={{ title: 'Update Lesson', onClick: alter }}
           title={lesson && lesson.title + ''}
         />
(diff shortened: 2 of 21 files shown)


prisma/migrations/20210816200009_lesson_slug/migration.sql
```diff
@@ -0,0 +1,12 @@
...
+
+-- Add slug's to existing lessons
+UPDATE "lessons"
+SET "slug" = 'js' || "lessons".order;
+
+-- Make slug field required
+ALTER TABLE "lessons" ALTER COLUMN "slug" SET NOT NULL;
+
+-- CreateIndex (ensure no deplicate slugs)
+CREATE UNIQUE INDEX "lessons.slug_unique" ON "lessons"("slug");
```
songz, line 12: nice! Index also makes querying by slug fast

songz (approved):

components/admin/lessons/AdminLessonInfo.tsx
```diff
@@ -32,7 +32,7 @@ type NewLessonProps = {
 
 // Creates card for a lessons's information to update
 const EditLesson: React.FC<EditLessonProps> = ({ setLessons, lesson }) => {
-  const [alterLesson, { loading, data }] = useMutation(updateLesson)
+  const [alterLesson, { loading, data, error }] = useMutation(updateLesson)
```
ggwadera, line 35: Could take this opportunity to also refactor and use the auto-generated typed mutation.

```jsx
import { useUpdateLessonMutation } from '../../../graphql/index'

const [alterLesson, { loading, data, error }] = useUpdateLessonMutation()
```

tomrule007: I'm running into a type issue when implimenting this refactor.
```
Argument of type 'Maybe<Maybe<{ __typename?: "Lesson" | undefined; id: number; docUrl?: Maybe<string> | undefined; githubUrl?: Maybe<string> | undefined; videoUrl?: Maybe<string> | undefined; ... 5 more ...; challenges: { ...; }[]; }>[]> | undefined' is not assignable to parameter of type 'SetStateAction<Lesson[] | null>'.
  Type 'undefined' is not assignable to type 'SetStateAction<Lesson[] | null>'.ts(2345)
```

I can solve it with casting the value but that seems like the wrong thing to do here 
```
    !loading && data && setLessons(data.updateLesson as Lesson[])
```

ggwadera: Just another victim of mistaken type on the graphql schema. Change the response type for this mutation to return `[Lesson!]!` on `graphql/typeDefs.ts`

```graphql
updateLesson(
 # Mutation args...
): [Lesson!]!
```

components/FormCard.tsx
```diff
@@ -121,6 +123,7 @@ export const FormCard: React.FC<FormCardProps> = ({
           )}
         </div>
         <div className="text-left">{optionsList}</div>
+        {submitError && <h6 className="text-danger">{submitError}</h6>}
```
hwong0305, line 126: For accessibility reason, it's best not to skip headings. If you need to use the `h6` styling. You could just use the class. 

https://webaim.org/techniques/semanticstructure/#contentstructure

components/admin/lessons/AdminLessonInfo.tsx
hwong0305: Maybe have a backlog issue where someone reviews all the types and make sure all the non-nullable values are set so it doesn't clash with typescript.

components/FormCard.tsx
tomrule007: That is good to know but I would like to open a separate issue for that as it looks like there are alot of accessibliy issues on the page that should all get fixed together.

Current headers are 
- no `h1` header
- `h2` is the form lesson title 
- `h5` as the form labels (which are not linked to the inputs at all) 
- `h6` are error messages.

Biggest accessibility issue is gonna be the fact that the form labels are not labels or linked to the inputs in any way

Then the lesson selector nav bar is not keyboard accessible at all either

components/admin/lessons/AdminLessonInfo.tsx
tomrule007: It also seems like we have the graphql codegen generating `typescript-resolvers` types but then we are not actually using them. Maybe if we wire those to the resolvers we will discover all the type mismatches we have.

tomrule007: Updated the typedefs and swapped in the new codegen mutations d4a97de

ggwadera: > It also seems like we have the graphql codegen generating typescript-resolvers types but then we are not actually using them. Maybe if we wire those to the resolvers we will discover all the type mismatches we have.

That's true, I've added some of those return types in the resolvers I worked with in the past. But there's some issues that may happen and it's not about the code being wrong, for example, TS complains about a field with a `Date` type not being a `string` which is expected on the return type, but that conversion is handled automatically by the GraphQL server.

ggwadera (approved):

PR #1041, opened 2021-08-19
Feature/lesson page use slug

tomrule007:
Wire `curriculum/[lesson]` & `review/[lesson]` to use the new slug field added in pr #1039

Preview Links: 
https://c0d3-app-n7gycibgm-c0d3.vercel.app/review/js0
https://c0d3-app-n7gycibgm-c0d3.vercel.app/curriculum/js0

Other issues included in this pr:
- Fixed issues with the `LessonCard` review count button
  - Was set to `network-only` fetch so it didn't show stale data but was causing multiple hits to the graphql api when component would rerender, now only hits once on first mount. <br> _\*You can view this bug in production by navigating to https://c0d3.com with your network tab open. if you are logged in it will redirect you to the curriculum page before the getApp fetch has finished and then start the review fetches with will then rerun when the initial fetch finishes (you may have to clear your cache to see it, and easiest to see when you filter graphql requests only)_
  - Converted to next/link to prevent the need to refetch the entire app when clicking the review button.
- Refactored `LessonTitleCard` test file to only be concerned about the component and not the pages it's used in

### DIFF
--- a/__tests__/pages/curriculum/[lesson].test.js
+++ b/__tests__/pages/curriculum/[lesson].test.js
@@ -76,7 +76,7 @@ const session = {
 }
 describe('Lesson Page', () => {
   const { query } = useRouter()
-  query['lesson'] = 2
+  query['lesson'] = 'js1'
   test('Should render correctly with valid lesson route', async () => {
     const mocks = [
       {
@@ -105,7 +105,7 @@ describe('Lesson Page', () => {
   })
 
   test('Should render correctly with invalid lesson route', async () => {
-    query['lesson'] = 100
+    query['lesson'] = 'js100'
     const mocks = [
       {
         request: { query: GET_APP },
@@ -129,7 +129,7 @@ describe('Lesson Page', () => {
     await waitFor(() => expect(container).toMatchSnapshot())
   })
   test("Should correctly render challenges page for students who hadn't passed previous lessons", async () => {
-    query['lesson'] = 25
+    query['lesson'] = 'js8'
     const mocks = [
       {
         request: { query: GET_APP },
@@ -179,7 +179,7 @@ describe('Lesson Page', () => {
     await waitFor(() => expect(container).toMatchSnapshot())
   })
   test('Should render with nulled submissions', async () => {
-    query['lesson'] = 2
+    query['lesson'] = 'js1'
     const mocks = [
       {
         request: { query: GET_APP },
--- a/components/FormCard.test.js
+++ b/components/FormCard.test.js
@@ -88,6 +88,19 @@ describe('FormCard Component', () => {
     expect(container).toMatchSnapshot()
   })
 
+  test('Should display submit error message', () => {
+    const mockError = 'Helpful Error Message Here :)'
+    const { queryByText } = render(
+      <FormCard
+        onChange={() => {}}
+        values={[{ title: '' }]}
+        submitError={mockError}
+        onSubmit={mockBtn}
+      />
+    )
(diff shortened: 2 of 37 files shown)


__tests__/pages/curriculum/[lesson].test.js
```diff
@@ -179,7 +179,7 @@ describe('Lesson Page', () => {
     await waitFor(() => expect(container).toMatchSnapshot())
   })
   test('Should render with nulled submissions', async () => {
-    query['lesson'] = 2
+    query['lesson'] = 'js1'
```
Comment (hwong0305) [inline __tests__/pages/curriculum/[lesson].test.js:182]: Do we really need to have `js0` and `js1` for tests? It seems a bit unnecessary.

components/FormCard.test.js
```diff
@@ -88,6 +88,19 @@ describe('FormCard Component', () => {
...
+    const mockError = 'Helpful Error Message Here :)'
+    const { queryByText } = render(
+      <FormCard
+        onChange={() => {}}
+        values={[{ title: '' }]}
+        submitError={mockError}
+        onSubmit={mockBtn}
+      />
+    )
+    expect(queryByText(mockError)).toBeTruthy()
```
hwong0305, line 101: FYI, you could just check so it's not null. That would be more explicit.

__tests__/pages/curriculum/[lesson].test.js
Comment (tomrule007) [reply __tests__/pages/curriculum/[lesson].test.js:182]: The current way the pages are setup it fetches all the lessons on initial load and then determines which lesson to render based on the the query parameter. So we do have to pass in the slug into our mocked next/router.

once this change goes through I would like to refactor both the review page and lesson page to be static generated and  then at that point we could possibly no longer need to pass the slug in

components/FormCard.test.js
ggwadera: Or use `jest-dom` matchers

```js
expect(queryByText(mockError)).toBeInTheDocument()
// OR
expect(queryByText(mockError)).toBeVisible()
```

tomrule007: Added jest-dom and switched to `toBeVisible()` matcher

ggwadera:
> Fixed issues with the LessonCard review count button
Was set to network-only fetch so it didn't show stale data but was causing multiple hits to the graphql api when component would rerender, now only hits once on first mount. 

About this point, on my end it's still calling the API again if I navigate to some other page and then go back to the curriculum page (without full reload, Firefox browser, Ubuntu 20.04).

Checked through the network tab on the developer tools

1. First Load 
image

2. Entered a lesson challenges page and clicked the go back link 
image

tomrule007:
> on my end it's still calling the API again if I navigate to some other page

I believe that is the intended function we wanted. This allows for the user to manually refresh the page and always get the newest submissions for review.

The bug I fixed was the same review fetch getting triggered multiple times on one navigation.

Here i just visited https://c0d3.com (already logged into my account which has the 7 completed lessons)
submission-triple-fetching

I didn't highlight all the refetching but you can see the one lesson got hit 3 times. There should only be a total of 9 fetches (7 reviews + getApp + getSession) but I 19.

We could go further and switch it to a polling type behavior but I would make a seperate pr for a change like that.

ggwadera:
Oh I see what you meant now, that's cool, nice work.

But yeah having it refresh is the intended way to prevent it showing stale submissions count after submitting a review.

ggwadera (approved):

ggwadera:
@all-contributors please add @tomrule007 for code and doc

PR #1067, opened 2021-08-24
Feature/self hosted lessons

tomrule007:
# Self Hosted Lessons

A large refactor of the full page lesson content added from this issue https://github.com/garageScript/c0d3-app/issues/555 and related pr's

Please comment on anything that you find hard to follow, confusing or plain bad practice. 

## Changes
- Create Static helper utility files (helpers/static/...) 
   - docs.ts & lessons.ts - used to get slug names and locate files associated with those slugs (needed during build process by `getStaticPaths` & `getStaticProps`
   - parseMDX - converts file buffer into serializable mdx source and frontmatter
- Remove layout wrapper from mdx files, which:
  - Makes the code more maintainable by not needing to maintain consistency between to separate layouts (ie challenge page , review page & lesson page can now all share the same layout)
  - Allows changes to Layout not requiring us to manually update each mdx file (also makes mdx files more friendly as they are much closer to md files with no import export statements) 
  - Always us to implement the static layout pattern which will let users only reload the content section of the page when switching between challenges, review and lessons <br>_\*Note so far this PR hasn't changed the challenge or review page to use the shared `LessonLayout` component but solves all the roadblocks on making those changes_ 
- Refactor 'mdx/LessonLayout` moving to component folder and extracting sub components to separate files. This change allows us to use the LessonLayout as a more general layout and keep only the features shared between all the lesson / review / challenge pages in one place.
- Move docs and lesson mdx pages to separate `content` folder
   - This change allows us to use the statically generated template pages (ie `[docSlug]` & `[subLessonSlug]` ) and being able to update one file when we want to change the look of those pages.
   - Also gives us the ability to potential move the content to a separate repo or cms
- Create [docSlug] page - used to display all our full page mdx docs (ie '/learn' page) w/ 5 minute ISR
- Create [subLessonSlug] page - used to display all lesson material that is currently hosted on notion.so also w/ 5 minute ISR

## Remaining work

- [ ] Remove old MDX provider from `_app.tsx` page (should speed up all non-mdx page loads) 
- [ ] Make sure I didn't miss any [ISR](https://nextjs.org/docs/basic-features/data-fetching#incremental-static-regeneration) edge cases (ie admin updates / creates a lesson what could happen?)
    - I believe right now we will just get stale data for 5 minutes and new lessons will not work without `fallback: true`
- [ ] Determine if static helpers should be Asynchronous. Currently the helper functions are using synchronous file system access functions because they run at build time and i believe it can be faster to batch the data access from the file system but I'm not positive how it behaves with ISR enabled  

Other than the remaining work listed above this pr is fully tested and typechecked and should be ready to go. In its current form it will require an admin to manually update the `lessonUrl` on the admin/lesson page to get the lesson links to work. (a future version could automatically detect static files and use that if a lesson url is not set)


JasirZaeem:
If the static helper can run in response to user requests because of isr, I think you shouldn't use sync fs functions, as they will block the main thread. It would likely be better to use fs.promises and then await them so that they aren't blocking other functions that may be running concurrently.

tomrule007:
> If the static helpers can run in response to user requests because of isr, I think you shouldn't use sync fs functions, as they will block the main thread. It would likely be better to use fs.promises and then await them so that they aren't blocking other functions that may be running concurrently.

I will convert them but I'm hoping I get a response on the nextjs discussion I created to find out how it really works. I know I read somewhere about next running getStaticProps and getStaticPaths on seperate worker threads (which is why you cannt  share data between those two operations) and in theory should not slow down anything on the main thread.

But if we were to go back and convert the page to `fallback: blocking` then you are right the user would potential see the slow down. Right now all the pages get built at build time and it always instantly servers the last built page and then rebuilds in the background for the next user to hit the page

PR #1085, opened 2021-08-27
Changes tests comparing paths to be cross platform Closes #1076

### LINKED ISSUE #1076: Tests comparing paths failing on windows
New tests added in #1067 depend on comparing paths stored in the tests and returned by node's fs functions. Tests depending on this are failing on windows because node on windows uses `\\` as path separators as opposed to `/` on posix.
Since the tests were created with posix separators in expected paths that are stored in tests, these tests are failing on windows. e.g.

image

This can be fixed by using something other than a simple string comparison to check these paths. Methods like `path.relative` or  `path.parse()`  and then comparing the resultant objects  can be used at places.

If there are tests where these cannot be used, then `path.sep` can be used to get the current platform's path separator and do some string manipulation before doing a string comparision, like `pathToCompare.split(path.sep).join(path.posix.sep)`

Some methods could also be potentially left untested manullly, like this which is wrapper over a node library function as pointed out by @ggwadera 
https://github.com/garageScript/c0d3-app/blob/da7b7dd3ae6ea7ce6efc400f7eea3b6b62f43bda/helpers/static/docs.ts#L19-L20

https://github.com/garageScript/c0d3-app/blob/da7b7dd3ae6ea7ce6efc400f7eea3b6b62f43bda/helpers/static/docs.test.js#L45-L50

JasirZaeem:
Closes #1076

As described in 1076, tests comparing file paths using simple string equality/matching don't work across platforms as different platforms can use different characters as file separators. Some of these tests were fixed in #1078 

https://github.com/garageScript/c0d3-app/pull/1078#issuecomment-907332905 these two checks are fixed in this PR.

Instead of comparing with `filePath`, `path.join(process.cwd(), filePath)` is used to evaluate the correct path for the platform and then compared.

### DIFF
--- a/helpers/static/docs.test.js
+++ b/helpers/static/docs.test.js
@@ -1,5 +1,7 @@
 import { promises as fs } from 'fs'
 import { getDocSlugs, getDocGithubFilePath, getDocContent } from './docs'
+import path from 'path'
+
 jest.mock('fs', () => {
   return {
     promises: {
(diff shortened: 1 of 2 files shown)


tomrule007 (approved):
Nice update to the tests, keeping them useful and corect.

PR #1092, opened 2021-08-29
bugfix/isr-sublesson-rebuilds

tomrule007:
# Description

Fixes this bug that was breaking ISR rebuilds on all our self hosted lesson pages added in PR #1067
>  {
        "level": "error",
        "message": "[Error: ENOENT: no such file or directory, scandir '/var/task/content/lessons/js5/sublesson'] {\n  errno: -2,\n  syscall: 'scandir',\n  path: '/var/task/content/lessons/js5/sublesson'\n}",
        "timestamp": "2021-08-28T23:58:26.070Z"
      }

\* This error has not been visible to our users because the server has no issue on the initial build and just servers the stale page to the user & we have yet to make these lesson links live so no one really knows about them yet.

## Cause 

Next.js by default strips down the repo to the bare minimum to run all the server side functions in a serverless environment. That means the mdx files that are needed to rebuild the `[subLessonSlug]` page in the `getStaticProps` function call were not present during the rebuilds. Interestingly they are included on the initial build and will show no error at build time.

## Solution

The solution is just use a webpack file copy plugin to make sure the mdx content folders get included in the server build and are available to the lambda functions. This was the popular solution taken from https://github.com/vercel/next.js/issues/8251#issuecomment-8541487184

\* If this `vercel.json` config file ever gets fixed we could use this as an alternative solution [how-can-i-use-files-in-serverless-functions](https://vercel.com/support/articles/how-can-i-use-files-in-serverless-functions) but currently it links to the issue I ended up following to find the above solution for SSR 

## Test Preview Links

https://c0d3-app-8w5asn4km-c0d3.vercel.app/curriculum/js0/functions_and_execution_context
https://c0d3-app-8w5asn4km-c0d3.vercel.app/curriculum/js1/hypertext_markup_language
https://c0d3-app-8w5asn4km-c0d3.vercel.app/curriculum/js2/basic_array_functions
https://c0d3-app-8w5asn4km-c0d3.vercel.app/curriculum/js3/preflight
https://c0d3-app-8w5asn4km-c0d3.vercel.app/curriculum/js4/css
https://c0d3-app-8w5asn4km-c0d3.vercel.app/curriculum/js5/server

### DIFF
--- a/helpers/static/lessons.ts
+++ b/helpers/static/lessons.ts
@@ -3,7 +3,16 @@ import { MDXRemoteSerializeResult } from 'next-mdx-remote'
 import path from 'path'
 
 const LESSONS_GITHUB_PATH = 'content/lessons'
-const LESSONS_PATH = path.join(process.cwd(), LESSONS_GITHUB_PATH)
+
+// This is the solution to make the content files available
+// on the lambda serverless functions that handle the ISR rebuilds.
+// The webpack copy plugin found in the next.config.js file copies the
+// content folder to the .next/Server/chunks directory
+const LESSONS_PATH =
+  process.env.NODE_ENV === 'production'
+    ? /* istanbul ignore next */
+      path.join(process.cwd(), '.next/server/chunks', LESSONS_GITHUB_PATH)
+    : path.join(process.cwd(), LESSONS_GITHUB_PATH)
 
 export type SubLesson = {
   frontMatter: {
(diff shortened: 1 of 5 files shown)


tomrule007:
Been following this issue: https://github.com/vercel/next.js/issues/24700 hoping it would get resolved on next.js end before I did a rewrite

ggwadera (approved):

songz (approved):

PR #1112, opened 2021-09-08
Update promises.mdx

mino323:
# Problem
`node-fetch v3` has dropped support for commonjs. As a result, `require('node-fetch')` no longer works, causing our sample code to give an error.

# Solution
This PR will give a quick update to students about how they could fix the problem by changing `package.json` type to allow them to use `import` instead of `require`.

### DIFF
--- a/content/lessons/js3/sublesson/promises.mdx
+++ b/content/lessons/js3/sublesson/promises.mdx
@@ -285,6 +285,25 @@ module.exports = [1, 2, 3] // module.exports is an array: [1,2,3]
 
 </Spoiler>
 
+<br>
+
+>Depending on which version of nodeJS you have installed you may have to use `import` instead of `require`.
+Using `import`  may need you to add `"type" : "module"`  into  the package.json file in your directory. `{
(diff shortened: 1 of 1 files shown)


tomrule007:
### About the PR
Thanks for creating the PR to improve our content! 

We still don't have a formal template in place for best practices on creating PRs but generally we need a little more context so we can understand what the PR is about and why the improvement is needed.

I saw there was some chat in the discord server about this and how `node-fetch v3` has dropped support for commonjs  but not everyone has seen that.

### About the content update

The current update says it depends on the node version but it isn't really a problem with the node version but a problem with the packages no longer supporting commonJS

I think we need to reword this update to just be a small lightbulb box that says something like:

> Some modules are now dropping support for the commonJS, require/module syntax, and forcing you to use the newer ESM, import/export syntax. Read more about [ESM-only modules](<some good link>)

mino323:
https://github.com/garageScript/c0d3-app/pull/1215 replaces this pr.

PR #1114, opened 2021-09-09
Bugfix/codepen mdx errors

tomrule007:
# Description

When testing the self hosted lessons I was seeing `internal server error` when I initially loaded into [js4/css](https://c0d3-app-8w5asn4km-c0d3.vercel.app/curriculum/js4/css) lesson page and then clicked the link to the part 1 or 3 lessons and was also seeing this message in the console.log:

> DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.

And our sentry.io log picked up some nice [additional details](https://sentry.io/share/issue/3c0cc8833ff745aeacd827e2ae44a2b2/) like the route it was happening on:
> navigation
> {
> from: /curriculum/js4/css, 
> to: /curriculum/js4/classes
> }

## The Problem

I was able to quickly guess it was connected to the embedded `codepens` since the only route the error was appearing on was the `css` page which is the only lesson with codepens.

We were using the recommended HTML embedding style that relies on a script tag to manipulate the dom. React wasn't liking some 3rd party tool stepping on its toes though and was balking about it when it tried to remove nodes that the script had already removed.

## Solution

Convert from embedding the script tag to using the `iframe` element embed option. To make it a little more reusable I created a separate `<CodePen />` component that is hardcoded to the c0d3pen user account and will allow us to make our lessons slightly more maintainable. 

I was tempted to reach for an already built off the shelf component from this [mdx-embedded package](https://github.com/PaulieScanlon/mdx-embed/blob/main/packages/mdx-embed/src/components/codepen/codepen.tsx) but in the end decided to roll my own for a little more flexibility in the future if we want to change things and not unnecessarily add an additional dependency

The working preview branch link: https://c0d3-app-mw3rh6zaq-c0d3.vercel.app/curriculum/js4/css

## Side note

This should be the last roadblock to going live with the self hosted lessons which has me really excited !

### DIFF
--- a/components/mdx/CodePen.test.js
+++ b/components/mdx/CodePen.test.js
@@ -1,0 +1,63 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import '@testing-library/jest-dom'
+import CodePen from './CodePen'
+describe('CodePen component', () => {
+  test('should render a single iframe element', () => {
+    const { container } = render(
+      <CodePen id="SOME_TEST" height="500" defaultTab="js" showResult={true} />
+    )
+    expect(container.children.length).toBe(1)
+    expect(container.children[0].tagName).toBe('IFRAME')
+  })
+
+  test('should render correct src value containing prop values', () => {
+    const { container } = render(
+      <CodePen
+        id="SOME_TEST"
+        height="1000"
+        defaultTab="html"
+        showResult={false}
+        theme="dark"
+      />
+    )
+    expect(container.children[0]).toHaveAttribute(
+      'src',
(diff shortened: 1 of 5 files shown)


tomrule007:
Well I need to look into enabling typescript on mdx files. Just force push an update because I realized I didn't update the css mdx lesson file to use my final codepen props. Originally I was planning on using an array like the mdx-embed component but when with an string instead and saw no errors during the build process.

Not a fan of seeing errors slip by like that especially when I want to be able to encourage students to use lesson content fixes as their first contributions to c0d3

ggwadera (approved):

PR #1124, opened 2021-09-14
Update#843 - Fixes utility functions in discordAuth.ts 

kondanna:
This PR adds two new User columns: the discord access token and its expiration date.

previously, our app was making two calls to userInfoController every time the user navigates. each call took a refresh token, got a new refresh token from it and updated it in database. however, the first call didn't update User's the refresh token in time for the second call, so the 'old' refresh token was coming up invalid.

discordAuth now has a new function getDiscordUserInfo, which takes in a User object (from prisma.user) and returns a discordUserInfo object. The improvement is that it just uses the stored access token to fetch discord user info, without going through the whole refresh token exchange step (unless the access token is expired).

two other files - userInfoController.ts and discord.ts used those utility functions, and therefore had to be updated with their tests.

### DIFF
--- a/__tests__/pages/discord.test.js
+++ b/__tests__/pages/discord.test.js
@@ -6,8 +6,8 @@ jest.mock('../../helpers/middleware/logger')
 
 import nextConnect from 'next-connect'
 import {
-  getTokenFromAuthCode,
-  getUserInfoFromRefreshToken
+  getDiscordUserInfo,
+  setTokenFromAuthCode
 } from '../../helpers/discordAuth'
 
 const mockDiscordUserInfo = {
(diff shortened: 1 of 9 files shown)


tomrule007 (approved):
Looks good, probably much better to not be constantly hitting the discord servers for a new token all the time anyways.

This will still have the double fetch edge case if the user visits the profile page with an expired token though right? since we didn't implement a lock feature to stop the double requests

kondanna:
> Looks good, probably much better to not be constantly hitting the discord servers for a new token all the time anyways.
> 
> This will still have the double fetch edge case if the user visits the profile page with an expired token though right? since we didn't implement a lock feature to stop the double requests

hmm you're right! Gotta find a workaround for that case.

songz (approved):

PR #1131, opened 2021-09-23
Session discord

kondanna:
1. Checks if user is connected to discord at /curriculum page
2. If not, redirect to /discord/connect page, then
3. upon completion of auth flow, redirects user back to /curriculum page


next.config.js
```diff
@@ -24,6 +24,11 @@ const moduleExports = {
         source: '/book',
         destination: '/docs/learn',
         permanent: true
+      },
+      {
+        source: '/api/auth/callback/discord',
+        destination: '/discord/success',
+        permanent: true
```
tomrule007, line 31: I think we need to update the discord application uri redirect setting and not try to have next do the redirect

pages/curriculum.tsx
```diff
@@ -130,13 +132,17 @@ export const Curriculum: React.FC<Props> = ({ lessons, alerts }) => {
   }, [])
   useEffect(() => {
     if (data && data.session) {
+      if (data.session.user && data.session.user.isConnectedToDiscord) {
```
tomrule007, line 135: Is there a big benefit to adding the extra `isConnectedToDiscord` field vs just checking that the discord username string isn't empty?

pages/curriculum.tsx
```diff
@@ -130,13 +132,17 @@ export const Curriculum: React.FC<Props> = ({ lessons, alerts }) => {
...
+        router.push('/discord/connect')
```
tomrule007, line 136: This also doesn't seem to be taking into account the user pressing the `no thank` button as it keeps send me right back to the connect page

tomrule007 (changes requested):
Looks like we got some failing tests, but I'm also questioning if we really need a new boolean value to check auth since the lack of a discord name should represent that. Or is it the boolean will be used to track if the user has opted out of authenticating with discord?

Also what about switching the initial connect page a popup modal instead of a full seperate page? I think that might be better UX and easier fix to the current issue where the curriculum flash loads before the redirect to the connect page.

And it also seems like we need to set the redirect page on the discord app setup page and not with next.

Just some of the issues I noticed we can talk about it at the sprint meeting tomorrow.

pages/curriculum.tsx
kondanna: Another error is that it should be !data.session.user.isConnectedToDiscord

the no thanks button has been updated in another PR to (unceremoniously) log out the user.

pages/curriculum.tsx
kondanna: ah, yeah. the discord username & avatar stuff is not available to session query, only userInfo query. 

the tokens and expiry are available in the req.user object, so I had to make use of those without exposing them to the frontend.

next.config.js
kondanna: I tried doing res.redirect from the /api/auth/callback/discord handler, according to the docs:

https://nextjs.org/docs/api-routes/response-helpers

But I kept getting errors. This was my workaround, heh.

res redirect
.

pages/curriculum.tsx
tomrule007: Ok that makes sense

next.config.js
tomrule007: That would still be a next redirect solution, I'm talking about setting the discord app to just send us to the right page either a success page or back to curriculum to the curriculum page.

Then we could merge the functionality of the discord/api/auth page into that page with a useEffect on the `code` query string

https://discord.com/developers/applications/ :our app id:/oauth2
discordRedirect

tomrule007: > I tried doing res.redirect from the /api/auth/callback/discord handler, according to the docs:
> 
> https://nextjs.org/docs/api-routes/response-helpers
> 
> But I kept getting errors. This was my workaround, heh.
> 
> res redirect
> 
> .

Though I would think the res.redirect("/discord/success") should work

pages/curriculum.tsx
tomrule007: seems harsh

Issue #892, opened 2021-06-23
Line comments expand when adding a new one


ggwadera:
As the title says, when you have hidden line comments they all get expanded when adding a new one, it's kinda annoying.

Issue #848, opened 2021-06-09
Opening another line comment box clears an already open one


ggwadera:
When you already have a line comment already open you better submit it before trying to do another one, otherwise all your effort in writing a good comment is wasted :(

Steps to reproduce:

1. Start a new line comment in a review and type something but don't submit.
![image](https://user-images.githubusercontent.com/16023489/121274861-ca0be880-c8a1-11eb-85df-44b20a4fc4d1.png)

2. Keep the previous comment open and open another one in another line.
![image](https://user-images.githubusercontent.com/16023489/121274915-e90a7a80-c8a1-11eb-8863-dbddbae13951.png)

3. Go back to the first comment, you will be sad to see that it was cleared and you will have to do it all again 😥
![image](https://user-images.githubusercontent.com/16023489/121274936-f162b580-c8a1-11eb-9f98-3848fecf3647.png)

ggwadera:
@Ulisseus

songz:
haha ouch!

Ulisseus:
Yes, I never thought that someone would write a comment and instead of submitting it would start writing the new one. Creating new commentbox triggers rerender so everything which isn't written to db disappears. 

I'm trying to add normal comments under submission right now. Once I finish it I'll come back to this.

PR #1138, opened 2021-10-10
Add Review Page Local Storage

ghost:
This will allow us to build solutions for issues #848 and #892


InerLeeLofgren:
> I think we first need to think through how to handle clearing of the localstorage. Like... what if I reviewed something, then decided to not review it. Would we be taking up storage forever?

Would adding a time value be an appropriate solution? 
And then perform a check to see if the time exceeds say 7 days, to clear any local storage values that meet that criteria?

InerLeeLofgren:
Closing Merge, going to apply a different approach

PR #1179, opened 2021-10-31
Update #843 - allows user to opt out of Discord

kondanna:
after further consideration, perhaps allowing user to opt out of connecting to Discord would make for a friendlier experience.

the modal will still pop up when /curriculum page refreshes though, since the component reloads.

### DIFF
--- a/components/ConnectToDiscordModal.test.js
+++ b/components/ConnectToDiscordModal.test.js
@@ -1,0 +1,15 @@
+import React from 'react'
+import ConnectToDiscordModal from './ConnectToDiscordModal'
+import { render, waitFor, fireEvent, screen } from '@testing-library/react'
+
+describe('ConnectToDiscordModal component', () => {
+  it('should close modal if user opts out of connecting to Discord', async () => {
+    let res = ''
+    const expectedResult = 'potatus maximus'
+    render(
+      <ConnectToDiscordModal show={true} close={() => (res = expectedResult)} />
+    )
+    await waitFor(() => fireEvent.click(screen.getByText(/No thanks/)))
+    expect(res).toEqual(expectedResult)
+  })
+})
--- a/components/ConnectToDiscordModal.test.js
+++ b/components/ConnectToDiscordModal.test.js
@@ -3,7 +3,7 @@ import ConnectToDiscordModal from './ConnectToDiscordModal'
 import { render, waitFor, fireEvent, screen } from '@testing-library/react'
 
 describe('ConnectToDiscordModal component', () => {
-  it('should close modal if user opts out of connecting to Discord', async () => {
+  it('should call close modal function if user opts out of connecting to Discord', async () => {
     let res = ''
     const expectedResult = 'potatus maximus'
     render(
@@ -11,5 +11,14 @@ describe('ConnectToDiscordModal component', () => {
     )
     await waitFor(() => fireEvent.click(screen.getByText(/No thanks/)))
     expect(res).toEqual(expectedResult)
+  })
+  it('should not show modal if user opts out of connecting to Discord', async () => {
+    const mockProps = {
+      show: true,
+      close: jest.fn()
+    }
+    render(<ConnectToDiscordModal {...mockProps} />)
+    await waitFor(() => fireEvent.click(screen.getByText(/No thanks/)))
+    expect(mockProps.close).toHaveBeenCalledTimes(1)
   })
 })
--- a/components/ConnectToDiscordModal.tsx
+++ b/components/ConnectToDiscordModal.tsx
@@ -53,10 +53,7 @@ const ConnectToDiscordModal: React.FC<ConnectToDiscordModalProps> = ({
             Connect Now
           </button>
         </NavLink>
-        <div
-          className="text-center"
-          onClick={/* istanbul ignore next */ () => close()}
-        >
+        <div className="text-center" onClick={() => close()}>
           <a href="#">No thanks, I&apos;ll study on my own.</a>
         </div>
       </div>
--- a/components/ConnectToDiscordModal.tsx
+++ b/components/ConnectToDiscordModal.tsx
@@ -30,7 +30,7 @@ const ConnectToDiscordModal: React.FC<ConnectToDiscordModalProps> = ({
   /*
     show = false // means connected to Discord
 
-    show | localStorage too old | Final
+    show | showAgain            | Final
     0    |  0  (doesn't matter) |  0
     0    |  1  (doesn't matter) |  0
     1    |  0                   |  0
--- a/components/ConnectToDiscordModal.tsx
+++ b/components/ConnectToDiscordModal.tsx
@@ -1,22 +1,23 @@
 import React from 'react'
 import { ModalCard, ModalSize } from './ModalCard'
 import NavLink from './NavLink'
-import LogoutContainer from './LogoutContainer'
 
 type ConnectToDiscordModalProps = {
   show: boolean
+  close: Function
 }
 
 const discordConnectPage = `${process.env.NEXT_PUBLIC_DISCORD_CALLBACK_URI}`
 
 const ConnectToDiscordModal: React.FC<ConnectToDiscordModalProps> = ({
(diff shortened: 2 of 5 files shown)


components/ConnectToDiscordModal.tsx
```diff
@@ -52,11 +53,12 @@ const ConnectToDiscordModal: React.FC<ConnectToDiscordModalProps> = ({
...
           </button>
         </NavLink>
-        <LogoutContainer>
-          <div className="text-center">
-            <a href="#">No thanks, I&apos;ll study on my own.</a>
-          </div>
-        </LogoutContainer>
+        <div
+          className="text-center"
+          onClick={/* istanbul ignore next */ () => close()}
```
hwong0305, line 58: What is istanbul ignored next used for? 

If it's not needed, it should be taken out.

kondanna: If it's not there, code cov will say that line isn't covered.

components/ConnectToDiscordModal.tsx
```diff
@@ -52,11 +53,9 @@ const ConnectToDiscordModal: React.FC<ConnectToDiscordModalProps> = ({
             Connect Now
           </button>
         </NavLink>
-        <LogoutContainer>
-          <div className="text-center">
-            <a href="#">No thanks, I&apos;ll study on my own.</a>
-          </div>
-        </LogoutContainer>
+        <div className="text-center" onClick={() => close()}>
```
anthonykhoa, line 56: The only thing the onClick callback is doing is calling the close function, so you can just pass the close function as the value. 
```
onClick={close}
```

components/ConnectToDiscordModal.tsx
hwong0305: Can you please add a comment explaining that?

anthonykhoa: Hmm so you are saying that the this line is actually covered by a test, but code cov doesn't recognize that its covered?

pages/curriculum.tsx
```diff
@@ -171,7 +171,12 @@ export const Curriculum: React.FC<Props> = ({ lessons, alerts }) => {
   })
   return (
     <>
-      <ConnectToDiscordModal show={showConnectToDiscordModal} />
+      <ConnectToDiscordModal
+        show={showConnectToDiscordModal}
+        close={
+          /* istanbul ignore next */ () => setShowConnectToDiscordModal(false)
```
anthonykhoa, line 177: Hmm so you are saying that the this line is actually covered by a test, but code cov doesn't recognize that its covered?

anthonykhoa: I don't see a test in here that corresponds to to adding this `close` button update.
https://github.com/garageScript/c0d3-app/blob/master/__tests__/pages/curriculum.test.js

Maybe if you updated the test, the line would be covered.

components/ConnectToDiscordModal.tsx
kondanna: Hmm this gives me the following TypeScript error:

Type 'Function' is not assignable to type 'MouseEventHandler<HTMLDivElement>'.
  Type 'Function' provides no match for the signature '(event: MouseEvent<HTMLDivElement, MouseEvent>): void

components/ConnectToDiscordModal.tsx
```diff
@@ -1,23 +1,47 @@
...
 
 type ConnectToDiscordModalProps = {
   show: boolean
+  close: Function
 }
 
 const discordConnectPage = `${process.env.NEXT_PUBLIC_DISCORD_CALLBACK_URI}`
 
+const CONNECT_TO_DISCORD_LOCALSTORAGE_KEY = 'connect-to-discord'
+const SEVEN_DAYS = 1000 * 60 * 60 * 24 * 7
```
anthonykhoa, line 13: Looks like there is another place that uses the `7 day` constant here:
https://github.com/garageScript/c0d3-app/blob/b1a6736084ac6cb1fde2a2d6fb57c2d64537979f/helpers/middleware/session.ts#L6

Can you make the `7 days` variable to be used as a constant?

I think there can probably be some little more refactoring to use constants to reduce duplicated code too:
https://github.com/garageScript/c0d3-app/search?q=1000+*+60+*+60+*+24+*+7

^ The refactoring should be another PR though. But I think you can start in this PR by making a constant for `7 days` to be used for any files in the future(`including this file`)

anthonykhoa: I just looked through the codebase and I don't see any kind of constants file. So I think you'd be making the first one in this repo.

I'm thinking it would be best to have a `constants` folder in the root of this repo, and inside there can be individual files relating to some topic. For example we could have a `requestMethods.ts` file that would only contain constants used for requests, like `GET` or `POST`.

Thoughts?
@songz @hwong0305 @kondanna @tomrule007

components/ConnectToDiscordModal.tsx
anthonykhoa: Ah, you are right -- the first param of onclick event is the `event` variable, so () => close() makes sense.

components/ConnectToDiscordModal.tsx
kondanna: Sure, I'm for it, if it's an accepted practice.

pages/curriculum.tsx
kondanna: any idea how to test this close prop? this modal is kind of a child component of <Curriculum />
i already test the ConnectToDiscordModal in its own test file.

components/ConnectToDiscordModal.tsx
kondanna: line covered

pages/curriculum.tsx
anthonykhoa: I think what's not being tested is the calling of the function inside of the function you are passing to `close` as a prop. 

```
setShowConnectToDiscordModal(false)
```

anthonykhoa: I think what's not being tested is the calling of the function inside of the function you are passing to close as a prop.

kondanna: That's what it looks like to me too. I figured if it's a function that calls a useState method, there's not much to test.

kondanna: Finally figured out how to cover it!

components/ConnectToDiscordModal.tsx
kondanna: let's revisit this on another PR

hwong0305: I think it's a good idea to store all our commonly used constants or variables that keeps getting reused into a constant folder. It would have us having to repeat ourselves and in the future if we need to change something, we can change it in one place instead of many different places.

kondanna: You want this refactor in this PR or a separate one?

kondanna: Decided to refactor in another PR: 
https://github.com/garageScript/c0d3-app/issues/1185

components/ConnectToDiscordModal.test.js
```diff
@@ -0,0 +1,24 @@
...
+    expect(res).toEqual(expectedResult)
+  })
+  it('should not show modal if user opts out of connecting to Discord', async () => {
+    const mockProps = {
+      show: true,
+      close: jest.fn()
+    }
+    render(<ConnectToDiscordModal {...mockProps} />)
+    await waitFor(() => fireEvent.click(screen.getByText(/No thanks/)))
+    expect(mockProps.close).toHaveBeenCalledTimes(1)
```
songz, line 22:
```js

const mockClock = jest.fn().mockImplementation( () => {
  res = expectedResult
  })

    await waitFor(() => fireEvent.click(screen.getByText(/No thanks/)))
    expect(res).toEqual(expectedResult)
    expect(mockClose).toHaveBeenCalledTimes(1)
    })
    ```
    
    Could you try this?

songz (approved):

mino323 (approved):

PR #1289, opened 2022-01-14
Update NextJS to v12

### LINKED ISSUE #1264: Update NextJS to v12

SahilKalra98:
**Closes #1264**

**This PR updates Next to version 12 using the steps here**:
https://nextjs.org/docs/upgrading

**In-Progress:**

The `helpers/apolloClient-server.ts` file needs to be refactored a bit more since there is duplicate code copied from `helpers/apolloClient.ts`.

The `apolloClient-server.ts` file needed to be created because it seems like dynamic importing broke while upgrading to version 12. [This link](https://github.com/garageScript/c0d3-app/issues/1264) talks more about the issue.

---
**Version 12 Changes**
Here is a [link](https://nextjs.org/blog/next-12#es-modules-support-and-url-imports) to all of the changes introduced into Next.js 12. Some of the big changes I saw was Next transitioning from Babel to SWC (Rust-based compiler to compile JavaScript/TypeScript) which is supposed to be a lot faster than babel and also support for ES Modules. I didn't include SWC in this PR because it could make the PR bigger so I decided we could tackle this in another PR if we decide to transition to it. A link to SWC is [here](https://swc.rs/)

**Concerns/Issues**

1. The first issue I saw was when I removed the `babel.config.js` file to use SWC, the server worked fine and the files were able to get compiled however all the tests broke. Most of the errors I saw were `cannot use import` errors or jsx experimental flag is not set. I looked at the Jest docs [here](https://jestjs.io/docs/ecmascript-modules) which we can try out in another PR to avoid making this PR too large. Also at the bottom of that link it states that `jest.mock` is not supported in a clean way and is still experimental. To look at the errors when removing the `babel.config.js` you can do the following:
   1. `rm babel.config.js`
   2. `yarn test`

2. The second issue that arose was after upgrading to version 12, the `__tests__/pages/curriculum.test.js` file was not passing. The error that came up was this `ReferenceError: setImmediate is not defined` from the `nodemailer` package in `helpers/mail.ts`. In order to solve the problem I used this [link](https://github.com/prisma/prisma/issues/8558) and the method described at the bottom of the discussion. I wasn't sure if this was a good way of solving the problem, would appreciate some feedback! In order to reproduce the error you can checkout this branch and then do the following:
    1. `yarn install`
    2. `rm jest.env.js`
    3. Remove the key `setupFilesAfterEnv` inside `jest.config.js`
    4. Then run `yarn test`

### DIFF
--- a/helpers/apolloClient-server.ts
+++ b/helpers/apolloClient-server.ts
@@ -3,61 +3,17 @@ import {
   InMemoryCache,
   NormalizedCacheObject
 } from '@apollo/client'
-import merge from 'deepmerge'
-import isEqual from 'lodash/isEqual'
 import { SchemaLink } from '@apollo/client/link/schema'
 import { schema } from '../graphql/schema'
 
-let apolloClient: ApolloClient<NormalizedCacheObject>
-const cache = new InMemoryCache()
-
-function createIsomorphLink() {
-  if (typeof window === 'undefined') {
-    return new SchemaLink({ schema })
-  } else {
-    const { HttpLink } = require('@apollo/client/link/http')
-    return new HttpLink({
-      uri: '/api/graphql',
-      credentials: 'same-origin'
-    })
-  }
-}
-
 function createApolloClient() {
   return new ApolloClient({
-    ssrMode: typeof window === 'undefined',
-    link: createIsomorphLink(),
-    cache
+    ssrMode: true,
+    link: new SchemaLink({ schema }),
+    cache: new InMemoryCache()
   })
 }
 
-export function initializeApollo(
-  initialState: NormalizedCacheObject | null = null
-) {
-  const _apolloClient = apolloClient || createApolloClient()
-
-  // If your page has Next.js data fetching methods that use Apollo Client, the initial state
-  // gets hydrated here
-  if (initialState) {
-    // Get existing cache, loaded during client side data fetching
-    const existingCache = _apolloClient.extract()
-
-    // Merge the existing cache into data passed from getStaticProps/getServerSideProps
-    const data = merge(initialState, existingCache, {
-      // combine arrays using object equality (like in sets)
-      arrayMerge: (destinationArray, sourceArray) => [
-        ...sourceArray,
-        ...destinationArray.filter(d => sourceArray.every(s => !isEqual(d, s)))
-      ]
-    })
-
-    // Restore the cache with the merged data
-    _apolloClient.cache.restore(data)
-  }
-  // For SSG and SSR always create a new Apollo Client
-  if (typeof window === 'undefined') return _apolloClient
-  // Create the Apollo Client once in the client
-  if (!apolloClient) apolloClient = _apolloClient
-
-  return _apolloClient
+export function initializeApollo(): ApolloClient<NormalizedCacheObject> {
+  return createApolloClient()
 }
--- a/helpers/apolloClient-server.ts
+++ b/helpers/apolloClient-server.ts
@@ -6,14 +6,10 @@ import {
 import { SchemaLink } from '@apollo/client/link/schema'
 import { schema } from '../graphql/schema'
 
-function createApolloClient() {
+export function initializeApollo(): ApolloClient<NormalizedCacheObject> {
   return new ApolloClient({
     ssrMode: true,
     link: new SchemaLink({ schema }),
     cache: new InMemoryCache()
   })
-}
-
-export function initializeApollo(): ApolloClient<NormalizedCacheObject> {
-  return createApolloClient()
 }
--- a/helpers/apolloClient-server.ts
+++ b/helpers/apolloClient-server.ts
@@ -1,5 +1,3 @@
-//from https://github.com/vercel/next.js/tree/canary/examples/with-apollo
-import { useMemo } from 'react'
 import {
   ApolloClient,
   InMemoryCache,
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -36,6 +36,8 @@
     "@prisma/client": "^2.29.1",
     "@quixo3/prisma-session-store": "^3.1.2",
     "@sentry/nextjs": "^6.13.3",
+    "@swc/cli": "^0.1.55",
```
songz, line 39: this is packaged into nextv12

tsconfig.json
```diff
@@ -19,13 +16,9 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "esnext"
+    "target": "esnext",
+    "incremental": true
```
songz, line 20: why did you add `incremental: true`?

helpers/apolloClient-server.ts
```diff
@@ -0,0 +1,63 @@
...
+function createApolloClient() {
+  return new ApolloClient({
+    ssrMode: typeof window === 'undefined',
+    link: createIsomorphLink(),
+    cache
+  })
+}
+
+export function initializeApollo(
+  initialState: NormalizedCacheObject | null = null
```
songz, line 35: if you look at [how its used](https://github.com/garageScript/c0d3-app/pull/1289/files#diff-127feee063dd438c812f7f1ddfadfe8f2b1c8e2b6a865342a668c0e7075f67c2R220), it seems like we don't need `initialState`

helpers/apolloClient-server.ts
```diff
@@ -0,0 +1,63 @@
...
+) {
+  const _apolloClient = apolloClient || createApolloClient()
+
+  // If your page has Next.js data fetching methods that use Apollo Client, the initial state
+  // gets hydrated here
+  if (initialState) {
```
songz, line 41: comment above

helpers/apolloClient-server.ts
```diff
@@ -0,0 +1,63 @@
...
+        ...sourceArray,
+        ...destinationArray.filter(d => sourceArray.every(s => !isEqual(d, s)))
+      ]
+    })
+
+    // Restore the cache with the merged data
+    _apolloClient.cache.restore(data)
+  }
+  // For SSG and SSR always create a new Apollo Client
+  if (typeof window === 'undefined') return _apolloClient
```
songz, line 58: we don't need this `if` check because it's server side apolloClient so window should always be `undefined`

helpers/apolloClient-server.ts
```diff
@@ -0,0 +1,63 @@
...
+import merge from 'deepmerge'
+import isEqual from 'lodash/isEqual'
+import { SchemaLink } from '@apollo/client/link/schema'
+import { schema } from '../graphql/schema'
+
+let apolloClient: ApolloClient<NormalizedCacheObject>
+const cache = new InMemoryCache()
+
+function createIsomorphLink() {
+  if (typeof window === 'undefined') {
```
songz, line 15: you don't need this check

helpers/apolloClient-server.ts
```diff
@@ -0,0 +1,63 @@
...
+    return new HttpLink({
+      uri: '/api/graphql',
+      credentials: 'same-origin'
+    })
+  }
+}
+
+function createApolloClient() {
+  return new ApolloClient({
+    ssrMode: typeof window === 'undefined',
```
songz, line 28: this could be hardcoded

helpers/apolloClient.ts
```diff
@@ -47,7 +47,7 @@ function createIsomorphLink() {
   }
 }
 
-function createApolloClient() {
+export function createApolloClient() {
```
songz, line 50: why does this need to be exported?

songz:
could you also update `apolloClient` accordingly? seems like `window` will never be `undefined` in `apolloClient` https://github.com/garageScript/c0d3-app/pull/1289/files#diff-f73b56ae46bdf4aff9c3bcea9558552bc858e86d51a638186f395753438bfa6eR37-R40

tsconfig.json
SahilKalra98: When I upgraded to v12, and then did `yarn dev`, the log statements said that it configured the tsconfig.json file. The only thing added was `"incremental": true`

songz: ah it stores info about the complied files: https://www.typescriptlang.org/tsconfig#incremental

helpers/apolloClient-server.ts
```diff
@@ -0,0 +1,19 @@
...
+function createApolloClient() {
+  return new ApolloClient({
+    ssrMode: true,
+    link: new SchemaLink({ schema }),
+    cache: new InMemoryCache()
+  })
+}
+
+export function initializeApollo(): ApolloClient<NormalizedCacheObject> {
+  return createApolloClient()
```
songz, line 18: `creatApolloClient` seems like an unnecessary function. Perhaps just `return new ApolloClient({....})`?

helpers/apolloClient.ts
```diff
@@ -33,24 +30,18 @@ const whiteList = ['/curriculum']
...
-    return new SchemaLink({ schema })
-  } else {
-    const { HttpLink } = require('@apollo/client/link/http')
-    return new HttpLink({
-      uri: '/api/graphql',
-      credentials: 'same-origin'
-    })
-  }
+function createHttpLink() {
+  const { HttpLink } = require('@apollo/client/link/http')
```
songz, line 34: seems like this doesn't need to be `require`'d dynamically. Could you try importing it

helpers/apolloClient.ts
```diff
@@ -33,24 +30,18 @@ const whiteList = ['/curriculum']
...
+  return new HttpLink({
```
songz, line 35: if the import above works then we don't need this `creatHttpLink` helper function anymore

songz (approved):

flacial (approved):

helpers/apolloClient.ts
```diff
@@ -18,10 +18,7 @@ const whiteList = ['/curriculum']
 
 //there is no global await and this logic can't be refactored into other functions because pages/_app.tsx can't be async
 ;(async function () {
-  if (
-    typeof window !== 'undefined' &&
-    whiteList.includes(window.location.pathname)
-  ) {
+  if (whiteList.includes(window.location.pathname)) {
```
flacial, line 21: I think this is why Vercel build throw errors. The code first runs server-side and at that period, `window` is undefined.

The past check of whether `window` is undefined or not has to return.

```suggestion
 if (
    typeof window !== 'undefined' &&
    whiteList.includes(window.location.pathname)
    )
```

flacial (changes requested):

helpers/apolloClient.ts
SahilKalra98: Thanks!

helpers/apolloClient.ts
SahilKalra98: I tried importing and in the testfile, it seems that MockFunction was never called [here](https://github.com/garageScript/c0d3-app/blob/master/helpers/apolloClient.test.js#L63)

flacial (approved):

PR #1298, opened 2022-01-14
Close #1257 - Fix broken login messages

### LINKED ISSUE #1257: Login Error Reporting Broken
image

It should say something like `user does not exist or incorrect password`.

flacial:
I'm not certain what the initial graphQL error responses were, but currently they only include the wanted messages `User does not exist` instead of `UserInputError: User does not exist`.

`const formattedMessage = message.split(':')[1]`

There's no element in the first index, hence the `undefined` message. `message` will be used as the alert text directly.

Before:
image

After:
image

**Demo:** [Link](https://c0d3-app-ak90eft7h-c0d3.vercel.app/login)


flacial:
Fixing it.

songz (approved):

mino323 (approved):

PR #1302, opened 2022-01-15
Close #1181 - Go back button isn't working as intended in some cases

### LINKED ISSUE #1181: "Go back" button does not work correctly for certain circumtances
Currently the `Go back` button is implemented by using `router.back`, which executes `history.back`. According to [the docs](https://developer.mozilla.org/en-US/docs/Web/API/History/back), `if there is no previous page, this method call does nothing.`

Consider these scenarios:

1) User is on `facebook.com`, then decides to enter https://www.c0d3.com/review/js4 into the address bar. When the user lands on https://www.c0d3.com/review/js4, if they click `go back`, the user will go back to `facebook.com`. I think it's weird for the `go back` button to redirect you to the last website you were on, even if it wasn't c0d3.com

2)  A user is on discord and notices a submission review post by the c0d3 bot. The user clicks on the link to review -- clicking on the link opens up a new tab for the user. When on the review page, if the user clicks `go back`, then nothing will happen because the first history of that tab is opening the review page. The button should do something when clicked, or else it can be perceived as broken

Here is where the `Go back` button is implemented:
https://github.com/garageScript/c0d3-app/blob/39e250c47adb3d4ed137d311c6db9dc3305be992/components/LessonTitleCard.tsx#L50

# Proposals

1) If a user clicks `go back` but the redirection url is not `c0d3.com` domain, then redirect go c0d3 homepage. If there is no `history` for the `go back` button to utilize to redirect, then just redirect back to c0d3 homepage

flacial:
Issue: #1181

When "Go back" button in Learn page and any of the lessons is clicked, it goes back to whatever previous route despite it being c0d3.com or google.com

Before: 

1. I'm here (GitHub)
2. Went to https://www.c0d3.com/curriculum/js0 OR https://www.c0d3.com/docs/learn
3. Clicked "Go back"
4. **Goes back to here again (GitHub)**

After: 

1. I'm here (GitHub)
2. Went to https://c0d3-app-lvfonog2g-c0d3.vercel.app/curriculum/js0 OR https://c0d3-app-lvfonog2g-c0d3.vercel.app/docs/learn
3. Clicked "Go back"
4. **Goes back to `/`** (If logged in, it'll redirect to `/curriculum`, else `/`)


flacial:
I was almost going to include another issue's changes here. I think to create two PRs, I would have to make each one from its own branch and merge them with master once they are done, like in the C0D3 challenges submissions.

components/LessonTitleCard.tsx
```diff
@@ -44,7 +44,8 @@ const LessonTitleCard: React.FC<LessonTitleProps> = ({
                   // Link does not get correct styles without the href
                   // Prevent scroll to top before going back
                   e.preventDefault()
-                  router.back()
+
+                  router.push('/')
```
anthonykhoa, line 48: Do we always want to redirect to homepage when a user clicks `Go back`?

Let's say I'm looking at js1 challenges. And then I click the `lesson` button that shows the lesson document for js1. Currently, if I click `Go back`, then I will land back on the js1 challenges page. 

What I believe your change will do is it will always redirect users back to the homepage -- is that what we want? IMO I think its a better user experience to redirect users to the last page they were on, and only redirect to the homepage if the redirectUrl in question is not c0d3.com domain

flacial: I'll look into it

flacial: My first thought is it needs a global state that will be true if the user accessed any page in c0d3.
```suggestion
                inC0d3 ? router.back() : router.push('/')
```

songz:
updated the main issue, let's vote? https://github.com/garageScript/c0d3-app/issues/1181

flacial:
I'll close it for now, as my master got messed up. I'll correct master to make a separate branch for this issue and other issues.

PR #1305, opened 2022-01-17
Fix - Bump @types/node from 16.7.5 to 17.0.8

flacial:
Fix the issues when updating to v17.0.8.

Issue:

URLSearchParams arguments can't be `type | undefined`

```
helpers/discordAuth.ts(65,31): error TS2345: Argument of type '{ grant_type: string; client_id: string | undefined; client_secret: string | undefined; code: string; redirect_uri: string | undefined; scope: string; }' is not assignable to parameter of type 'string | URLSearchParams | Record<string, string | readonly string[]> | Iterable<[string, string]> | readonly [string, string][] | undefined'.
```

### DIFF
--- a/helpers/discordAuth.ts
+++ b/helpers/discordAuth.ts
@@ -4,9 +4,9 @@ import { User } from '.prisma/client'
 import fetch from 'node-fetch'
 
 const discordAPI = 'https://discordapp.com/api'
-export const client_id = process.env.DISCORD_KEY
-export const client_secret = process.env.DISCORD_SECRET
-export const redirect_uri = process.env.DISCORD_REDIRECT_URI // {baseurl}/api/auth/callback/discord
+export const client_id = process.env.DISCORD_KEY!
+export const client_secret = process.env.DISCORD_SECRET!
+export const redirect_uri = process.env.DISCORD_REDIRECT_URI! // {baseurl}/api/auth/callback/discord
 
 type AccessTokenResponse = {
   access_token: string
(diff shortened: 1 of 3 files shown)


JasirZaeem:
The types being used should correspond to the library being used, or the node engine in case of @types/node. We are using 14 on vercel I don't think we need to update types to node 17

flacial:
> The types being used should correspond to the library being used, or the node engine in case of @types/node. We are using 14 on vercel I don't think we need to update types to node 17

I just did some research and what you said makes sense. 

According to what you said, we should let the library handle updating node types?

helpers/discordAuth.ts
```diff
@@ -99,19 +95,21 @@ const getTokenFromRefreshToken = (
       grant_type: 'refresh_token',
       refresh_token
     })
-  }).then(r => r.json())
+  })
+  return await r.json()
```
songz, line 99: I don't think its right to do `return await <promise>` because the nature of `async` will return a promise already so we don't have to wait for the last promise to resolve before returning.

```js
return r.json()
``` 

should suffice. same with the rest.

songz:
we could upgrade node to v17: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version

songz:
Actually just checked, vercel only supports v14.

 good catch @JasirZaeem ! Thanks.

Updated our docs so we know about this in the future: https://github.com/garageScript/c0d3-app/wiki/Dependabot#notes

I vote to close this.

flacial:
Closed it as NextJS only supports node v14.

Issue #1192, opened 2021-12-08
C0d3 Dojo


mino323:
# Intro

## Goals

One of the main benefits of learning is for users to engage creatively with the exercise sets. Their learning would be improved if they could solve creative uses of the concepts they are learning, and also come up with their own creative exercises.

This feature is designed to engage users by giving them more practice exercises and helping them engage other students with their own creative exercises.

### Problem

Currently, the only engagement that users have in their learning journey is submitting challenge exercises for each lesson module. 8-12 challenges per module. Besides this, it is all reading and self paced, self guided exercises. Most students do not do all of the exercises, and new exercises are difficult to add, so content is a bit old. New exercises are added at a rate of a few exercises a year!

### Solution

There are two types of student groups in our solution. First is the learning group, second is the mentor group.

**Learners Group**

When a student starts a module, they could jump right into exercises and get quick feedback by submitting answers or writing code, and getting quick feedback (your answer or implementation is right / wrong). Each exercise should also support a discussion where students could ask questions. This approach has several benefits:

- Quick feedback - When users submit a solution, right away they know whether it is correct or wrong. This is a huge improvement over our existing "reading" format where the users could have made a mistake and would never know.
- Direct discussion and higher student - student engagement - If a student has a question, they could directly ask on the page, on the specific exercise. Currently, when a student has a question on the *exercise* section, it is difficult to ask for help. The student would have to share a link to the exercise and ask on discord. And not every student has done the exercise before so a student currently don't get quick support when they ask.
- Fast navigation and review - System could keep track of what the users are weak at. Students are able to pickup where they left off.

Sample UI - Note: A exercise could also ask users to submit a function to do X. The check for correctness would then be a test suite instead of a hard-coded string.

![Untitled](https://s3-us-west-2.amazonaws.com/secure.notion-static.com/97779590-24a7-4cb6-a4b2-76a87687f9c7/Untitled.png)

Notes:

- Infinite number of exercises for each section. Some students feel comfortable after solving 3 exercises, some after 100. Up to them!

**Mentors Group**

Mentors are users who has completed a module. They could then submit exercise to challenge the future generations of students (which would show up in the prior example UI). Submitting exercises could be fun (and therefore more engaging for the students) and educational in the following ways:

- Seeing all the wrong submissions from students: This will help the mentor gain a deeper understanding of where common traps students may fall for, which would help them pay attention to their own code in the future!
[...]

PR #1354, opened 2022-01-28
adds to typeDefs.ts

mino323:
Next steps of https://github.com/garageScript/c0d3-app/issues/1192

This pr updates typedefs to include a way to create, read and delete modules. 

After writing I ran ```yarn generate``` to update index.tsx types.

Next step is to write the resolvers.


graphql/typeDefs.ts
```diff
@@ -5,6 +5,7 @@ export default gql`
     lessons: [Lesson!]!
     session: Session!
     allUsers: [User]
+    allModules: [Module]
```
songz, line 8: should this be `[Module]!` so we guarantee returning an array always?

Also, I think `modules` is better than `allModules`. `allUsers` should also be changed to `users` but that's for another time

graphql/typeDefs.ts
```diff
@@ -53,6 +54,13 @@ export default gql`
       submissionId: Int!
       content: String!
     ): Comment
+    addModule(
+      authorId: Int!
+      lessonId: Int!
+      name: String!
+      content: String!
+    ): Module
+    deleteModule(id: Int, name: String!): SuccessResponse
```
songz, line 63: I don't think name is needed.

mino323: I'll try it out

flacial (approved):

graphql/typeDefs.ts
```diff
@@ -206,4 +214,14 @@ export default gql`
     student: User!
     lesson: Lesson!
   }
+
+  type Module {
+    id: Int!
+    author: User!
+    authorId: Int!
```
songz, line 221: authorId is inside the author object, so it is redundant here

graphql/typeDefs.ts
```diff
@@ -206,4 +214,14 @@ export default gql`
...
+    lesson: Lesson!
+    lessonId: Int!
```
songz, line 223: `lessonId` is inside the lesson object, so this is redundant here. Users could just use `lesson.id`

graphql/typeDefs.ts
```diff
@@ -53,6 +54,13 @@ export default gql`
       submissionId: Int!
       content: String!
     ): Comment
+    addModule(
+      authorId: Int!
```
songz, line 58: when you add module, `authorId` should not be in the input. This comes from the cookie. If you do it this way, it opens up a potential security flaw where I would be able to add a module and put someone else's id to pretend like someone else created the module.

PR #1377, opened 2022-01-31
Adds module crud and module crud tests

mino323:
This pr adds module crud and module crud tests.


graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,35 @@
...
+export const modules = async () => {
+  return prisma.module.findMany({
+    include: {
+      author: true
+    }
+  })
+}
+
+export const addModule = async (_parent: void, args: MutationAddModuleArgs) => {
+  //isAdminOrThrow(req) will add after, easier for testing
```
anthonykhoa, line 15: I am not understanding this. Why are we not using `isAdminOrThrow`, and instead have it commented out?

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,35 @@
+// import { Context } from '../../@types/helpers'
```
anthonykhoa, line 1: Remove this commented out line

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,35 @@
...
+  return prisma.module.findMany({
+    include: {
+      author: true
+    }
+  })
+}
+
+export const addModule = async (_parent: void, args: MutationAddModuleArgs) => {
+  //isAdminOrThrow(req) will add after, easier for testing
+  // const authorId = ctx.req.user?.id
```
anthonykhoa, line 16: Remove this commented line if its unneeded

graphql/queryResolvers/moduleCrud.ts
songz: @anthonykhoa is right, dont put testing code into production!

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,35 @@
...
+  })
+}
+
+export const deleteModule = async (
+  _parent: void,
+  args: MutationDeleteModuleArgs
+) => {
+  // isAdminOrThrow(req)
+  const { id } = args
+  if (!id) throw new Error('Missing parameter')
```
songz, line 32: question here, if you have marked `id` as required: https://github.com/garageScript/c0d3-app/pull/1354/files#diff-0776d3ad89186de849e067b2a7b9a9701408e7b983a6387c50819581ad6f8f9aR63

Do you still need that check?

anthonykhoa:
What are these changes for? What do we need module crud for?

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,29 @@
+import type { MutationDeleteModuleArgs, MutationAddModuleArgs } from '..'
+import prisma from '../../prisma'
+
+export const modules = async () => {
+  return prisma.module.findMany({
```
anthonykhoa, line 5: Hmm `findMany` should already return a promise, right? If that's the case, then it is redundant to make this function async

anthonykhoa: Actually, I think it'd be better to keep the function async, and then wrap the promise in a try/catch block
```
try {
  const allModules = await prisma.module.findMany(...)
} catch(err) {
  // SOME LOGGING
  // error handling logic
}
```

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+  })
+}
+
+export const addModule = async (_parent: void, args: MutationAddModuleArgs) => {
+  const { authorId, content, lessonId, name } = args
+  if (!authorId || !content || !name || !lessonId) {
+    throw new Error('Missing parameters')
+  }
+  return await prisma.module.create({
```
anthonykhoa, line 17: What if this results in error? We should wrap this in a try/catch block

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,29 @@
...
+    data: { authorId, content, lessonId, name }
+  })
+}
+
+export const deleteModule = async (
+  _parent: void,
+  args: MutationDeleteModuleArgs
+) => {
+  const { id } = args
+  await prisma.module.delete({ where: { id } })
```
anthonykhoa, line 27: We should wrap this in a try/catch block

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,29 @@
...
+  return prisma.module.findMany({
+    include: {
+      author: true
+    }
+  })
+}
+
+export const addModule = async (_parent: void, args: MutationAddModuleArgs) => {
+  const { authorId, content, lessonId, name } = args
+  if (!authorId || !content || !name || !lessonId) {
```
anthonykhoa, line 14: Are we only trying to check if all arguments are present, or instead are we checking if the arguments are not the correct value? For example, if I passed the `content` value as empty string(`''`), is that considered a missing parameter?

anthonykhoa: If you are only checking if all required arguments are present, you don't have to do that. Graphql takes care of that for you already -- if you marked in the typedefs that arguments are required for this resolver, then the graphql request will already return an error if a person does not include required arguments in their request.

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,29 @@
+import type { MutationDeleteModuleArgs, MutationAddModuleArgs } from '..'
+import prisma from '../../prisma'
+
+export const modules = async () => {
```
anthonykhoa, line 4: You should use typescript for all functions you declare so that you can define what they return

.env.example
```diff
@@ -1,34 +0,0 @@
-#
```
songz, line 1: please undo the deletion of this file. We don't want this file to be deleted

graphql/queryResolvers/moduleCrud.ts
songz: @anthonykhoa good point about `async.

also your sample code is wrong, `try/catch` does not work for promises. You would have to do:

```js
return prisma.module.findMany(...).catch(...)
```

We could return empty array on errors so we always guarantee returning an array:

```js
return prisma.module.findMany(...).catch( _e => return [])
```

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,30 @@
...
+  return prisma.module.findMany({
+    include: {
+      author: true,
+      lesson: true
+    }
+  })
+}
+
+export const addModule = async (_parent: void, args: MutationAddModuleArgs) => {
+  const { authorId, content, lessonId, name } = args
```
songz, line 14: we shouldn't have `authorId` be passed in, it should be retrieved from the cookie.

graphql/queryResolvers/moduleCrud.ts
songz: if this returns an error the front end will get an error, which could render a message. What do you suggest we do in the `catch` @anthonykhoa ?

graphql/queryResolvers/moduleCrud.ts
songz: what do you suggest we do in the `catch` block @anthonykhoa ?

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: Ah i see that makes sense. In this case I think we should only log the error then.
Something that tells us creation of module failed, for example `MODULE_CREATION_FAILED_ARGS:${JSON.stringify(args)}`

anthonykhoa: Hmm wait hold on

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: Log the error. Incase anything goes wrong, logs would be helpful in issue triaging.

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: Ah good idea. returning empty array is appropriate for this scenario.

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: If we don't want to return an error to the frontend, I guess we can return some default value like `{}`.

But I think we should throw the error. And then we handle the error from the frontend too. If there is an error thrown in this resolver, we can have error notifications popup on the screen saying something like `Module Creation Failed`.

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: @mino323 See comment thread on the `createModule` function for the discussion about error handling

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: Is authorId the same as userId?
If we can assume the user will always be logged in when calling this resolver, and if `authorId=userId`, then we actually have it in `req.session` already:
https://github.com/garageScript/c0d3-app/blob/4b140cb4926a4c92ffcbc782a972e44b64185fc2/helpers/controllers/passwordController.ts#L101

If we can't assume the user will be logged on while making a request so this resolver, then

anthonykhoa: If authorId=userId, then we set in in `req.session` already
https://github.com/garageScript/c0d3-app/blob/4b140cb4926a4c92ffcbc782a972e44b64185fc2/helpers/controllers/authController.ts#L54

anthonykhoa: Hmm I don't think authorId=userId, so that probably won't work.

anthonykhoa:
Is it a requirement that user MUST be logged in order to do these crud operations for modules? If so, then we should include a check to make sure users are logged in these resolvers

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: Apollo already provides `onError` options with their useMutation and useQuery hooks. 
https://www.apollographql.com/docs/react/data/error-handling/

PR #1398, opened 2022-02-02
Fix Challenge description sluggishness

### LINKED ISSUE #1395: Fix Challenge description accordion sluggishness
It's sluggish because it moves all the elements below it. One tried solution is to make it absolute and the container that has it and `diffView` - relative.

flacial:
Close #1395 

The accordion is sluggish because when it opens, it pushes all the elements below it. The solution in this PR is to make it absolute and set its `z-index` to a high value.

Before:

https://user-images.githubusercontent.com/35906419/152234797-7134582c-bd2c-426e-abf3-0a7b2cf4537a.mp4

After:

https://user-images.githubusercontent.com/35906419/152234802-01538bae-b197-427a-94b4-e615a30d81ac.mp4

### DIFF
--- a/components/ReviewCard.test.js
+++ b/components/ReviewCard.test.js
@@ -334,4 +334,21 @@ describe('ReviewCard Component', () => {
       )
     ).toBeVisible()
   })
+  test('Should render Accordion with shadow when Accordion header is clicked', () => {
+    const { container, getByText } = render(
+      <MockedProvider mocks={mocks} addTypeName={false}>
+        <ReviewCard
+          submissionData={submissionData}
+          session={dummySessionData}
+        />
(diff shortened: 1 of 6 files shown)


songz (approved):

components/ReviewCard.test.js
```diff
@@ -334,4 +334,21 @@ describe('ReviewCard Component', () => {
       )
     ).toBeVisible()
   })
+  test('Should render Accordion with shadow when Accordion header is clicked', () => {
```
songz, line 337: description should match the tests: `Accordion should be visible when accordion header is clicked`

SahilKalra98 (approved):

PR #1404, opened 2022-02-03
Fixes StoryBook errors

### LINKED ISSUE #1374: Broken Storybook
Example link: https://deploy-preview-1372--youthful-thompson-448332.netlify.app/?path=/story/components-reviewcard--active-card
Screen Shot 2022-01-30 at 1 33 16 AM

SahilKalra98:
Closes #1374 

**Problem**

The getPreviousSubmissions query was not provided in the mocks and that is the reason it displayed a query error. You can see the error [here](https://deploy-preview-1372--youthful-thompson-448332.netlify.app/?path=/story/components-reviewcard--active-card)

**Solution**
I mocked the getPreviousSubmissions query and added to the mocks array which was able to fix the problem. [This](https://medium.com/@emmanuellaaninye/errors-when-testing-apollo-graphql-in-react-no-more-mocked-responses-for-the-query-fe1b54756491) helped me solve the problem and lists several different reasons why this error might come up

**Issues**
After Solving the problem, I was expecting the SelectIteration component to render as well but it didn't.

Expected:
image

Received:
image

Also while looking at the different states of ReviewCard I noticed some of them didn't match the state that it was named. An example was ReviewCard with Long Comment but when I checked the storybook, the comment was not rendered. I checked the mock queries and responses and they seem to match up so I wasn't sure what the problem could be. Here is an example:

image

I would appreciate some feedback! I will label it a WIP in the meantime.

### DIFF
--- a/stories/components/ReviewCard.stories.tsx
+++ b/stories/components/ReviewCard.stories.tsx
@@ -8,47 +8,12 @@ export default {
 import { MockedProvider } from '@apollo/client/testing'
 import ACCEPT_SUBMISSION from '../../graphql/queries/acceptSubmission'
 import REJECT_SUBMISSION from '../../graphql/queries/rejectSubmission'
+import GET_PREVIOUS_SUBMISSIONS from '../../graphql/queries/getPreviousSubmissions'
 import { SubmissionStatus } from '../../graphql'
-const mocks = [
-  {
-    request: {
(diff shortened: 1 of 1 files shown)


songz (approved):

Cijin (approved):

flacial (approved):

songz:
good seeing you again @Cijin !

Issue #893, opened 2021-06-23
Edit/delete comments


ggwadera:
Currently we can only add line comments while reviewing. It would be nice if we could also edit and/or delete a comment.

# Requirements

- [ ] Ability to edit comments
  - [x] Mutation and resolver on the back-end
  - [ ] UI for the front-end
- [x] Ability to delete comments
  - [x] Mutation and resolver on the back-end
  - [x] UI for the front-end
- [x] Only the comment owner must be able to edit/delete a comment
  - [x] Edit/delete options should not be rendered on a comment the user doesn't own
  - [x] The backend should also check if the user owns the comment by the user id on the session.

PR #1428, opened 2022-02-08
Create the resolver to delete comments

flacial:
Related to https://github.com/garageScript/c0d3-app/issues/893

Create the resolver to delete the comments.


songz (approved):

graphql/queryResolvers/deleteComment.test.js
```diff
@@ -0,0 +1,61 @@
...
+        {},
+        {
+          id: 2
+        },
+
+        { req: {} }
+      )
+    } catch (e) {
+      expect(e.message).toEqual('No authorId field')
+    }
```
JasirZaeem, line 44: Instead of catching error and then comparing it, you can use jest's matchers to directly check the error thrown.
e.g.
```javascript
test('should throw error if no user.id in context', async () => {
  await expect(
    deleteComment(
      {},
      {
        id: 2
      },
      { req: {} }
    )
  ).rejects.toThrow('No authorId field');
}
```

JasirZaeem (approved):

PR #1430, opened 2022-02-08
Add theme/Button outline variant

flacial:
Add theme/Button outline variant.

[You can check it here](https://deploy-preview-1430--youthful-thompson-448332.netlify.app/?path=/story/theme-button--outlined-button)


songz:
is the word `outline` standard or did you make it up? It seems more like "hover-link-primary-button" to me because it is only on hover states.

components/theme/colors.tsx
```diff
@@ -1,8 +1,18 @@
-export type ColorTypes = 'white' | 'black' | 'darkgrey' | 'lightgrey'
+import styles from '../../scss/__variables.module.scss'
+
+export type ColorTypes =
```
anthonykhoa, line 3: Since this is going to be used by other people, we should turn this into a typescript enum to export, instead of type

components/theme/colors.tsx
```diff
@@ -1,8 +1,18 @@
...
+  | 'lightgrey'
+  | 'danger'
+  | 'primary'
 
 export const colors = {
   white: '#fff',
   black: '#292929',
   darkgrey: '#5d656e',
-  lightgrey: '#8e8e8e'
+  lightgrey: '#8e8e8e',
```
anthonykhoa, line 15: These `color` constants should also be enums since we are using typescript

anthonykhoa: Oh wait hmm this might require more refactoring, since other files may already be using this colors constant object. Might wanna leave it for another PR.

components/theme/colors.tsx
anthonykhoa: Oh actually I just looked at the codebase and I see `ColorTypes` is just used as purely a type.. so ignore my comment. This is fine

songz:
thanks for clarifying!

songz (approved):

PR #1445, opened 2022-02-10
UpdateCache/helper: add support for deleting submission comment

flacial:
Related to #893 

`UpdateCache` helper will support deleting comments. It'll then be used in `SubmissionComments` to delete a comment when the user click the delete button.

### DIFF
--- a/helpers/updateCache.ts
+++ b/helpers/updateCache.ts
@@ -15,9 +15,10 @@ import _ from 'lodash'
 */
 export const updateCache = (
   submissionId: number,
-  content: string,
-  name: string,
-  username: string,
+  commentToDeleteId?: number,
+  content?: string,
+  name?: string,
+  username?: string,
   lessonId?: number,
   line?: number,
   fileName?: string,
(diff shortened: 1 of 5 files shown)


songz (approved):

helpers/updateCache.ts
```diff
@@ -15,9 +15,10 @@ import _ from 'lodash'
 */
 export const updateCache = (
   submissionId: number,
-  content: string,
-  name: string,
-  username: string,
+  commentToDeleteId?: number,
```
anthonykhoa, line 18: I have concerns about `updateCache` function taking in so many arguments. As the codebase expands I can see a possible situation where this `updateCache` . With the addition of `commentToDeleteId`, that makes a total of 11 arguments passed to the updateCache function. I don't think its a good idea to keep stacking these arguments.. what if one day, this function can take up to 20 arguments? It doesn't make sense to me that we pass in `undefined` as an argument value for optional arguments. In terms of readability, I think it would make more sense `not` to use `undefined` when we pass in an optional argument value.
Picture this scenario: Let's say that one day, there is a use case such that we only want to update the ` challengeId` in the apollo cache. In this scenario, we'd have to do something like
```
updateCache(undefined, undefined, undefined, undefined, undefined, undefined, undefined, 4)
```

I propose that updateCache gets refactored so take in an object as an argument, instead of a whole list of all arguments. That way, no one has to pass in `undefined` as argument parameter for optional parameters. I see that for `submissionId`, it is a required argument.
https://github.com/flacial/c0d3-app/blob/9b8703c09bc5f354ee25653d974de9a16f182c11/helpers/updateCache.ts#L17
So what i'm thinkin is that for we have arguments for all required arguments. For anything that is optional, we can put them all into an object

updateCache = (submissionId, fieldsToUpdate) => {}

anthonykhoa: Can someone explain to me why we are specifically updating query `GET_PREVIOUS_SUBMISSIONS` in the updateCache function? Is this the query we always want to update? Will there every be a realistic future use case where we have to update a cache of a different query?

anthonykhoa (approved):
Approved, but we should definitely look into refactoring the

Issue #1324, opened 2022-01-21
Copy all code feature


flacial:
### Feature
Add the ability to copy all the code in the `diffView` component. The feature will make reviewing code easier, since running the code locally will be faster.

### Example
UI: a button on the opposite side of the `diffView` title (e.g., `js1/4.js`)
`js1/4.js ---------------------------------------------------------------- [BUTTON]`
`[------------------------------------- CODE -------------------------------------]`

### Code
Something like this would work:

```js
const newValue = props.newValue;
//...
<Button
  onClick={() => navigator.clipboard.writeText(newValue)}
>
Copy
</Button>
```

### Solved issues
Besides making it easy for the reviewer to paste the code into the code editor, copying `Show Converstation` text by accident will also be solved.

flacial:
Might be closed for https://github.com/garageScript/c0d3-app/issues/1418, but being able to copy each file individually is another nice feature to have.

PR #1447, opened 2022-02-11
Create CopyButton component

flacial:
Related to #1324 

It creates `CopyButton` that its purpose is to copy text to the user's clipboard onClick event. Could also be used in other places like Dojo exercises.

Will be implemented in the next PR.

Demo:
CopyButton demo

### DIFF
--- a/components/CopyButton.tsx
+++ b/components/CopyButton.tsx
@@ -1,0 +1,63 @@
+import React, { useState } from 'react'
+import { Button } from './theme/Button'
+import { AlertIcon, CheckIcon, CopyIcon } from '@primer/octicons-react'
+import * as Sentry from '@sentry/react'
+import { ColorTypes } from './theme/colors'
+
+enum State {
+  NotCopied,
+  Copied,
+  Error
+}
+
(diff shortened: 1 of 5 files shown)


components/CopyButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+  Error
+}
+
+enum ColorType {
+  Primary = 'primary',
+  Info = 'info',
+  Success = 'success'
+}
+
+const CopyButton: React.FC<{ value: any }> = ({ value }: { value: any }) => {
```
anthonykhoa, line 19: We shouldn't use `any` as a typescript definition whenever possible, because using `any` as a type pretty much defeats the purpose of typescript

flacial: Oh shoot. It was `clipboard.write` which accepts `any` value. Will change it to string.

anthonykhoa:
Please include an image of the new component

mino323 (approved):

songz (approved):

anthonykhoa (approved):

Issue #1387, opened 2022-02-01
fix module redundency in typedefs


songz:
2 comments on this PR: https://github.com/garageScript/c0d3-app/pull/1354/files

PR #1449, opened 2022-02-12
Adds module crud

mino323:
This Pr closes https://github.com/garageScript/c0d3-app/issues/1387
I have edited the typedefs to get rid of redundancies. 

Purpose of this PR is to continue the Dojo backend foundation. This Pr adds module crud operations and test. Not including update.  Modules will be the subset of of lessons that exercises will be organized by. 
It adds the module crud to the graphql query resolvers.


graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,44 @@
...
+  _parent: void,
+  arg: MutationDeleteModuleArgs,
+  ctx: Context
+) => {
+  const { req } = ctx
+  isAdminOrThrow(req)
+  const authorId = req.user?.id
+  if (!authorId) throw new Error('No User')
+  const { id } = arg
+  await prisma.module.delete({ where: { id } })
```
flacial, line 42: What if the module to delete didn't exist? (`prisma.module.delete({ where: { id } })` returns null)

Shouldn't it inform with something else than success?

mino323: Graphql will respond with an error

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,44 @@
...
+): Promise<Module> => {
+  const { req } = ctx
+  await isAdminOrThrow(req)
+  const { content, lessonId, name } = args
+  const authorId = req.user?.id
+  if (!authorId) throw new Error('No User')
+  const module = await prisma.module.create({
+    data: { authorId, content, lessonId, name }
+  })
+  return module
```
songz, line 29: your `module` variable seems unnecessary.

```js
return prisma.module.create(...)
```

songz (approved):

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const deleteModule = async (
+  _parent: void,
+  arg: MutationDeleteModuleArgs,
+  ctx: Context
+) => {
```
anthonykhoa, line 36: Looks like this function is missing typescript definition for what it returns

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,44 @@
+import type { MutationDeleteModuleArgs, MutationAddModuleArgs } from '..'
+import prisma from '../../prisma'
+import { Context } from '../../@types/helpers'
+import { isAdminOrThrow } from '../../helpers/isAdmin'
+import type { Module } from '@prisma/client'
+
+export const modules = (): Promise<Module[]> => {
+  return prisma.module.findMany({
```
anthonykhoa, line 8: Do we want to an error to be thrown when findMany throws an error, or we do we want to just return an empty array?

graphql/queryResolvers/moduleCrud.ts
```diff
@@ -0,0 +1,44 @@
...
+  _parent: void,
+  args: MutationAddModuleArgs,
+  ctx: Context
+): Promise<Module> => {
+  const { req } = ctx
+  await isAdminOrThrow(req)
+  const { content, lessonId, name } = args
+  const authorId = req.user?.id
+  if (!authorId) throw new Error('No User')
+  const module = await prisma.module.create({
```
anthonykhoa, line 26: We should catch possible errors and log them. Logs are useful in issue triaging.

flacial (approved):

graphql/queryResolvers/moduleCrud.ts
songz: > would we rather just return an empty array?

I think this hides the array from the end user, who are aspiring engineers. I think its better for them to realize there's an error so they can report it.

> Do we want to an error to be thrown when findMany throws an error

Graphql will automatically catch and log the error, which gets logged into sentry with line number and the exact error. Here's an example:
Screen Shot 2022-02-13 at 9 41 59 AM

So I think we are sufficiently covered

anthonykhoa: Oh wow. So with Sentry, we don't have to do any error logging ourselves?

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: @mino323 I saw you marked this as resolved, but didn't see changes. Is there a reason for wanting to declare the module variable? It seems unnecessary to me

graphql/queryResolvers/moduleCrud.ts
anthonykhoa: Wow that's awesome! Thanks for the context @songz

graphql/queryResolvers/moduleCrud.ts
songz: This is a resolver function so it may be redundant to write typescript return value. Also, the return value is already defined in the typedef. 

None of the other resolvers have output types: https://github.com/garageScript/c0d3-app/blob/master/graphql/queryResolvers/allUsers.ts

anthonykhoa: From my perspective, it makes the function more readable for me when defined with a typescript definition. When I look at the function, I can see exactly what it is returning without having to look into the function's code or without having to refer back to typedefs. Missing typescript definition for what the function returns means we aren't taking advantage of typescript to be more clear in our code. 

Just because we have been doing things this way(`not having typescript return value on resolvers`) doesn't mean its best IMO. For example, no files currently use any kind of `constants file` to pull constant values from, and although we have been doing things this way, I think its way better to make a constants file rather than not have one.

songz: > it makes the function more readable for me when defined with a typescript definition. When I look at the function, I can see exactly what it is returning without having to look into the function's code

100% agree for functions.

> I can see exactly what it is returning without having to look into the function's code or without having to refer back to typedefs.

For resolvers though, it is better to always refer to typedefs as the source of truth, because it is the source of truth. If you have typedefs AND define a return type for the function then you have additional work or making sure they are in sync and potentially run the risk of making them out of sync (therefore errors).

For the reasons above, it is better (and safer) to not have return types in resolvers.

PR #1460, opened 2022-02-14
edits typedefs to include exercises and exerciseCrud

mino323:
This Pr edits the graphql typedef to include exercise and exercise Crud. 
I also ran ```yarn generate``` to update index.tsx
Exercises will be the foundation to the c0d3 dojo.

### DIFF
--- a/graphql/typeDefs.ts
+++ b/graphql/typeDefs.ts
@@ -6,6 +6,7 @@ export default gql`
     session: Session!
     allUsers: [User]
     modules: [Module]!
+    exercises: [Exercise]!
     getLessonMentors(lessonId: Int!): [User]
     userInfo(username: String!): Session
     isTokenValid(cliToken: String!): Boolean!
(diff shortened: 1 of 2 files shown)


graphql/typeDefs.ts
```diff
@@ -57,6 +58,22 @@ export default gql`
     deleteComment(id: Int!): Comment
     addModule(lessonId: Int!, name: String!, content: String!): Module
     deleteModule(id: Int!): SuccessResponse
+    addExercise(
+      moduleId: Int!
+      description: String!
+      answer: String!
+      testable: Boolean!
+      testStr: String
```
anthonykhoa, line 66: What is this field used for? `testStr`

songz: if user wants to add a test to be run:

```js
test('should equal 55', () => {
  expect(inputValue).toEqual(55)
}
```

anthonykhoa: Thanks!

graphql/typeDefs.ts
```diff
@@ -57,6 +58,22 @@ export default gql`
...
+    ): Exercise
```
anthonykhoa, line 67: @mino323 Question. Why is it for `addExercise` we return type `Exercise`, but for ` updateExercise`, we return `Exercise!`?(One uses exclamation point, the other doesn't).

anthonykhoa:
Nice PR description

graphql/typeDefs.ts
mino323: My thinking was when you add an exercise you might not need anything returned. When you update you want to see the updated data.

SahilKalra98: My thinking is whenever you create/update/delete a resource, it should return that resource. Not sure though

flacial: I agree with Karla. In my point of view, it works as a confirmation that this request is successful.

songz: @SahilKalra98 is correct. whenever you create/update a resource, unless there is error you should be returning the resource. I'm not sure about delete though, my memory tells me you simply return the id of the deleted resource (no point returning a data that no longer exists).

mino323: Got it thanks, all!

flacial (approved):

PR #1477, opened 2022-02-16
Fix LessonCard completion badge styles

### LINKED ISSUE #1405: Lesson status badge has incorrect styles on mobile
As the title says. 

image

flacial:
Closes #1405 

This PR force `background: transparent` declaration to be applied with the `!important` flag so it overrides bootstrap background styles.

Before:

image

After:

image

### DIFF
--- a/components/LessonCard.tsx
+++ b/components/LessonCard.tsx
@@ -50,8 +50,7 @@ const ReviewCount: React.FC<ReviewCountProps> = props => {
   const pendingSubmissionsCount = data.submissions.reduce(
     (acc: number, val: any) => {
       if (val.status === SubmissionStatus.Open) {
-        acc = acc + 1
-        return acc
+        return acc + 1
       }
       return acc
     },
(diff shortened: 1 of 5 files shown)


scss/lessonCard.module.scss
```diff
@@ -13,7 +13,7 @@
   padding-right: 1em;
   margin-left: auto;
   @media (max-width: 768px) {
-    background-color: transparent;
+    background-color: transparent !important;
```
songz, line 16: `!important` is bad practice, if you must use it it should be accompanied with a comment explaining why.

songz: did you try removing the `.bg-success` class?
Screen Shot 2022-02-16 at 5 39 51 AM

flacial: Removing it for mobile and tablet fixes it but not desktop. One solution is to separate tablet and desktop styles but I chose to set its default bg to success and change it to transparent when the media query condition is true.

Now the flag isn't needed.

components/LessonCard.tsx
```diff
@@ -50,8 +50,7 @@ const ReviewCount: React.FC<ReviewCountProps> = props => {
   const pendingSubmissionsCount = data.submissions.reduce(
     (acc: number, val: any) => {
       if (val.status === SubmissionStatus.Open) {
-        acc = acc + 1
-        return acc
+        return acc + 1
```
flacial, line 53: Shouldn't have done it but had to 🙂

songz (approved):

songz:
This new solution is much nicer. Good job!

mino323 (approved):

SahilKalra98 (approved):

Issue #1470, opened 2022-02-15
Tech Debt - return correct module information


songz:
Tech debt for pull request that has unresolved comments: https://github.com/garageScript/c0d3-app/pull/1460#discussion_r806368242

PR #1494, opened 2022-02-18
Adds exercise crud

mino323:
This pr closes https://github.com/garageScript/c0d3-app/issues/1470 
The typedefs were updated to make the response a non nullable exercise return for addExercise.

It also changes updateExercise  in typedefs.  It now make arguments non nullable.  This makes graphql throw an error right away instead of checking if every argument is null in the code.  They are non nullable in the db so they must be checked. 

This pr also adds exercise crud and test.
The exercises a building block to the dojo. They will allow students to interactively learn. 

```yarn generate``` was also run to update index.ts. 

This pr also changes delete of modules and exercises to return the deleted item itself.

### CODE REVIEW

File: graphql/queryResolvers/exerciseCrud.ts [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+
+export const updateExercise = async (
+  _parent: void,
+  args: MutationUpdateExerciseArgs,
+  { req }: Context
+): Promise<Exercise> => {
+  const authorId = req.user?.id
+  const { id, testable, testStr, description, answer, moduleId } = args
+  if (testable && !testStr) throw new Error('Testable must have test string')
+  if (!authorId) throw new Error('No user')
```
flacial, line 43: I think this and line `41` should be swapped.

graphql/queryResolvers/exerciseCrud.ts
```diff
@@ -0,0 +1,78 @@
...
+  const exercise = await prisma.exercise.findUnique({
+    where: {
+      id
+    }
+  })
+  if (!isAdmin(req) && exercise?.authorId !== authorId) {
+    throw new Error('Not authorized to delete')
+  }
+  await prisma.exercise.delete({ where: { id } })
+  return { success: true }
```
flacial, line 77: Should return the `id`?

graphql/queryResolvers/exerciseCrud.ts
```diff
@@ -0,0 +1,78 @@
...
+  MutationUpdateExerciseArgs,
+  MutationDeleteExerciseArgs,
+  SuccessResponse
+} from '..'
+import { Context } from '../../@types/helpers'
+import type { Exercise } from '@prisma/client'
+import { isAdmin } from '../../helpers/isAdmin'
+
+export const exercises = async () => {
+  return await prisma.exercise.findMany({
```
songz, line 13: no need to `return await` just return the promise itself. `return prisma.exercise.findMany....` And take out `async`

graphql/queryResolvers/exerciseCrud.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const updateExercise = async (
+  _parent: void,
+  args: MutationUpdateExerciseArgs,
+  { req }: Context
+): Promise<Exercise> => {
+  const authorId = req.user?.id
+  const { id, testable, testStr, description, answer, moduleId } = args
+  if (testable && !testStr) throw new Error('Testable must have test string')
```
songz, line 42: on second thought, we don't need the `testable` field because we could just check `testStr` is present right (I think this would require a db migration), but db migration is better to be done earlier than later.

songz (approved):

flacial (approved):

PR #1495, opened 2022-02-18
drops column testable

mino323:
This pr drops the testable column in the db. The testable column was redundant. Instead we will just check if there is a testStr.

### DIFF
--- a/prisma/migrations/20220218161843_deletes_testable/migration.sql
+++ b/prisma/migrations/20220218161843_deletes_testable/migration.sql
@@ -1,0 +1,8 @@
+/*
+  Warnings:
+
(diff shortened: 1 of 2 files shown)


flacial (approved):

songz (approved):

songz (changes requested):
could you create another migration to fix your typo

prisma/migrations/20220218172433_adds_explaination/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "exercises" ADD COLUMN     "explaination" TEXT;
```
songz, line 2: `explanation`

SahilKalra98 (approved):

PR #1065, opened 2021-08-24
[WIP] Add NextAuth for Discord OAuth integration

ggwadera:
Initial (very ugly) example for NextAuth implementation, using Discord's OAuth for login and also email magic link.

Included a `/session` page which displays the data from the `useSession()` hook from next-auth.

PR #1497, opened 2022-02-18
[WIP] Adds next-auth for Discord Oauth

SahilKalra98:
This PR looks into using next-auth for Discord Oauth. I used [this](https://github.com/garageScript/c0d3-app/pull/1065) PR for reference and to get started I followed the setup guide provide by next-auth at this [link](https://next-auth.js.org/getting-started/example)

**Achieved so Far**
I created a session page so you can see what type of data gets returned from the discord signin. Here is before and after signing with Discord.

Before:
image

After:
image

I was also able to connect the Prisma Schema with the Discord provider following the instructions [here](https://next-auth.js.org/adapters/prisma) and for adding the Discord provider I followed the steps [here](https://next-auth.js.org/providers/discord). For the schema I had to modify the Users and Session table. A new Accounts table was added in that keeps track of the next-auth providers (in this case Discord). This however breaks the current logging with username and password.

**SignIn Flow**
When you click on the Sign In button on the session page, it takes you to next-auth's already built in sign in page:

image

Then Discord will ask if it is okay for the authorization:
image

Then it will take you back to the session page with your data displayed shown above. During the process next-auth looks up the user's id in the database and if it can't find the specific id, it will look for the email. If it doesn't find an email, then prisma will create the user. It will then proceed to create the Account in the Accounts table and link it to the User. After the account has been linked up, the session is created for the User.

**Issues/Concerns**
I am not sure what to do next because next-auth provides a session and we currently are using express-session to create a session. With the changes in the PR the current Login Page with username and password doesn't work. Any feedback will be appreciated on how to better integrate next-auth!

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.19",
+    "@next-auth/prisma-adapter": "^1.0.1",
     "@primer/octicons-react": "^16.3.1",
     "@prisma/client": "^2.29.1",
     "@quixo3/prisma-session-store": "^3.1.3",
@@ -53,6 +54,7 @@
     "markdown-to-jsx": "^7.1.6",
     "nanoid": "^3.3.1",
     "next": "12",
+    "next-auth": "^4.2.1",
     "next-connect": "^0.12.1",
     "next-mdx-remote": "^3.0.8",
     "nodemailer": "^6.7.2",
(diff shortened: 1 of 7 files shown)


flacial:
I think if next-auth would make it easier to integrate with other login services such as GitHub in the future and Discord currently, we should let next-auth handle it all include signing with username and password. 

Amay pointed there would be database issues though. I don't know much about database, so I can't tell if it's easy to migrate.

PR #1507, opened 2022-02-20
Fix: empty comment get submitted

### LINKED ISSUE #1455: mdInput submits empty comment text
The `mdInput` in the review page allow empty text to be submitted as comment.

To replicate:

1. View any submission
2. Put the feedback type to comment
3. Submit without any text in the `mdInput`

### Solution

Should check if the `textarea` has a value, if it doesn't, it shouldn't submit it.

flacial:
Closes #1455 

This PR updates `ChallengeMaterial` and `ReviewCard` to check if the comment value isn't empty before adding the comment.

### DIFF
--- a/components/ReviewCard.test.js
+++ b/components/ReviewCard.test.js
@@ -14,6 +14,11 @@ import { ContextProvider, GlobalContext } from '../helpers/globalContext'
 import dummySessionData from '../__dummy__/sessionData'
 import previousSubmissionsData from '../__dummy__/getPreviousSubmissionsData'
 import _ from 'lodash'
+import REJECT_SUBMISSION from '../graphql/queries/rejectSubmission'
+
+jest.mock('../helpers/updateCache')
+import { updateCache } from '../helpers/updateCache'
+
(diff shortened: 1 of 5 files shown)


components/ReviewCard.test.js
```diff
@@ -247,6 +304,9 @@ describe('ReviewCard Component', () => {
     )
   })
   test('Should be able to add comment', async () => {
+    const updateCacheMock = jest.fn()
+    updateCache.mockImplementation(() => updateCacheMock())
```
flacial, line 308: This has be done because UpdateCache tries to get the cache but there's no cache: (Error: `No Apollo Cache found)`

flacial: I couldn't fix even after passing a new cache to the component. Any feedbacks are appreciated.

songz (approved):

components/ReviewCard.test.js
songz: not sure!

SahilKalra98 (approved):

PR #1516, opened 2022-02-22
Adds ability to collapse Submission Files

### LINKED ISSUE #1420: Ability to collapse file/submission that has been viewed/reviewed
Just like GitHub PR `Viewed button` or the left arrow:

image
image

Imagine there's 4 - 400+ lines submissions. You reviewed the first one by just providing comments and didn't settle on a final feedback (reject or accept). You'll definitely want it to be collapsed, as it'll be hard to navigate through the rest of the submissions

SahilKalra98:
**Closes #1420**

I used the checkbox element from bootstrap used [here](https://getbootstrap.com/docs/5.0/forms/checks-radios/#checks)
Added a state (viewableStates) that keeps track when a diff is viewed or not for each file in a submission and renders accordingly. Below are the differences in collapsed and not collapsed states.

**IsViewable is true**
image

**IsViewable is false**
image

If you guys have any styling suggestions to make it look better, let me know!

### DIFF
--- a/components/DiffView.tsx
+++ b/components/DiffView.tsx
@@ -40,7 +40,9 @@ const DiffView: React.FC<{
   type fileComments = Record<string, { lines: number[]; comments: Comment[] }>
   //every file gets unique index in format of submissionId:fileName
   const [commentsState, setCommentsState] = React.useState<fileComments>({})
-  const [isViewable, setIsViewable] = React.useState(false)
+  const [viewableStates, setViewableStates] = React.useState(
+    new Array(files.length).fill(false)
+  )
 
   useEffect(() => {
     const commentsMap =
@@ -98,16 +100,18 @@ const DiffView: React.FC<{
 
     return (
       <div className="position-relative">
-        <div className="position-absolute w-100 d-flex justify-content-end p-1">
-          <div className="form-check p-2">
+        <div className="position-absolute w-100 d-flex flex-row justify-content-end p-1">
+          <div className="form-check mt-2 me-2">
             <input
               className="form-check-input"
               type="checkbox"
               value=""
               id={`checkBox-${id}-${fileIdx}`}
-              checked={isViewable}
+              checked={viewableStates[fileIdx]}
               onChange={() => {
-                setIsViewable(!isViewable)
+                const newViewableStates = [...viewableStates]
+                newViewableStates[fileIdx] = !newViewableStates[fileIdx]
+                setViewableStates(newViewableStates)
               }}
             ></input>
             <label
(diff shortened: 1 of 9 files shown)


flacial:
After adding the viewed checkbox, the copy button seems off or placed in a random place. Maybe some container around them would be sufficient with a separator between them. 

`[checkbox | toolbar buttons]`

flacial:
Style suggestion: A container around the viewed text and checkbox would give the idea that they're a single element or grouped together.

flacial:
Style suggestion: The `viewed` text colour is black which is similar to the filename. Maybe a lighter colour (dark) would be fine to make it appear less important.

flacial (approved):

components/DiffView.tsx
```diff
@@ -93,11 +94,29 @@ const DiffView: React.FC<{
     return (
       <div className="position-relative">
         <div className="position-absolute w-100 d-flex justify-content-end p-1">
+          <div className="form-check p-2">
+            <input
+              className="form-check-input"
+              type="checkbox"
+              value=""
+              id={`checkBox-${id}-${fileIdx}`}
```
songz, line 102: question why do we need `fileIdx` here? 

Looking at just the code, a `uuid` would have worked too right? or `Math.rand()*Date.now()`

songz (approved):

components/DiffView.tsx
```diff
@@ -35,6 +35,7 @@ const DiffView: React.FC<{
   type fileComments = Record<string, { lines: number[]; comments: Comment[] }>
   //every file gets unique index in format of submissionId:fileName
   const [commentsState, setCommentsState] = React.useState<fileComments>({})
+  const [isViewable, setIsViewable] = React.useState(false)
```
flacial, line 38: There's an issue with this way. When I try to set a single file to `viewed`, all the other files get hidden as well.

image

If seems like all `DiffView` components share this single state instead of one for each `DiffView`.

flacial: You can try to replicate it on https://c0d3-app-qb075zyys-c0d3.vercel.app/review/js5 if the submission still exists

flacial (changes requested):
Each `DiffView` should have its own state rather than sharing a single state between all of them.

flacial (approved):

components/DiffView.tsx
SahilKalra98: Yeah that would have worked too. Now looking at the bug that flacial stated below, I used fileIdx to map each of the files viewableStates. I'm not sure if that was a good solution or not. 

I can definitely change the id above to `checkBox-${id}-${Math.rand()*Date.now()}`. 

I thought it would make more sense if it was mapped to the submissionId and then the id of the file for each diff i.e. `checkbox-100-0`, `checkbox-100-1 etc`

components/DiffView.tsx
SahilKalra98: I changed the state to a map of booleans which represents the viewable state for each file. Here is the example below:

image

Then I update the state accordingly based on the fileIdx for each submission. Let me know if you have any suggestions on how to make it better!

flacial: One issue appeared, when you click the checkbox, the background color of the filename container gets reduced to only fill the least space possible which is the filename text since the `newValue` isn't there anymore.

image

Note: changed the bg color to green to make it obvious

anthonykhoa (approved):

PR #1524, opened 2022-02-25
All right reserved added for preventing plagiarism.

shockz09:
I am adding this as it is important , many writes just copy and paste the content from any other source, the thing is taking idea/inspiration is good but most people don't take ideas , they just copy and paste.

### DIFF
--- a/__tests__/__snapshots__/storyshots.test.js.snap
+++ b/__tests__/__snapshots__/storyshots.test.js.snap
@@ -4863,7 +4863,7 @@ Array [
   <footer
     className="mt-5 fw-light text-center pb-5 text-muted"
   >
-    © Copyright 2022 GarageScript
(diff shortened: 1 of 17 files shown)


songz:
Adding `all right reserved` seems to have no extra meaning to the copy right: https://www.legalzoom.com/articles/the-term-all-rights-reserved-explained

| This statement is not legally required, and failure to include it has no legal significance.

Since this project is open sourced though, `no rights reserved` actually makes sense: 

| If you include the statement "no rights reserved," it means that you are permitting anyone else to use your work. This is occasionally done when the work is being published for some charitable or public-interest purpose, and the creator wants it to be available for general use by others.

We don't mind if someone forks our repo and builds a better version of c0d3

shockz09:
yeah

PR #1553, opened 2022-03-07
Mitigate the issue of creating a submission based on the cliToken ID

### LINKED ISSUE #1551: Security - create submission
After I login on c0d3 cli using `c0d3 login`:

1. print out `~/.c0d3/credentials.json`
2. base64 decode it. Notice that it is an object with id and token.

Screen Shot 2022-03-06 at 7 05 19 AM

Looking at our [mutation for creating submission](https://github.com/garageScript/c0d3-app/blob/master/helpers/controllers/submissionController.ts#L23), it seems like we are just decoding the `cliToken` and getting the `id` field to use as user's id.

Problem: So if I was a malicious user, I could simply change the id in the file and then submit as any user I want. 

Here's how I would do it.

1. Base64 encode: `{id: <hackedUserId>}` - Seems like `cliToken` isn't even used so we don't even need that.
2. Paste the base64 string into `~/.c0d3/credentials.json`
3. Submit. 

----

Proposal:

1. In my experience, tokens should be sent in the request header. Perhaps similar to chatroom jwt auth in [js5](https://www.c0d3.com/curriculum/js5). `Bearer` Authorization type seems reasonable because [bearer token](https://oauth.net/2/bearer-tokens/) means `an opaque string, not intended to have any meaning to clients using it`. 
2. In our [user middleware](https://github.com/garageScript/c0d3-app/blob/master/helpers/middleware/user.ts#L17), check to see if user's `Authorization` Header is present. If so, lookup user by the [`cliToken` field](https://github.com/garageScript/c0d3-app/blob/master/prisma/schema.prisma#L142), which [should be indexed](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#defining-an-index) for fast lookups, and populate `req.user` object based on the `cliToken`. 
3. remove `cliToken` check from [create submission](https://github.com/garageScript/c0d3-app/blob/master/helpers/controllers/submissionController.ts#L23) and simply use `req.user` object.

--- 

Other thoughts

We ideally also want to remove `cliToken` from [createSubmission typeDef](https://github.com/garageScript/c0d3-app/blob/master/graphql/typeDefs.ts#L47) but that may break all existing c0d3 CLI. 

Perhaps we could create an announcement, giving 1 week to upgrade? 

Maybe we should have [this issue](https://github.com/garageScript/c0d3-app/issues/1552) looked into as well so we are better prepared for potential CLI breakage in the future.

flacial:
Closes #1551 

This PR will:
- Populate the user object based on the cliToken if it's available
- Retrieve the user ID from the req.user object rather than the `cliToken` but keep the `cliToken` to keep it compatible with users who have older CLI versions.
- Send an event to Sentry with the user's username and ID when `cliToken` exists and `req.user` doesn't

### DIFF
--- a/helpers/controllers/submissionController.ts
+++ b/helpers/controllers/submissionController.ts
@@ -12,6 +12,7 @@ import prisma from '../../prisma'
 import { hasPassedLesson } from '../hasPassedLesson'
 import { updateSubmission } from '../updateSubmission'
 import { sendSubmissionNotification, IdType } from '../discordBot'
+import { decode } from '../encoding'
 
 export const createSubmission = async (
   _parent: void,
@@ -19,10 +20,12 @@ export const createSubmission = async (
   ctx: Context
 ): Promise<CreateSubmissionMutation['createSubmission']> => {
   const { req } = ctx
-  const { id } = req.user!
 
   if (!args) throw new Error('Invalid args')
-  const { challengeId, diff, lessonId } = args
+  const { challengeId, cliToken, diff, lessonId } = args
+
+  const id = !req.user || !req.user.id ? decode(cliToken).id : req.user.id
+
   const previousSubmission = await prisma.submission.findFirst({
     where: {
       challengeId,
--- a/helpers/controllers/submissionController.ts
+++ b/helpers/controllers/submissionController.ts
@@ -26,10 +26,10 @@ export const createSubmission = async (
   if (!args) throw new Error('Invalid args')
   const { challengeId, cliToken, diff, lessonId } = args
 
-  const decodedCliToken: CliToken = cliToken && decode(cliToken)
-  let id = !req.user ? decodedCliToken.id : req.user.id
+  let id = req.user!?.id
 
-  if (!req.user && cliToken) {
+  if (!id && cliToken) {
+    const decodedCliToken: CliToken = cliToken && decode(cliToken)
     const user = await prisma.user.findFirst({
       where: { cliToken: decodedCliToken.cliToken }
     })
--- a/helpers/controllers/submissionController.ts
+++ b/helpers/controllers/submissionController.ts
@@ -27,12 +27,14 @@ export const createSubmission = async (
   const { challengeId, cliToken, diff, lessonId } = args
 
   const decodedCliToken: CliToken = cliToken && decode(cliToken)
-  const id = !req.user || !req.user.id ? decodedCliToken.id : req.user.id
+  let id = !req.user ? decodedCliToken.id : req.user.id
 
   if (!req.user && cliToken) {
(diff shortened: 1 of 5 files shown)


songz (approved):

SahilKalra98 (approved):

helpers/controllers/submissionController.ts
```diff
@@ -9,18 +9,36 @@ import {
...
+  ctx: Context
 ): Promise<CreateSubmissionMutation['createSubmission']> => {
+  const { req } = ctx
+
   if (!args) throw new Error('Invalid args')
   const { challengeId, cliToken, diff, lessonId } = args
-  const { id } = decode(cliToken)
+
+  const decodedCliToken: CliToken = cliToken && decode(cliToken)
+  const id = !req.user || !req.user.id ? decodedCliToken.id : req.user.id
```
songz, line 30: same security bug. `decodedCliToken.id` should be 

```js
const user = await prisma.user.findFirst({
      where: { cliToken: decodedCliToken.cliToken }
    })
    
 id = user.id
```

songz (changes requested):

helpers/controllers/submissionController.ts
```diff
@@ -9,18 +9,38 @@ import {
...
+  ctx: Context
 ): Promise<CreateSubmissionMutation['createSubmission']> => {
+  const { req } = ctx
+
   if (!args) throw new Error('Invalid args')
   const { challengeId, cliToken, diff, lessonId } = args
-  const { id } = decode(cliToken)
+
+  const decodedCliToken: CliToken = cliToken && decode(cliToken)
+  let id = !req.user ? decodedCliToken.id : req.user.id
```
songz, line 30: bad coding practice to have double negatives: `!req.user` unless you absolutely have to, which is not the case here.

```js
let id = _.get(req, 'user.id')
if (!id && cliToken) {
  const user = ...
  id = user!?.id
}
```

helpers/controllers/submissionController.ts
```diff
@@ -9,18 +9,38 @@ import {
...
+  args: MutationCreateSubmissionArgs,
+  ctx: Context
 ): Promise<CreateSubmissionMutation['createSubmission']> => {
+  const { req } = ctx
+
   if (!args) throw new Error('Invalid args')
   const { challengeId, cliToken, diff, lessonId } = args
-  const { id } = decode(cliToken)
+
+  const decodedCliToken: CliToken = cliToken && decode(cliToken)
```
songz, line 29: move this inside the `if (!id && cliToken)` check below, its not needed until then.

songz (approved):

PR #1577, opened 2022-03-12
Improve DiffViewer styles

### LINKED ISSUE #1576: Make DiffViewer round
An attempt to modernize the DiffViewer design by making it round.

Something like:

image

flacial:
Closes #1576 

This PR will:

- Increase the borders radius
- Add margin between each file (diffview)
- Change the Accordion border color to make it similar to the DiffViewer header border (that has the file name)

Before:

image

image

After:

image

image

### DIFF
--- a/scss/reviewCard.module.scss
+++ b/scss/reviewCard.module.scss
@@ -1,16 +1,16 @@
-.descAccordion {
+.card__descAccordion {
   z-index: 3;
   top: 0;
   width: 100%;
   position: absolute;
 }
 
-.accordionItemShow {
+.accordion__accordionItem--show {
   box-shadow: 0px -3px 50px -6px #0f235a96;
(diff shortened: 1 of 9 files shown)


SahilKalra98 (approved):

scss/reviewCard.module.scss
```diff
@@ -5,6 +5,12 @@
   position: absolute;
 }
 
-.accordionItem {
+.accordionItemShow {
```
songz, line 8: best practice is to name css using BEM: http://getbem.com/introduction/

songz (approved):

scss/reviewCard.module.scss
flacial: I'll look into it and update the names.

PR #1606, opened 2022-03-17
Updates module crud

mino323:
This  pr updates the module crud to include an update. 
This was required with the new admin page
Screen Shot 2022-03-12 at 4 41 59 PM
.
Admin can now update modules.

### DIFF
--- a/graphql/typeDefs.ts
+++ b/graphql/typeDefs.ts
@@ -57,6 +57,12 @@ export default gql`
     ): Comment
     deleteComment(id: Int!): Comment
     addModule(lessonId: Int!, name: String!, content: String!): Module!
+    updateModule(
+      id: Int!
+      lessonId: Int!
+      name: String!
+      content: String!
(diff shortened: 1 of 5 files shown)


graphql/typeDefs.ts
```diff
@@ -57,6 +57,12 @@ export default gql`
     ): Comment
     deleteComment(id: Int!): Comment
     addModule(lessonId: Int!, name: String!, content: String!): Module!
+    updateModule(
+      id: Int!
+      lessonId: Int!
+      name: String!
```
flacial, line 63: I think `title` instead of `name` is more suitable.

flacial (approved):

graphql/typeDefs.ts
mino323: I want to keep it consistent with addModule.

SahilKalra98 (approved):

PR #1616, opened 2022-03-18
Module crud for front end module Admin page

mino323:
This Pr continues to add crud mutations for the module admin page.  The module page is another step in completing the admin update for the dojo.

The admin module page components will be able to use these mutations for basic crud with the modules. 
They are in the query folder with the rest of the front end queries. 

I updated the the addMoldule query to send more info of the created module. This will make updating useState in the front end easier.

`yarn generate` was also run to give props types and useMutation functions for the queries.


graphql/queries/addModule.ts
```diff
@@ -4,6 +4,11 @@ const ADD_MODULE = gql`
   mutation addModule($content: String!, $lessonId: Int!, $name: String!) {
     addModule(content: $content, lessonId: $lessonId, name: $name) {
       id
+      name
```
flacial, line 7: We already have these in the frontend data the moment the mutation `addModule` is ran. So, I imagine something like this:

```js
const { data, error } = addModule({ content, lessonId, name })

if (data && data?.id) setState(content, lessonId, name)

if (error) showError(true)
```

SahilKalra98 (approved):

songz (approved):

PR #1617, opened 2022-03-19
Updates formValidation and adminHelpers

mino323:
This pr update formValidation and adminHelpers to include module related information. 
This will allow future the admin module page components to use the form components. 
Admin module page is the next step in https://github.com/garageScript/c0d3-app/issues/1192
FormValidations will now have module schema to error check module information. 
adminHelpers now changes 'lessonId' in a form, from a string  to a int which is require for a graphql mutation.

adminHelper test were updated to include 'lessonId'


helpers/formValidation.tsx
```diff
@@ -37,6 +37,15 @@ const lessonSchema = Yup.object({
   chatUrl: Yup.string()
 })
 
+const moduleSchema = Yup.object({
```
flacial, line 40: `moduleSchemaValidation` to match the file naming pattern of `name+vValidation`?

helpers/admin/adminHelpers.ts
```diff
@@ -58,6 +58,10 @@ export const makeGraphqlVariable = (options: any, addProp?: any) => {
     res.id = parseInt(res.id ? res.id + '' : '')
   }
 
+  if (res.hasOwnProperty('lessonId')) {
```
songz, line 61: Question: How does `challenges` work? DOesn't `challenge` also have a field for `lessonId`?

flacial (approved):

helpers/admin/adminHelpers.ts
mino323: It doesn't ask for the lesson. When you click a lesson all the challenges currently come with it. When you add a challenge you are on the lesson page already.

helpers/formValidation.tsx
songz: yeah good idea to keep things consistent.

SahilKalra98 (approved):

Issue #1508, opened 2022-02-20
DOJO - Create Admin pages components


flacial:
Create Admin Page by following the [design](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4486).

### Components to create and update

- [ ] #1709
<details>
  <summary>AppNav</summary>
  
![image](https://user-images.githubusercontent.com/35906419/156900641-2be9ff3f-6ab2-49ed-826c-15b79d7c7e14.png)
Info: has two variants, as Admin and not Admin
</details>

- [ ]  Create web spider tree thing (help with name)
<details>
  <summary>web spider tree thing (help with name)</summary>
  
![image](https://user-images.githubusercontent.com/35906419/156900508-baa0f319-7b43-4c56-8cec-c0fb140caeb2.png)
Find it on: https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=107%3A831
</details>

- [x]  Create Admin side lessons TOC
<details>
  <summary>Admin side lessons TOC</summary>
  
![image](https://user-images.githubusercontent.com/35906419/156900578-b62a6fcb-d3ca-4223-b5d5-9bbd9b4f03f7.png)
</details>

- [x]  Create Admin tabs bar
<details>
  <summary>Admin tabs bar</summary>
  
![image](https://user-images.githubusercontent.com/35906419/156900807-41495954-b072-4497-bf35-d2a3a017bb55.png)
</details>

- [ ]  Create Input fields
<details>
  <summary>Input fields</summary>
  
![image](https://user-images.githubusercontent.com/35906419/156900958-4d079563-6f0e-446f-b34a-95c9a8700d17.png)
</details>

flacial:
Closing this issue in favor of the following:

https://github.com/garageScript/c0d3-app/issues/1733
https://github.com/garageScript/c0d3-app/issues/1734
https://github.com/garageScript/c0d3-app/issues/1735
https://github.com/garageScript/c0d3-app/issues/1736

PR #1637, opened 2022-03-23
Admin: Create AdminLessonNav component

flacial:
Related to #1508 

This PR create the component that's used to switch the tabs (e.g, `INTRODUCTION`, `MODULES`...etc).

Image:

image

[Storybook](https://deploy-preview-1637--youthful-thompson-448332.netlify.app/?path=/story/components-adminlessonnav--basic)


scss/adminLessonNav.module.scss
```diff
@@ -0,0 +1,98 @@
...
+
+.lessons__tabsNav__nav {
+  display: flex;
+  column-gap: 20px;
+}
+
+.lessons_tabsNav__nav__item {
+  padding: 7px 16px 7px 16px;
+  border-radius: 4px !important;
+  background-color: variables.$primary !important;
```
flacial, line 26: Used `!important` to override bootstrap styles. I'm aware it's bad practice but it was necessary.

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,71 @@
...
+        <Tab.Pane key={index} eventKey={item.value}>
+          {tabs[index]()}
+        </Tab.Pane>
+      )
+
+      return acc
+    }, [] as JSX.Element[])
+  }
+
+  const setActiveKey = (k: any) => setKey(k)
```
anthonykhoa, line 55: What is the type of parameter `k`?

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,71 @@
...
+          {toUpper(item.value)}
+        </Nav.Link>
+      </Nav.Item>
+    ))
+
+  const renderTabPanes = (
+    items: { value: string; children?: string[] }[],
+    tabs: (() => JSX.Element)[]
+  ) => {
+    return items.reduce((acc, item, index) => {
```
anthonykhoa, line 44: Your code can be made more simple by using `map` here instead of `reduce` because there will be less logic involved
```
items.map((item, index) => {
    return (
            <Tab.Pane key={index} eventKey={item.value}>
              {tabs[index]()}
            </Tab.Pane>
      )
    })
```

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,71 @@
...
+      )
+
+      return acc
+    }, [] as JSX.Element[])
+  }
+
+  const setActiveKey = (k: any) => setKey(k)
+
+  return (
+    <Tab.Container activeKey={key} onSelect={k => setActiveKey(k)}>
```
anthonykhoa, line 58: We should be able to do `onSelect={setActiveKey}`

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,71 @@
...
+            }) as string
+          }
+          eventKey={item.value}
+        >
+          {toUpper(item.value)}
+        </Nav.Link>
+      </Nav.Item>
+    ))
+
+  const renderTabPanes = (
```
anthonykhoa, line 40: `renderItems` as well as this function shouldn't be functions. I don't see them being used more than once in this component, the function isn't passed down as a prop, and I notice these functions only return an array of jsx elements by using some higher order component.

Having them as functions complicates things -- in this context I think the code can be made more simple by just assigning the array of jsx elements to a variable directly, instead of having a function to do so.
```
navItems = items.map((item, index) => (
      <Nav.Item key={index}>
        <Nav.Link
          active={isActive(item.value) as boolean}
          className={
            isActive(item.value, {
              active: styles.lessons_tabsNav__nav__item,
              inactive: styles['lessons_tabsNav__nav__item--inactive']
            }) as string
          }
          eventKey={item.value}
        >
          {toUpper(item.value)}
        </Nav.Link>
      </Nav.Item>
))

tabPanes = items.map((item, index) => (
    <Tab.Pane key={index} eventKey={item.value}>
          {tabs[index]()}
    </Tab.Pane>
))
```

flacial: I agree.

components/admin/lessons/AdminLessonNav.tsx
flacial: `string | null`

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,74 @@
...
+    </Nav.Item>
+  ))
+
+  const tabPanes = navItems.map((item, index) => (
+    <Tab.Pane key={index} eventKey={item.value}>
+      {tabs[index]()}
+    </Tab.Pane>
+  ))
+
+  const setActiveKey = (k: string | null) => setKey(k)
```
anthonykhoa, line 58: What is the purpose of making the `setActiveKey` function? Why not just use `setKey`?

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  return (
+    <Tab.Container activeKey={key!} onSelect={setActiveKey}>
```
anthonykhoa, line 61: What does the `!` mean?

components/admin/lessons/AdminLessonNav.tsx
flacial: Testing leftovers I think. I should remove it.

components/admin/lessons/AdminLessonNav.tsx
flacial: It means the `key` variable won't be null.

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,74 @@
+import { toUpper } from 'lodash'
+import React, { useState } from 'react'
+import { Tab, Col, Nav } from 'react-bootstrap'
+import styles from '../../../scss/adminLessonNav.module.scss'
+
+type AdminLessonNavProps = {
+  navItems: { value: string }[]
```
anthonykhoa, line 7: Can a person pass in an empty array with this type definition?

flacial: They can't I think and they shouldn't as it'll break it. There will be tabs but no navItems to use for switching between tabs.

anthonykhoa: What i'm worried about is that on line 54 we do `tabs[index]()`. If someone is able to pass in an empty array for `tabs`, then `tabs[index]()` will break things. Can we confirm that the type for `tabs` blocks people from being able to pass in an empty array?
https://github.com/garageScript/c0d3-app/pull/1637/files#diff-d8433cac2e9017e63fddfdb72a729f8fe01dcd7945242e577aea97802e299c6dR54

scss/adminLessonNav.module.scss
flacial: Oh, it's possible without it. react-bootstrap components has this `bsPrefix` that reset the component's styles and apply the styles you assign to it.

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const validateProps = ({ navItems, tabs }: AdminLessonNavProps) => {
+  if (navItems.length !== tabs.length)
+    throw new Error(
+      `navItems and tabs should have the same count. navItems: ${navItems.length} -- tabs: ${tabs.length}`
+    )
+
+  const mapObjArrayToValues = navItems.map(item => item.value)
+
+  if (new Set(mapObjArrayToValues).size !== mapObjArrayToValues.length)
```
anthonykhoa, line 19: Instead of
```
const mapObjArrayToValues = navItems.map(item => item.value)
if (new Set(mapObjArrayToValues).size !== mapObjArrayToValues.length)
```
Can you do

```
if (_.uniq(navItems).length !== navItems.length) {
    ................
}
``` 

uniq function is from lodash library
https://lodash.com/docs/4.17.15#uniq

flacial: Sure, that'd better.

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,72 @@
...
+    throw new Error('navItems should have unique items value. No duplicates')
```
anthonykhoa, line 20: What happens when these validation errors get thrown in the front end?

flacial: An error will appear in the page indicating that the way the developer used the component is wrong.

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Tab, Col, Nav } from 'react-bootstrap'
+import styles from '../../../scss/adminLessonNav.module.scss'
+
+type NavItem = { value: string }
+type Tab = () => JSX.Element
+
+type NavItems = [NavItem, ...NavItem[]]
+type TabsType = [Tab, ...Tab[]]
+
+type Props = {
```
anthonykhoa, line 13: What was wrong with the previous type definition for props?
```
type AdminLessonNavProps = {
  navItems: { value: string }[]
  tabs: (() => JSX.Element)[]
}
```

flacial: The current one doesn't allow empty arrays to be passed.

components/admin/lessons/AdminLessonNav.tsx
anthonykhoa: We should use typescript to make sure that navItems and tab props are the same array length
https://stackoverflow.com/questions/62206320/typescript-require-that-two-arrays-be-the-same-length

anthonykhoa: Not sure if its exactly possible(i haven't tried myself), but please try and see if its possible with typescript instead of

flacial: I'll look into it

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -0,0 +1,80 @@
...
+  navItems: NavItems
+  tabs: TabsType
+}
+
+const validateProps = ({ navItems, tabs }: Props) => {
+  if (navItems.length !== tabs.length)
+    throw new Error(
```
songz, line 20: why did you design it this way, where it could be error prone?

I would have done something like 

```js
[{
  tabName: ...,
  tabComponent: ...
  }, ... ]
  ```
  
  This way we don't have to worry about this error right?

flacial: Yeah, this is a better design.

songz (approved):

anthonykhoa (approved):

SahilKalra98 (approved):

Issue #1550, opened 2022-03-06
Next-Auth Integration Design Doc


flacial:
# Introduction

Previously, we completed our [discord integration](https://github.com/garageScript/c0d3-app/issues/843) with c0d3.com. Users are prompted to connect to discord after they log in.

Looking at the previous architecture, we notice quite a few moving parts and a lot of code, which would all be eliminated if we used the reputable library, [next-auth](https://next-auth.js.org/), designed to help with OAuth integrations.

By migrating to next-auth, we would simplify our codebase and pave the way for other partner integrations like Github.

The goal of this document would be to document all the steps necessary to successfully implement next-auth. By the completion of this project (roughly in implementation order)

- Login/Signup with c0d3 would be replaced with next-auth.
- Users would continue to be prompted for "connecting to discord" if they haven't already after they log in. The [architecture](https://github.com/garageScript/c0d3-app/issues/843) would be replaced by next-auth.
- Users with a connected discord account would be able to sign in with discord.

By the end of this integration, our authentication logic would be simplified by Next-Auth, and for the future, it should be easy to add/connect other platforms like Google, Facebook, Github, etc. 

**Important**: After this change, engineers would need to use [useSession](https://next-auth.js.org/getting-started/client#usesession) for React components and [getSession](https://next-auth.js.org/getting-started/client#getsession) for non-react components such as [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props).

Far graphql resolvers, we would continue to use `req.user`.

**Drawbacks**

next-auth doesn't support sending data in the response body, so it won't work well with our CLI as the CLI requires the `cliToken` that's sent with the [login](https://github.com/garageScript/c0d3-app/blob/3d0be4968dcd3a7e616db8c1b796f16f1a6f8207/helpers/controllers/authController.ts#L16) resolver

Solution: Use `next-auth` API for web, keep GraphQL login resolver for CLI

# Architecture & Research

The diagram below shows the general flow of 3 cases:

1. C0D3's username / pw sign in
2. Connecting to Discord in a logged in user
3. Signing in to c0d3.com using Discord account

Below the diagram, documentation links are provided

[...]

PR #1651, opened 2022-03-26
feat(helpers): update signupValidation to validate password

flacial:
Related to #1550 

This PR update `signupValidation` to also validate password. The `signup` resolver will be used to set the user password and ultimately create an account with needing to redirect the user to `/confirm/[token]` page (where the user is asked to set a new password for their account).

### DIFF
--- a/helpers/formValidation.tsx
+++ b/helpers/formValidation.tsx
@@ -60,7 +60,10 @@ const signupValidation = Yup.object({
     .min(TEXT_MIN, `Must be at least ${TEXT_MIN} characters`)
     .max(TEXT_MAX, `Must be ${TEXT_MAX} characters or less`)
     .trim('Leading and trailing value must be alphanumeric character')
-    .required('Required')
+    .required('Required'),
+  password: Yup.string()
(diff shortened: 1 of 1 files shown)


SahilKalra98 (approved):

helpers/formValidation.tsx
```diff
@@ -60,7 +60,10 @@ const signupValidation = Yup.object({
     .min(TEXT_MIN, `Must be at least ${TEXT_MIN} characters`)
     .max(TEXT_MAX, `Must be ${TEXT_MAX} characters or less`)
     .trim('Leading and trailing value must be alphanumeric character')
-    .required('Required')
+    .required('Required'),
```
anthonykhoa, line 63: I notice lots of repeated strings in this file. There should be more constants made to be used in this file, rather than repeating the same strings over and over again

flacial: Do you think it should be done in this PR or another PR?

songz: I think refactoring and feature should be done in different PRs, makes it easier to read

songz (approved):

PR #1683, opened 2022-04-03
sidebar

mino323:
This pr continues is for the admin sidebar in the modules page. 
The props for this component will be sent from the page. 

Screen Shot 2022-04-02 at 11 43 35 AM


components/admin/modules/SideBar.tsx
```diff
@@ -0,0 +1,54 @@
...
+  selectedModule: number
+}
+export const SideBar: React.FC<SideBarProps> = ({
+  modules,
+  selectedModule
+}) => {
+  const moduleListData = [_.cloneDeep(modules) || []]
+  const lessonList = moduleListData[0].map((obj: any, arrIndex: number) => (
+    <>
+      <a
```
flacial, line 18: Anchor element is used to create a hyperlink. So I think if the `href` prop is passed which means it won't redirect the user to any other page. Maybe it should be turned into a `button`.

components/admin/modules/SideBar.test.js
```diff
@@ -0,0 +1,29 @@
...
+    id: 12,
+    name: 'Math II',
+    content: 'How '
+  }
+]
+
+describe('AdminLessonSideBar component', () => {
+  test('Should render lessons', async () => {
+    render(<SideBar modules={moduleData} selectedModule={1} />)
+    render(<SideBar selectedModule={1} />)
```
flacial, line 26: Why is `render` called twice?

components/admin/modules/SideBar.tsx
```diff
@@ -0,0 +1,54 @@
...
+        className={
+          `${styles.sidebar_tabsNav}` +
+          (arrIndex === selectedModule
+            ? ` ${styles.sidebar_tabsNav__active}`
+            : '')
+        }
+        data-toggle="pill"
+        style={{ wordBreak: 'break-word', cursor: 'pointer' }}
+      >
+        <div className="innerText">{obj.name}</div>
```
flacial, line 31: This should be a `span`, `hX` element or `Text` from our [components](https://github.com/garageScript/c0d3-app/blob/2d5961899002c4bba480e5c04971ee8e4b205554/components/theme/Text.tsx). 

OR

Wrap the `{obj.name}` with one of the above elements. E.g., `<div ...> <Text>{obj.name}</Text> </div>`

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+  margin-bottom: 22px;
+}
+
+.sidebar_mainbody {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  padding: 28px 21px;
+
+  position: static;
```
flacial, line 24: `position, left, top, flex, order, flex-grow` should be removed. width and height could turn into:

```css
width: 100%;
max-width: 369px;
height: 100%;
max-height: 460px;
```

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+  width: 327.01px;
+  height: 52px;
+  background: $primary-color;
+  border-radius: 10px;
+  padding-left: 6.12%;
+  display: flex;
+  justify-content: space-between;
+  text-decoration: none;
+  font-family: 'Inter';
```
flacial, line 52: We don't have Inter yet. Please create a PR that [adds](https://nextjs.org/docs/basic-features/font-optimization#usage) the font Inter before merging this PR.

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+  width: 20px;
+  height: 20px;
+  text-align: center;
+  font-weight: 900;
+  transition: transform 600ms;
+  margin-top: -5px;
+}
+
+.sidebar_deleteX {
+  background: none;
```
flacial, line 83: `background: none` overrides all the other `background-*` properties. If you meant to make the background transparent, you do `background-color: transparent`

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+  margin-right: 10px;
+  text-decoration: none;
+  display: none;
+  width: 20px;
+  height: 20px;
+  margin-top: -5px;
```
flacial, line 89: A negative margin might make things confusing.

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+
+  &:hover {
+    filter: brightness(150%);
+  }
+}
+
+.sidebar_tabsNav__active {
+  color: black;
+  background-color: white;
+  border: 1px solid #d8d8d8;
```
flacial, line 99: If it's possible, I would change HEX to HSL because it's more [human readable](https://blog.bitsrc.io/hex-vs-rgb-vs-hsl-what-is-the-best-method-to-set-css-color-property-f45d2debeee)

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+  &:hover {
+    background-color: #ffffff;
+  }
+
+  .arrow {
+    display: none;
+  }
+}
+
+.sidebar_addButton {
```
flacial, line 114: Same as the above. Remove unnecessary properties.

components/admin/modules/SideBar.tsx
```diff
@@ -0,0 +1,54 @@
...
+        data-testid="module-title"
+        role="tab"
+        className={
+          `${styles.sidebar_tabsNav}` +
+          (arrIndex === selectedModule
+            ? ` ${styles.sidebar_tabsNav__active}`
+            : '')
+        }
+        data-toggle="pill"
+        style={{ wordBreak: 'break-word', cursor: 'pointer' }}
```
flacial, line 29: This could go into the `scss` file to keep things consistent.

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+  width: 20px;
+  height: 20px;
+  margin-top: -5px;
+
+  &:hover {
+    filter: brightness(150%);
+  }
+}
+
+.sidebar_tabsNav__active {
```
flacial, line 96: I feel this could be `...--active` because `--` is used to represent modifiers.

Example:
```css
/* Modifier that changes the style of the block */
.btn--orange {} 
.btn--big {}
```

components/admin/modules/SideBar.test.js
```diff
@@ -0,0 +1,29 @@
...
+  },
+  {
+    id: 12,
+    name: 'Math II',
+    content: 'How '
+  }
+]
+
+describe('AdminLessonSideBar component', () => {
+  test('Should render lessons', async () => {
```
flacial, line 24: It's missing `expect.assertions(1)` to make sure the test assertions work.

scss/sideBar.module.scss
```diff
@@ -0,0 +1,144 @@
...
+  font-style: normal;
+  font-weight: 700;
+  font-size: 18px;
+  line-height: 22px;
+  text-align: center;
+  color: #000000;
+  margin-bottom: 22px;
+}
+
+.sidebar_mainbody {
```
flacial, line 18: What if there were 30 modules, would it go into overflow or user could scroll?

flacial:
I feel like the `ADD MODULE` should always be at the bottom

image

It's possible by giving the `ADD MODULE` button the following declaration `margin-top: auto;`

flacial (changes requested):
It's mostly CSS and using the right HTML elements.

components/admin/modules/SideBar.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import _ from 'lodash'
+import { Module } from '../../../graphql/index'
+import styles from '../../../scss/sideBar.module.scss'
+import { XIcon, ChevronRightIcon } from '@primer/octicons-react'
+
+type SideBarProps = {
+  modules: Module[] | []
```
anthonykhoa, line 8: Is the `| []` needed? Cant a person still pass in a empty array with `Module[]`?

components/admin/modules/SideBar.tsx
```diff
@@ -0,0 +1,54 @@
...
+  selectedModule: number
+}
+export const SideBar: React.FC<SideBarProps> = ({
+  modules,
+  selectedModule
+}) => {
+  const moduleListData = [_.cloneDeep(modules) || []]
+  const lessonList = moduleListData[0].map((obj: any, arrIndex: number) => (
```
anthonykhoa, line 16: Using `any` here defeats the purpose of using typescript -- please use the correct type definition

components/admin/modules/SideBar.test.js
mino323: One time without data

flacial: Is it for the test coverage? I feel it could be separated into two tests, one to check if modules are rendered, and one to check the opposite.

Issue #1185, opened 2021-11-28
Refactor constants to their own folder and files


kondanna:
see https://github.com/garageScript/c0d3-app/pull/1179

SlyBouhafs:
@anthonykhoa wrote in discord:

> I looked at the codebase today and see that there's a lot of repeated values in the codebase -- for values that are repeated and for values that have high potential to be repeated in the future, I suggest we turn them into a variable to use as a constant.
Examples:
/curriculum
https://github.com/garageScript/c0d3-app/search?p=1&q=%2F%2Fcurriculum

>Auth headers such as Authorization and content-type(content-type is capitalized differently on different files.. and it works, but it makes our codebase less consistent when we use different capitalizations for the same word)
https://github.com/garageScript/c0d3-app/search?q=content-type
https://github.com/garageScript/c0d3-app/search?q=authorization

>Day and Days
https://github.com/garageScript/c0d3-app/search?q=day
https://github.com/garageScript/c0d3-app/search?q=days

>Errors about session not existing(this file has several examples of a repeated string value Session Error)
https://github.com/garageScript/c0d3-app/blob/master/graphql/resolvers/authController.ts#L100

>No user error(theres more examples than just this file, but its hard to find them all because they use different error messages)
https://github.com/garageScript/c0d3-app/blob/master/graphql/resolvers/moduleCrud.ts#L46

>/assets
https://github.com/garageScript/c0d3-app/search?q=%2F%2Fassets

PR #1710, opened 2022-04-10
New updated AppNav

### LINKED ISSUE #1709: Update AppNav
Update the AppNav component to the new design.

SlyBouhafs:
This PR closes #1709.
Here's a link for a demo: http://164.92.183.234:3002/

Old AppNav

Screen Shot 2022-04-10 at 17 31 10
Screen Shot 2022-04-10 at 17 31 30

New AppNav

Screen Shot 2022-04-10 at 17 32 15
Screen Shot 2022-04-10 at 17 31 50
Screen Shot 2022-04-10 at 17 32 02

### Ux changes:
- Highlight the current active navigation link when a user is under the same path.
- Add an Admin badge for admin users.
- Create a new user menu and move the admin menu inside of it.


components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+type ProfileDropDownMenuProps = {
+  username: string
+  isAdmin: boolean
+}
+
+type CustomToggleProps = {
+  children?: React.ReactNode
+  className?: string | undefined
+  onClick?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {}
```
flacial, line 21: I think it could be `() => void` because `onClick` returns no value

flacial:
image

AFAIK, the hover effect means this element isn't active (clicked state) and could be clicked and then something would happen or change. When `CURRICULUM` is active or we're in `/curriculum`, I don't think it should show the hover effect. Instead, it should keep showing the active style:

image

scss/appNav.module.scss
```diff
@@ -1,30 +1,92 @@
...
 .nav-item {
-  margin-left: 2em;
-  margin-right: 2em;
+  margin-left: 1em;
+  margin-right: 1em;
+  font-size: 14px;
+  font-weight: 700;
+  color: #565656;
+  text-transform: uppercase;
+  background-color: #fbfbfb;
```
flacial, line 26: NIT: could be hsl

flacial:
If I'm in one of the curriculum's subpages, shouldn't the nav item be active?

image

scss/appNav.module.scss
```diff
@@ -1,30 +1,92 @@
...
+
+.nav-link {
+  color: #565656;
+  font-weight: 700;
+  &:hover {
+    background-color: color.change(
+      variables.$primary,
+      $lightness: 90%
+    ) !important;
+    color: variables.$primary !important;
```
anthonykhoa, line 40: is it necessary to use `!important` in your css changes?

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,138 @@
...
+      key={title}
+      bsPrefix={isActive(path)}
+    >
+      {title}
+    </Dropdown.Item>
+  ))
+
+  const buttonMenuToggle = React.forwardRef(
+    (props: CustomToggleProps, ref: React.Ref<HTMLDivElement>) => (
+      <div
```
anthonykhoa, line 71: Can you try
```
<div
        ref={ref}
        {...props}
      >
        {props.children}
</div>
```

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const location = router.asPath
+
+  const fullname = _.get(data, 'userInfo.user.name', '')
+
+  const userInfo: UserInfo = {
+    // 'A' stands for Anonymous, in case user did not put in full name
+    username,
+    firstName: fullname.split(' ')[0] || 'A',
+    lastName: fullname.split(' ')[1] || ' ',
+    discordUserId: data?.userInfo?.user?.discordUserId ?? '',
```
anthonykhoa, line 42: Is there a need to check if these fields are ` null` or `undefined` using the nullish coalescing operator? Seems like unneeded additional complexity -- why not just use the logical OR operator? `||`

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,138 @@
...
+  isAdmin
+}) => {
+  const { data } = useUserInfoQuery({
+    variables: { username },
+    skip: !username
+  })
+  const router = useRouter()
+  const location = router.asPath
+
+  const fullname = _.get(data, 'userInfo.user.name', '')
```
anthonykhoa, line 35: Why do you use lodash `get` here, but elsewhere you use optional chaining?

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,138 @@
...
+import styles from '../scss/profileDropDown.module.scss'
+
+type ProfileDropDownMenuProps = {
+  username: string
+  isAdmin: boolean
+}
+
+type CustomToggleProps = {
+  children?: React.ReactNode
+  className?: string | undefined
```
anthonykhoa, line 20: Is the `| undefined` needed? Since you already use the `?` typescript operator to signify its an optional field

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,138 @@
...
+            <>
+              <Dropdown.Header>ADMIN</Dropdown.Header>
+              {adminDropdownMenu}
+              <Dropdown.Divider />
+            </>
+          )}
+          <Dropdown.Item
+            className={`${styles['dropdown-item']} nav-link `}
+            bsPrefix={isActive(`/profile/${username}`)}
+            href={`/profile/${username}`}
```
anthonykhoa, line 106: There should be a constant made for `/profile`, seems like this string is repeated  a lot in this file

components/ProfileDropdownMenu.tsx
anthonykhoa: I think you're right about returning void, but `event` should still be needed as a parameter since it is a mouse event

components/ProfileDropdownMenu.tsx
songz: Yeah let's be consistent.

scss/appNav.module.scss
songz: yeah let's never use `!important`

components/AppNav.tsx
```diff
@@ -39,63 +32,48 @@ const navItems: NavItem[] = [
...
-      </NavLink>
-
-      <LogoutContainer>
-        <div className={`${styles['light-button']} d-inline`}>
-          <Button border ml="2" type="light">
-            Logout
-          </Button>
-        </div>
-      </LogoutContainer>
+    <div>
```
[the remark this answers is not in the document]
SlyBouhafs: No it isn't, i removed it!

components/ProfileDropdownMenu.tsx
SlyBouhafs: No, it isn't really needed! Thanks!

components/ProfileDropdownMenu.tsx
flacial: Use lodash `get` when you want to return a value other than `undefined` when the key is not found. Optional chaining when the `undefined` value is fine.

components/ProfileDropdownMenu.tsx
SlyBouhafs: Yeah, i think you're right about the added complexity but i get typescript errors when i use the `||` operator.
Screen Shot 2022-04-14 at 20 27 10

songz: maybe `_.get(data, 'userInfo.user.discordUserId', '')`

components/ProfileDropdownMenu.tsx
anthonykhoa: @flacial If we want to display username as empty string instead of undefined, then just add `||`

```
data?.userInfo?.user?.name || ''
```


IMO I prefer to use lodash `get`, but I believe some conversation was had awhile ago that this codebase should use optional chaining instead of `get`. We really shouldn't be mixing the two, its better to be consistent

components/ProfileDropdownMenu.tsx
anthonykhoa: @SlyBouhafs Thats because you're not using optional chaining. If you used optional chaining it should work. Or if you do what @songz suggested it should work

components/ProfileDropdownMenu.tsx
SlyBouhafs: Thanks! Fixed it!

components/AppNav.tsx
```diff
@@ -39,63 +32,48 @@ const navItems: NavItem[] = [
...
-          bsPrefix={styles['dropdown-item']}
-        />
-      )}
     </>
   )
 }
 
-const LoggedInAuthNav: React.FC<{ username: string }> = ({ username }) => {
-  const capitalized = username.charAt(0).toUpperCase() + username.slice(1)
+const LoggedInAuthNav: React.FC<{ username: string; session: any }> = ({
```
anthonykhoa, line 55: Please don't use `any` as a type definition when possible. It defeats the purpose of using typescript

components/AppNav.tsx
```diff
@@ -39,63 +32,48 @@ const navItems: NavItem[] = [
...
+  username,
+  session
+}) => {
+  const isAdmin = _.get(session, 'user.isAdmin', false) as boolean
```
anthonykhoa, line 59: Is type casting to Boolean necessary?

SlyBouhafs: Indeed, It wasn't necessary, the returned type is already a boolean! Thanks

SlyBouhafs:
> image
> 
> AFAIK, the hover effect means this element isn't active (clicked state) and could be clicked and then something would happen or change. When `CURRICULUM` is active or we're in `/curriculum`, I don't think it should show the hover effect. Instead, it should keep showing the active style:
> 
> image

After some thought, i think we should leave it for accessibility, when using tab to navigate, it feels like a good indicator.

scss/appNav.module.scss
SlyBouhafs: For the colors values, i'll leave them as is and fix them once we setup a final color palette.

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,138 @@
...
+      href={path}
+      key={title}
+      bsPrefix={isActive(path)}
+    >
+      {title}
+    </Dropdown.Item>
+  ))
+
+  const buttonMenuToggle = React.forwardRef(
+    (props: CustomToggleProps, ref: React.Ref<HTMLDivElement>) => (
```
anthonykhoa, line 70: Your CustomToggleProps have fields that are already in `React.FC`. Is the typescript definition for `CustomToggleProps` needed?

SlyBouhafs: Sorry, i'm not sure what you mean, the `CustomToggleProps` already uses react typescript definitions except for the classname.

flacial (approved):

scss/navLink.module.scss
```diff
@@ -9,3 +12,8 @@
 .noUnderline {
   text-decoration: none !important;
 }
+
+.active {
```
SlyBouhafs, line 16: Some bootstrap components don't have the bsPrefix or the bsClass so i'm not sure if it's fixable.

components/ProfileDropdownMenu.tsx
anthonykhoa: If you look at the fields in your `customToggleProps`, you'll notice the fields are already attributes of a normal div component, except for the `children`. So there should already be some existing typescript definition out there you can use for this use case. For example an existing typescript definition you used is `React.Ref<HTMLDivElement>` to define the type for ref. There should similarly be one available for div element props

yarn.lock
```diff
@@ -2663,11 +2663,30 @@
     "@sentry/utils" "6.19.6"
     tslib "^1.9.3"
 
+<<<<<<< Updated upstream
 "@sentry/types@6.19.6", "@sentry/types@^6.11.0":
+=======
+<<<<<<< HEAD
+"@sentry/types@6.19.6":
```
anthonykhoa, line 2670: What package updates were needed in this file? Why were they needed?

scss/navLink.module.scss
anthonykhoa: What proposals did you try out to try to fix the problem of `using !important`? I'm wanting to know what you tried that failed, so that I know not to suggest the methods

components/AppNav.tsx
```diff
@@ -39,72 +32,52 @@ const navItems: NavItem[] = [
...
   return (
     <>
       {navItems.map(button => (
         <NavLink
           {...button}
-          className={`${styles['nav-item']} nav-link`}
+          className={`${styles['nav-item']} ${styles['nav-link']} px-3`}
           key={button.name}
-          activePath={location === button.path}
+          activePath={location.includes(button.path)}
```
anthonykhoa, line 46: What was wrong with the previous logic? `location === button.path`

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,133 @@
...
+  username,
+  isAdmin
+}) => {
+  const { data } = useUserInfoQuery({
+    variables: { username },
+    skip: !username
+  })
+  const router = useRouter()
+  const location = router.asPath
+  const PROFILE_PATH = '/profile/' + username
```
anthonykhoa, line 34: profile path should go in a constants file that should be accessible from any part of the codebase. I see a high potential for all `path` values to be re-used in the codebase. For example `/curriculum` appears a lot
https://github.com/garageScript/c0d3-app/search?q=curriculum

components/AppNav.tsx
```diff
@@ -124,15 +97,25 @@ const AppNav: React.FC<{}> = () => {
     // TODO: replace with typing
     const username = _.get(session, 'user.username', '')
 
-    return <LoggedInAuthNav username={username} />
+    return <LoggedInAuthNav username={username} session={session} />
```
anthonykhoa, line 100: This component grabs the isAdmin value that `AppNav` component already grabs, and there isn't any other logic other than returning the ProfileDropdownMenu component. What is the benefit of having this `LoggedInAuthNav` component?

components/AppNav.tsx
SlyBouhafs: I used `include` to highlight the navigation link when the user is under a subpage of that path, for now the only use case is mostly for the curriculum subpages, to indicate that they are still inside the curriculum path, i believe it's great for the UX. It will also be useful later if we add submenus for the the different new features.

anthonykhoa: Can you please list all UX changes in your PR description

yarn.lock
SlyBouhafs: Sorry, i didn't pay full attention to the yarn.lock, i didn't introduce any new deps, that's probably due to my branch not being up to date with master, i'll get it fixed.

components/ProfileDropdownMenu.tsx
SlyBouhafs: Thanks for explanation, i got it fixed.

components/AppNav.tsx
SlyBouhafs: You're right about that!

components/AppNav.tsx
SlyBouhafs: Description updated!

components/ProfileDropdownMenu.tsx
SlyBouhafs: Right, although the profile path is only used in the profile dropdown menu for now, I feel like this needs to be another different PR, the constants file doesn't exist yet.
I'll reference this issue here so that we can keep track of it: #1185

scss/navLink.module.scss
SlyBouhafs: Alright, i imported the AppNav styles and used them instead because they have higher specificity.

components/ProfileDropdownMenu.tsx
anthonykhoa: So far the only file that uses `/profile` is components/AppNav, but now we are introducing a new file components/ProfileDropdownMenu in this PR that uses it as well. We shouldn't be introducing more tech debt. `components/AppNav` is already an impacted file, so you're really only adding 1 more file to your PR. This current PR is kinda big already though, so I'd suggest to split this PR into smaller ones.

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,127 @@
...
+      href={path}
+      key={title}
+      bsPrefix={isActive(path)}
+    >
+      {title}
+    </Dropdown.Item>
+  ))
+
+  const buttonMenuToggle = React.forwardRef(
+    (props, ref: React.Ref<HTMLDivElement>) => (
```
anthonykhoa, line 65: Why does ref have a type definition, but props doesn't?

anthonykhoa: Actually we have no unique fields in the props object so this is ok

anthonykhoa: I think this might throw typescript error saying children doesn't exist in props

flacial: > by passing it generic types

What do you mean?

SlyBouhafs: i am not familiar yet with all the type stuff so I used `<HTMLDivElement>` as a type for forwardRef and tsc didn't scream.

components/ProfileDropdownMenu.tsx
SlyBouhafs: Created Constants for all the paths in AppNav and ProfileDropdownMenu.

components/ProfileDropdownMenu.tsx
anthonykhoa: This guy had code similar to yours and it ran into issue of children not appearing in props object
https://stackoverflow.com/questions/54654303/using-a-forwardref-component-with-children-in-typescript

In the documentation, it looks like we have to use `props.children` whenever the dropdown component we use contains children
https://react-bootstrap.github.io/components/dropdowns/

I tried it out on a typescript playground too and it gives me the error of children not existing in props
https://codesandbox.io/s/forwardref-typescript-c0b4r?file=/src/App.tsx
Try this code in the `App.tsx` file:
```
import * as React from "react";


const Search = React.forwardRef<HTMLDivElement>((props, ref) => {
  return <div ref={ref} {...props}>{props.children}</div>;
});
Search.displayName = "Search";

export default function App() {
  const input = React.useRef<HTMLInputElement>(null);
  React.useEffect(() => {
    if (input.current) {
      input.current.focus();
    }
  }, []);
  return <Search ref={input}>Banana</Search>;
}
```

I am not sure why the error does it appear on your end, but either way since we are using typescript, we should make sure to have a type definition for all parameters.

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,126 @@
...
+      key={title}
+      bsPrefix={isActive(path)}
+    >
+      {title}
+    </Dropdown.Item>
+  ))
+
+  const buttonMenuToggle = React.forwardRef<HTMLDivElement>((props, ref) => (
+    <div ref={ref} {...props}>
+      {props.children}
```
anthonykhoa, line 67: we actually dont need to pass props.children since we already do {...props} on line 66

yarn.lock
```diff
@@ -2814,8 +2814,7 @@
     remark-slug "^6.0.0"
     ts-dedent "^2.0.0"
     util-deprecate "^1.0.2"
-
-"@storybook/addon-essentials@^6.4.22":
+  "@storybook/addon-essentials@^6.4.22":
```
anthonykhoa, line 2817: Can we remove these changes? Looks like only formatting was changed.

SlyBouhafs: Fixed.

components/ProfileDropdownMenu.tsx
SlyBouhafs: After reading this: https://stackoverflow.com/a/66876771
It is suggested to either create a type with a children prop, use `React.PropsWithChildren` or use `React.ComponentPropsWithoutRef` if we want to forward all the props, i used the later so that it passes all the props.

constants/index.js
```diff
@@ -0,0 +1,10 @@
+// File for all project constant
```
flacial, line 1: I thought this file will be created in another PR. This PR creates the constants file AND update the AppNav now.

components/AppNav.tsx
```diff
@@ -1,110 +1,85 @@
...
-  { title: 'Lessons', path: '/admin/lessons' },
-  { title: 'Users', path: '/admin/users' },
-  { title: 'Alerts', path: '/admin/alerts' }
-]
 const navItems: NavItem[] = [
-  { path: '/curriculum', name: 'Curriculum' },
+  { path: CURRICULUM_PATH, name: 'Curriculum' },
   {
-    path: 'https://github.com/garageScript/c0d3-app',
+    path: REPO_PATH,
```
anthonykhoa, line 28: Looks like this file doesn't use `/profile` anymore, so all these changes to use constants in this file is just tech debt cleanup. Can we move this to another PR? What needs to be prevented is the creation of new tech debt, which the profiledropdown file will introduce without the `/profile` constant. I'd suggest you break this PR into smaller PRs, with one being for profiledropdown menu with constants file.

songz (approved):

components/AppNav.tsx
songz: > I'd suggest you break this PR into smaller PRs, with one being for profiledropdown menu with constants file.

I think @SlyBouhafs can keep this in mind for the future. Current refactoring has no issues I think its good to merge (PRs been open for awhile).

components/ProfileDropdownMenu.tsx
```diff
@@ -0,0 +1,125 @@
...
+              </Button>
+            </div>
+          </LogoutContainer>
+        </Dropdown.Item>
+      </div>
+    </Dropdown>
+  )
+}
+
+export default ProfileDropdownMenu
```
anthonykhoa, line 125: Lets make a story for this component. This will be useful in the future incase anyone needs to use this component again in the future.

Profile component stories are here
https://github.com/garageScript/c0d3-app/tree/master/stories/profile

anthonykhoa: This PR has been open for awhile so I don't wanna make this a blocker, but please do follow up with making a story for this in another PR since it is a part of engineering practices

anthonykhoa: This component isn't reusable, so story is not needed

SlyBouhafs: Marked as resolved because there is no need for a story, the component isn't reusable.

anthonykhoa (approved):

PR #1722, opened 2022-04-12
ref: Export repeated strings into an `enum`

### LINKED ISSUE #1652: Refactor formValidation file
https://github.com/garageScript/c0d3-app/pull/1651#discussion_r835832334

> I notice lots of repeated strings in this file. There should be more constants made to be used in this file, rather than repeating the same strings over and over again

[The file location](https://github.com/garageScript/c0d3-app/blob/master/helpers/formValidation.tsx)

flacial:
Closes #1652 

This PR moves all the repeated strings into 2 objects: `errorMessages, requirements`


helpers/formValidation.tsx
```diff
@@ -1,107 +1,121 @@
...
+  PasswordMin = 6,
+  LeadingTrailing = 'Leading and trailing value must be alphanumeric character',
+  NumbersOnly = 'Numbers only',
+  PositiveNumbers = 'Positive numbers only',
+  AlphaNumChars = 'Must be alphanumerics characters',
+  MustBeLowercase = 'Must be a lowercase string'
+}
+
 const REGEX_ALPHANUMERICS_AND_SPACE = /^[a-zA-Z0-9_\s]*$/
+const MustAtLeastTextMinChars = `Must be at least ${Constants.TextMin} characters`
```
flacial, line 16: This can't go into the enum because it doesn't allow computed values such as `${}`

SahilKalra98 (approved):

helpers/formValidation.tsx
```diff
@@ -1,107 +1,121 @@
 import * as Yup from 'yup'
 
-const TEXT_MIN = 2
-const TEXT_MAX = 64
-const PASSWORD_MIN = 6
+enum Constants {
```
anthonykhoa, line 3: Can we choose a name for this Enum that contains more meaning? Also, it seems we have a mix of 2 types of constants in this enum: error messages and requirements. IMO we can increase organization and readability of code by separating this into 2 enums

```
enum ErrorMessages {
  LeadingTrailing = 'Leading and trailing value must be alphanumeric character',
  NumbersOnly = 'Numbers only',
  PositiveNumbers = 'Positive numbers only',
  AlphaNumChars = 'Must be alphanumerics characters',
  MustBeLowercase = 'Must be a lowercase string',
  Required = 'Required'
}

enum Requirements {
  TextMin = 2,
  TextMax = 64,
  PasswordMin = 6
}
```

helpers/formValidation.tsx
anthonykhoa: The convention for constants is to user uppercase with words separated by underscore, please rename the variable names for these constants
https://google.github.io/styleguide/jsguide.html#naming-constant-names

anthonykhoa: Actually since we have these strings that have computed values, I suggest we have an `ERROR_MESSAGES` object instead of the `ErrorMessages` I suggested in my other comment. This way we can just add these constants to the ERROR_MESSAGES object, which will improve readability

flacial: I agree with you. I'll apply your comments soon.

helpers/formValidation.tsx
```diff
@@ -1,107 +1,128 @@
 import * as Yup from 'yup'
 
-const TEXT_MIN = 2
-const TEXT_MAX = 64
-const PASSWORD_MIN = 6
-const REGEX_ALPHANUMERICS_AND_SPACE = /^[a-zA-Z0-9_\s]*$/
+const Regex = {
```
flacial, line 3: It wasn't needed to make it an object but made for consistency.

anthonykhoa: Unless there is high potential for there to be more regexes used in this file in the future, then this is fine. But if this is the only regex we can foresee us using, then this constant should be declared directly as a regex rather than an object

helpers/formValidation.tsx
```diff
@@ -1,107 +1,128 @@
...
+  REGEX_ALPHANUMERICS_AND_SPACE: /^[a-zA-Z0-9_\s]*$/
+}
+
+const Requirements = {
+  TEXT_MIN: 2,
+  TEXT_MAX: 64,
+  PASSWORD_MIN: 6
+}
+
+const ErrorMessages = {
```
anthonykhoa, line 13: enums use pascalCase but javascript uses camelCase by convention. Please rename your variables appropiately

anthonykhoa (approved):

Issue #1566, opened 2022-03-10
DiffView toolbar goes over the submission filename


flacial:
![image](https://user-images.githubusercontent.com/35906419/157671721-bf265ac5-9739-41cb-801e-b9576e0d86bd.png)

Extreme proposal: Replace DiffViewer package
Soft proposal: Modify the DiffViewer source and move the toolbar into it

PR #1731, opened 2022-04-14
feat: Move the toolbar to ReactDiffViewer `toolbar` prop

flacial:
Related to #1566 
Depends on #1730 

`c0d3-diff` component now has a `toolbar` prop that renders a *toolbar*. This PR moves it to the prop so that the fix for #1566 can be created in the next PR (sub-pr).


SahilKalra98 (approved):

SlyBouhafs (approved):

components/DiffView.tsx
```diff
@@ -108,33 +106,35 @@ const DiffView: React.FC<{
...
-            >
-              Viewed
-            </label>
-          </div>
-          <CopyButton value={newValue.join('\n')} />
-        </div>
         <div className={scssStyles.diffView}>
           <ReactDiffViewer
+            toolbar={() => (
+              <>
```
flacial, line 112: This is a mess. It should be exported into a function.

```ts
const toolbar = () => { ... }

toolbar={toolbar}
```

PR #1775, opened 2022-04-24
Fix AppNav logout link

### LINKED ISSUE #1774: bug: Can't logout unless I click on the text "LOGOUT"
When I try to click on the logout button to log out, it doesn't work until I click on the text itself.

Green area: logout works
Any other area: logout doesn't

image

Component: https://github.com/garageScript/c0d3-app/blob/32da31c3a7b5a35284ad179ceed7024d4158b5b3/components/ProfileDropdownMenu.tsx#L98

SlyBouhafs:
This PR fixes #1774 where the logout link wouldn't work if the user clicks outside the logout text.


flacial (approved):

SahilKalra98 (approved):

anthonykhoa:
What can we do to prevent more of these occurrences in the future?

@SlyBouhafs Is it possible to add tests to test for clicking outside of login text? First we can catch this error in a unit test. Additionally, we don't have this setup yet, but what would be nice is a set of e2e integration tests that must be run before pushing code to production

PR #1783, opened 2022-04-26
fix: PT-MONO font doesn't load

### LINKED ISSUE #1781: bug: PT-MONO font doesn't load
A few instances have been reported of the font `PT-MONO` font not loading. This issue is caused by the value `format('tff')`

Not working:

https://github.com/garageScript/c0d3-app/blob/013d9d27d3eabbc8103b564ce546b0f8c6fcea15/scss/fonts.scss#L3

Working (remove `format('tff')` to fix it):

```css
 src: url('/fonts/PTMono-Regular.ttf'); 
```

flacial:
Closes #1781 

This PR will...

- Remove `format('tff')` because it's the cause of the issue
- Create a folder `/public/fonts` for the fonts


songz (approved):

SlyBouhafs (approved):

SlyBouhafs (commented):
I checked the preview on another device and the fix doesn't seem to be working.

SlyBouhafs (changes requested):
I don't think the font name matters at all, when i used google to load the font, the font worked straight away without having to change the name.
`@import url('https://fonts.googleapis.com/css2?family=PT+Mono&display=swap');`
you can try it here: http://164.92.183.234:3002/

Edit: After reading this: https://nextjs.org/docs/basic-features/static-file-serving
It seems that there is no need to add `/public` to the path when using Nextjs.

flacial:
>  It seems that there is no need to add /public to the path when using Nextjs.

I think this is exclusive for components and pages, but for SCSS files we need to explicitly add `/public`.

SlyBouhafs (approved):

songz (approved):

Issue #1769, opened 2022-04-23
design: Add the text styles to the variables


flacial:
Add the following [styles](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=727%3A4422) to the styles [variables](https://github.com/garageScript/c0d3-app/blob/master/scss/__variables.module.scss).

![image](https://user-images.githubusercontent.com/35906419/164944674-53fc0c49-6199-49b7-a0ce-f324c957bca4.png)

PR #1792, opened 2022-04-27
feat: Add Inter font

### LINKED ISSUE #1789: feat: Add Inter font
[Download](https://fonts.google.com/specimen/Inter) it, add it to the `/public/fonts` folder, and then create a `@fontface` in [fonts.scss](https://github.com/garageScript/c0d3-app/blob/master/scss/fonts.scss)

flacial:
Closes #1789 

This PR will add Inter font so the [font style](#1769) that uses it can be created


scss/fonts.scss
```diff
@@ -1,4 +1,11 @@
+// Used for Buttons text and h3 header
```
SlyBouhafs, line 1: This is also used in navigation links right?

SlyBouhafs (approved):

SahilKalra98 (approved):

scss/fonts.scss
flacial: Yeah. I think buttons cover both.

PR #1793, opened 2022-04-27
feat: Add Rubik font

### LINKED ISSUE #1790: feat: Add Rubik font
[Download](https://fonts.google.com/specimen/Rubik) it, add it to the `/public/fonts` folder, and then create a `@fontface` in [fonts.scss](https://github.com/garageScript/c0d3-app/blob/master/scss/fonts.scss)

flacial:
Closes #1790 

This PR adds the font needed to create the [text style](#1769) that uses it.


scss/fonts.scss
```diff
@@ -2,3 +2,9 @@
   font-family: 'PT MONO';
   src: url('/fonts/PTMono-Regular.ttf');
 }
+
+// Used for h1 header
+@font-face {
+  font-family: 'Rubik';
+  src: url('/fonts/Rubik-Light.woff2');
```
SlyBouhafs, line 9: Are we sure that adding format("woff2") breaks the font when we use woff2? It seems like it is used everywhere and people are not having issues with it.

flacial: `tff` didn't work because it's not a valid type according to MDN and it should be `format('truetype')`:

> The available types are: "woff", "woff2", "truetype", "opentype", "embedded-opentype", and "svg".

I'll update the fonts with `format('fontType')` because it also says:

> To provide the browser with a hint as to what format a font resource is — so it can select a suitable one

SlyBouhafs (approved):

SahilKalra98 (approved):

PR #1809, opened 2022-04-28
feat(DOJO): Create module inputs component

### LINKED ISSUE #1734: DOJO - Create module inputs component
[Component on Figma](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4568)

image

flacial:
Closes #1734 

**Description**

This component will be used to either update a module or add a new one. It decides to add a new module based on the existence of the prop `module`. It takes a function (`OnAddModule`) that will be called when a module is added/updated.

This PR creates the module inputs component that will...o

- Allow the admin to add modules
- Show either an error or success message

Image:

image

Issues:

- The story isn't working because of #1810 and will be fixed in a separate PR


components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+      }
+
+      setErrorMsg('')
+      await addModuleMutation()
+    } catch (err) {
+      setErrorMsg((err as GraphQLError).message)
+    }
+  }
+
+  const Message = () => {
```
flacial, line 60: The purpose of this component is to provide information on whether the module has been successfully added or not. And it's not related to the design.

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { AlertFillIcon, CheckCircleIcon } from '@primer/octicons-react'
+import styles from '../../../scss/adminLessonInputs.module.scss'
+import { Spinner } from 'react-bootstrap'
+import { GraphQLError } from 'graphql-request/dist/types'
+
+type Props = {
+  lesson: { title: string; id: number }
+}
+
+enum Error {
```
anthonykhoa, line 14: Is there potential for more error messages to be added in the future? If so, making an enum is fine. If not, then it would be more appropriate to declare this as a regular const variable instead of enum. If there no potential to add more items to the enum list, then declaring an enum(set of constants) for just one item(`INVALID_DATA`) is a readability issue since it will increase complexity without good reason

anthonykhoa: I think there is potential for more error messages to be added to the future for this file, but I just wanted to confirm this with you since you have more context than me

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+  INVALID_DATA = 'missing module name or description'
```
anthonykhoa, line 15: The convention for typescript enums is to use pascalCase
https://www.typescriptlang.org/docs/handbook/enums.html

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+const values: Option[] = [
+  {
+    title: 'Module Name'
+  },
+  {
+    title: 'Description',
+    type: 'MD_INPUT'
```
anthonykhoa, line 24: please use the constant instead
https://github.com/garageScript/c0d3-app/blob/master/components/FormCard.tsx#L13

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+  })
+
+  const [errorMsg, setErrorMsg] = useState(error?.message || '')
+
+  const handleChange = async (value: string, propertyIndex: number) => {
+    await formChange(value, propertyIndex, formOptions, setFormOptions)
+  }
+
+  const OnSubmit = async () => {
```
anthonykhoa, line 46: should be `onSubmit`

components/admin/lessons/AdminLessonInputs.tsx
anthonykhoa, line 60: Can we change this variable name to contain more meaning? For example `NotificationMessage`, or anything that conveys more of what this component is doing.

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+        </div>
+      )
+    }
+
+    if (data) {
+      return (
+        <div className={styles.success}>
+          <CheckCircleIcon />
+          <span>
+            Added the module <strong>{data.addModule.name}</strong>{' '}
```
anthonykhoa, line 84: Why is `{' '}` needed?

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,112 @@
...
+    title: 'Module Name'
+  },
+  {
+    title: 'Description',
+    type: 'MD_INPUT'
+  }
+]
+
+const AdminLessonInputs = ({ lesson }: Props) => {
+  const [formOptions, setFormOptions] = useState(values)
```
anthonykhoa, line 29: Can you define a type for `formOptions`? That should remove the need to do type casting on line 34 and 36. Type casting basically tells typescript `stop type checking and trust me, I know what I’m doing`, which weakens type safety. So optimally we should not use type casting unless it is the most appropriate choice

components/admin/lessons/AdminLessonInputs.tsx
flacial: There will be more errors from the GraphQL resolver after we update it to validate the arguments (it currently doesn't). For example, checking if the module name consists of at least alphanumeric characters or the description satisfies the max length.

components/admin/lessons/AdminLessonInputs.tsx
flacial: It's auto-added by the formatter, but I think it's safe to remove it.

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+type Props = {
+  lesson: { title: string; id: number }
+}
+
+enum Error {
+  InvalidData = 'missing module name or description'
+}
+
+type FormOptions = (Option & { value?: string })[]
```
anthonykhoa, line 18: Why is `& { value?: string }` needed? Seems there is already a value field in the Option type definition
https://github.com/garageScript/c0d3-app/blob/master/components/FormCard.tsx#L20

components/admin/lessons/AdminLessonInputs.tsx
anthonykhoa: Yeah lets remove it if its not needed

anthonykhoa: Also, is it guaranteed that `data.addModule` and `data.addModule.name` both exist? If not guaranteed, then this could lead to an issue due to property not existing

flacial: If there's `data`, there's `data.addModule`. But I'm not sure about `data.addModule.name`. If the property name changes, an issue will appear BUT, if that ever happens, TypeScript will scream with `name does not exist in { ... }`.

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,114 @@
...
+    if (loading) {
+      return (
+        <div className={styles.loading}>
+          <Spinner animation="grow" size="sm" />
+          <span>Adding the module...</span>
+        </div>
+      )
+    }
+
+    if (errorMsg || error) {
```
anthonykhoa, line 72: Is the `errorMsg ||` needed? Why are we doing `errorMsg || error`?

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,114 @@
...
+    try {
+      // Empty strings aren't handled in the resolver
+      if (!content.value || !name.value) {
+        return setErrorMsg(Error.InvalidData)
+      }
+
+      setErrorMsg('')
+      await addModuleMutation()
+    } catch (err) {
+      setErrorMsg((err as GraphQLError).message)
```
flacial, line 58: This should be `ApolloError` and not `GraphQLError`

components/admin/lessons/AdminLessonInputs.tsx
flacial: `errorMsg` is [set](https://github.com/garageScript/c0d3-app/pull/1809/files#diff-ce1a23c3d41862ff0962cd9bad62e18be06b4c3b5d8c03fb303e6e67fb24c99aR52) as a custom error message when the inputs are empty because the resolver doesn't validate the arguments.

It also [set](https://github.com/garageScript/c0d3-app/pull/1809/files#diff-ce1a23c3d41862ff0962cd9bad62e18be06b4c3b5d8c03fb303e6e67fb24c99aR58) the query error.

I think it could be `if (errorMsg)`.

components/admin/lessons/AdminLessonInputs.tsx
anthonykhoa: Is `name` something that is guaranteed to be a field of the `addModule` object? If the `name` isn't guaranteed to exist, then `addModule.name` will output `undefined`. I'm worried that will result in the word `undefined` being rendered here

flacial: Doesn't the [query](https://github.com/flacial/c0d3-app/blob/673ac51e43be35c7a3e125a26cfd0060a5104d83/graphql/queries/addModule.ts#L7) definition guarantee that the field will always exist?

components/admin/lessons/AdminLessonInputs.tsx
anthonykhoa: Ah, I see. Can you try passing a type variable to the useState call for formOptions? And see if that eliminates the need to do `& { value?: string }`

```
useState<Option[]>(
```

I'm thinking that in the future, there is potential for additional fields added to the form. And one of these additional fields might be a dropdown menu, so it might not be a good idea to restrict value to just `string`

flacial: `useState<Option[]>(values)` is similar to how it originally was _as_ TypeScript inferred the type as `Option[]`. Now we'll have to add type casting to prevent it from throwing `Type 'string | number | Item[]' is not assignable to type 'string'`.

components/admin/lessons/AdminLessonInputs.tsx
anthonykhoa: The query definition guarantee that the field will always exists, but it does not guarantee it will always be non-null, because fields can return `null`. 

I just looked up and confirmed that `name` will always return a non-null value though.
For the `addModule` resolver, the typedefs said it will return `Module`, and it will always be non-null
https://github.com/garageScript/c0d3-app/blob/master/graphql/typeDefs.ts#L59

And then in `Module` typedef, we have a `name` field that is also non-nullable and must be a string.
https://github.com/garageScript/c0d3-app/blob/master/graphql/typeDefs.ts#L241

If `name` was not non-nullable that might have been problematic edge case since the word `null` might be rendered when `name` is null.

components/admin/lessons/AdminLessonInputs.tsx
anthonykhoa: Hmm I see.. I'll try to look more into this today to see if I can find a solution

songz: any update on this?

songz (approved):

components/admin/lessons/AdminLessonInputs.tsx
flacial: > Why is {' '} needed?

@anthonykhoa When a line length is greater than the max length, it has to be wrapped to the next line. In JSX's case, it needs this `{'  '}` or there won't be a space between the text.

components/admin/lessons/AdminLessonInputs.tsx
anthonykhoa: A few options:

### Typescript 

Pros:
* With this method, we correctly define the type in the values array.
* Doing things this way will enable us to add a dropdown menu option in the future.

Cons:
* If this array grows large, it can be a readability concern to have to manually add the types for each element in the array. I am predicting that this array will not grow too large though, since this form is only for a specific use case. 
```
type textField = Omit<Option, 'value'> & {value: string}
const values: [textField, textField] = [
  {
    title: 'Module Name',
    value: '' // notice here that we define initial value as empty string for `value` field, instead of giving it a fallback value of empty string when passing it as a variable to the graphql mutation
  },
  {
    title: 'Description',
    type: 'MD_INPUT',
    value: ''
  }
]
```

### Using typeof operator

Pros:
* We don't have to make any new type definitions

Cons:
* We are doing the type checking ourselves, which seems like an anti-pattern because are doing the job that typescript should already do
* Every time a new field gets added to this form, we will have to do typechecking for that new field as well. If a dropdown menu gets added as a field, we would have to implement logic to check for type, and pull the right value correspondingly.

### Using String function

Pros:
* We don't have to make any new type definitions

Cons: 
* If the field is already a string, it seems redundant to have to stringify the string.
* If a dropdown menu gets added, a developer could fall into the problem of using String function on the dropdown menu value(which is an array). This can potentially cause a future issue of not properly grabbing value from the dropdown menu array.

### My opinion
I like the typescript solution the best. This solution allows us to still add a dropdown menu field to this form in the future, and we won't be skipping any type checking either, thus there will be no decrease in type safety. Since these form fields do not have dropdown menus yet, type casting field values to strings is actually not a huge safety concern. The potential for problems arises if we add a dropdown menu in the future. If we do type casting for those dropdown menus(they have an array of Item type ass their type), we can run into issues since type checking will not be strict. This is why we should avoid setting the precedent of using type casting in this file, and should only use it if it is the most appropriate choice to use

anthonykhoa:
Please include an image of the component in PR description as well

components/admin/lessons/AdminLessonInputs.tsx
flacial: Nice solution. It even solves another issue I had. I should learn more about TypeScript utilities.

components/admin/lessons/AdminLessonInputs.tsx
flacial: Is this resolved?

anthonykhoa: yeah just follow up with code changes as per our discussion about avoiding using type casting

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+]
+
+const AdminLessonInputs = ({ title, lessonId, onAddModule }: Props) => {
+  const [formOptions, setFormOptions] = useState<any>(values)
+  const [name, content] = formOptions
+
+  const [addModuleMutation, { data, error, loading }] = useAddModuleMutation({
+    variables: {
+      content: content.value || '',
```
anthonykhoa, line 41: Since we changed type to `TextField`, it should mean content.value guarantees to exist now as a string data type, so I think we can remove the `|| ''` since we don't have to check for undefined value anymore

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,130 @@
...
+      lessonId,
+      name: name.value || ''
+    }
+  })
+
+  const [errorMsg, setErrorMsg] = useState(error?.message || '')
```
anthonykhoa, line 47: This is not a PR blocker, but I noticed in @SlyBouhafs's PR he used lodash get instead of optional chaining. We should agree on which one to use for this codebase, or else inconsistencies will continue to appear

flacial: One issue I've with lodash `get` is that it hinders code completion. I think the rule to use them should be as if statement and ternary operator.

- Use loadash `get` when the property is deeply nested. We don't want to do `obj?.obj?.obj?.obj || ''`.
- Use optional chaning when the property isn't deeply nested.

anthonykhoa: What do you mean by `it hinders code completion`?

Both optional chaining and lodash get do the same thing though. We're using 2 different tools that do exactly the same thing, its like using 2 different API's that do exactly the same as each other.

anthonykhoa (approved):

components/admin/lessons/AdminLessonInputs.tsx
flacial: > What do you mean by it hinders code completion?

With lodash `get` we use a string `_.get('obj.obj')`. By using a string, the IDE/Extension/Plugin won't display auto-suggestions of the properties that this object has.

With string:
image

Without:
image

components/admin/lessons/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,130 @@
...
+  },
+  {
+    title: 'Description',
+    type: MD_INPUT,
+    value: ''
+  }
+]
+
+const AdminLessonInputs = ({ title, lessonId, onAddModule }: Props) => {
+  const [formOptions, setFormOptions] = useState<any>(values)
```
anthonykhoa, line 36: Wait, why are we passing `any` as a type variable?

anthonykhoa (changes requested):
Why are we now passing `any` as a type variable to useState? I think typescript should be able to infer the type already
```
useState<any>(values)
```

flacial:
> Why are we now passing `any` as a type variable to useState? I think typescript should be able to infer the type already
> ```
> useState<any>(values)
> ```

[`formChange`](https://github.com/garageScript/c0d3-app/blob/master/helpers/formChange.ts#L7) has the type of `state` and `setState` as `any`. Typescript will infer the last type we added and it causes an issue if it's anything other than `any`.

anthonykhoa (approved):

anthonykhoa:
> > Why are we now passing `any` as a type variable to useState? I think typescript should be able to infer the type already
> > ```
> > useState<any>(values)
> > ```
> 
> [`formChange`](https://github.com/garageScript/c0d3-app/blob/master/helpers/formChange.ts#L7) has the type of `state` and `setState` as `any`. Typescript will infer the last type we added and it causes an issue if it's anything other than `any`.
> 
> `Argument of type 'Dispatch<SetStateAction<[TextField, TextField]>>' is not assignable to parameter of type 'Dispatch<SetStateAction<any[]>>'`

Thanks for the explanation @flacial. Seems that formChange function did not use typescript properly, leading to this issue. Not a PR blocker though since this is a separate issue.

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+type Module = { id: number; name: string; content: string }
+
+type Props = {
+  lessonId: number
+  refetch: RefetchQueriesFunction
+  title?: string
+  onAddModule?: (
+    m:
+      | (AddModuleMutation['addModule'] & { lesson: { id: number } })
```
flacial, line 24: I had to define the lesson ID field because `UpdateModuleMutation` and [`AddModuleMutation`](https://github.com/flacial/c0d3-app/blob/e0515ccd9df5f9ece6b52eed8e19b55f691a09ac/graphql/index.tsx#L458) doesn't have it.

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+    e: { name: string; content: string } | null
+  ) => void
+  module?: Module | undefined
+}
+
+enum Error {
+  InvalidData = 'missing module name or description'
+}
+
+const initValues = (
```
flacial, line 36: This has been turned into a function to auto-fill the forms based on the `module` prop data.

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    return <></>
+  }
+
+  return (
+    <div className={styles.container}>
+      <QueryStateMessage />
+      <FormCard
+        title={name.value || title || 'Untitled'}
```
flacial, line 171: `name.value` is used to dynamically fill the title as the admin is typing it in the input.

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+      if (!content.value || !name.value) {
+        return setErrorMsg(Error.InvalidData)
+      }
+
+      setErrorMsg('')
+      const newModule = await moduleMutation()
+      refetch()
+
+      if (onAddModule) {
+        const updateModuleData = get(newModule, 'data.updateModule')
```
flacial, line 101: I would use `?.` but when I do I get `Property 'updateModule' does not exist on type 'AddModuleMutation | UpdateModuleMutation'` This is why I didn't use it.

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+import styles from '../../../scss/adminModuleInputs.module.scss'
+import { Spinner } from 'react-bootstrap'
+import { get } from 'lodash'
+import { RefetchQueriesFunction, ApolloError } from '@apollo/client'
+
+type Module = { id: number; name: string; content: string }
+
+type Props = {
+  lessonId: number
+  refetch: RefetchQueriesFunction
```
flacial, line 20: This is a `useQuery` refetch function. It's used to refetch the modules and update the modules list with the newly added modules.

components/admin/lessons/AdminModuleInputs.tsx
anthonykhoa: This is an inconsistency. On line 83 I see we use optional chaining instead of lodash get. If you want to use optional chaining for this file, please fix the issues that arise when you use optional chaining so that you can use it.

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+  title,
+  lessonId,
+  onAddModule,
+  module,
+  refetch
+}: Props) => {
+  const [formOptions, setFormOptions] = useState<any>(initValues())
+  const [name, content] = formOptions
+
+  useEffect(
```
anthonykhoa, line 61: What is this useEffect used for?

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+  refetch: RefetchQueriesFunction
+  title?: string
+  onAddModule?: (
+    m:
+      | (AddModuleMutation['addModule'] & { lesson: { id: number } })
+      | (UpdateModuleMutation['updateModule'] & { lesson: { id: number } })
+      | null,
+    e: { name: string; content: string } | null
+  ) => void
+  module?: Module | undefined
```
anthonykhoa, line 29: do we need `| undefined`? Since module is already optional

components/admin/lessons/AdminModuleInputs.tsx
flacial: It's used to update the `formOptions` when the currently active module is changed.

components/admin/lessons/AdminModuleInputs.tsx
flacial: No, we don't.

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,185 @@
...
+        variables: {
+          id: module.id,
+          content: content.value,
+          lessonId,
+          name: name.value
+        }
+      })
+    : useAddModuleMutation({
+        variables: {
+          content: content.value,
```
anthonykhoa, line 77: lines 76 - 80 are repeated on lines 68 - 73. Can you reduce the code duplication?

components/admin/lessons/AdminModuleInputs.tsx
```diff
@@ -0,0 +1,188 @@
...
+  ApolloError,
+  OperationVariables,
+  ApolloQueryResult
+} from '@apollo/client'
+
+type Module = { id: number; name: string; content: string }
+
+type Props = {
+  lessonId: number
+  refetch: (variables?: Partial<OperationVariables> | undefined) => Promise<
```
anthonykhoa, line 24: Is `| undefined` needed here? Since variables is optional

stories/components/AdminModuleInputs.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+          }
+        }
+      ]}
+    >
+      <AdminModuleInputs
+        lessonId={lesson.id}
+        title={lesson.title}
+        refetch={
+          (() => {}) as (
+            variables?: Partial<OperationVariables> | undefined
```
anthonykhoa, line 61: Do we need `| undefined` here since variables is optional?

components/admin/lessons/AdminModuleInputs.tsx
flacial: No, it isn't.

anthonykhoa (approved):

Issue #1805, opened 2022-04-28
fix: Breadcrumbs has extra padding-left


flacial:
Add the property `padding: 0` to [`.breadcrumbs`](https://github.com/flacial/c0d3-app/blob/ba78dd725d754e1442bee06521d0fb3169d09c9f/scss/breadcrumbs.module.scss#L8)

![image](https://user-images.githubusercontent.com/35906419/165819248-8d182f72-16ce-4f9b-b56b-5b0f06c98f4d.png)

shockz09:
fixed

Issue #1758, opened 2022-04-20
bug: discord component story isn't working


flacial:
![image](https://user-images.githubusercontent.com/35906419/164321570-9f5d9a5f-d801-4548-ae1e-a847d7c2a452.png)

The error originated from attempt to access `jest.fn()` to mock `global.fetch` or `fetch` function. 

- `jest-mock` package isn't installed, and even if it got installed webpack will fail to compile it.

### Solution

- Install [fetch-mock](https://www.npmjs.com/package/fetch-mock)
- Follow its docs to mock each story fetch request.

Example stories fetch mock:

```tsx
export const ErrorFetching: React.FC = () => {
  fetchMock.restore().mock('*', 404)
  return (
    <Wrapper>
      <DiscordBar />
    </Wrapper>
  )
}
```

```tsx
export const EmptyMembers: React.FC = () => {
  fetchMock.restore().mock('*', { presence_count: 0, status: 200, members: [] })
  return (
    <Wrapper>
      <DiscordBar />
    </Wrapper>
  )
}
```

shockz09:
fixed in https://github.com/garageScript/c0d3-app/pull/1817

flacial:
@shockz09 I replaced the solution as the previous one wasn't working. I also added more details and some examples.

shockz09:
i see

PR #1817, opened 2022-04-30
fix: remove Breadcrumbs padding

shockz09:
solves https://github.com/garageScript/c0d3-app/issues/1805 issue


shockz09:
now also solves https://github.com/garageScript/c0d3-app/issues/1758

shockz09:
@flacial

.storybook/main.js
```diff
@@ -5,6 +5,6 @@ module.exports = {
     '@storybook/addon-actions',
     '@storybook/addon-viewport',
     '@storybook/addon-storysource',
-    "storybook-addon-next-router",
```
flacial, line 8: Rather than modifying an existing addon. It's better to add it to the next line:

```
...
"storybook-addon-next-router",
'@storybook/addon-mock'
```

flacial: This PR should be moved to another PR, so this PR's goal is to only fix the padding.

shockz09: sure

shockz09: > Rather than modifying an existing addon. It's better to add it to the next line:
> 
> ```
> ...
> "storybook-addon-next-router",
> '@storybook/addon-mock'
> ```

it would be cuz of the git pull i think, I will fix this

flacial: You can remove the discord bug solution entirely because it won't work, haha. I updated the issue with a working solution: https://github.com/garageScript/c0d3-app/issues/1758

.storybook/main.js
```diff
@@ -5,6 +5,6 @@ module.exports = {
...
+    '@storybook/addon-mock'
```
anthonykhoa, line 8: How does this solve the issue? Please add screenshots of the fix as well.

scss/breadcrumbs.module.scss
```diff
@@ -10,7 +10,7 @@
   column-gap: 24px;
   list-style: none;
   align-items: center;
-
+  padding:0;
```
anthonykhoa, line 13: please use formatting that the rest of the css file uses.

.storybook/main.js
shockz09: It was a mistake actually, because of some git issues, i will fix it asap

flacial:
@shockz09 Any updates?

shockz09:
@flacial i will do it today for sure

flacial:
Closed due to idleness.

PR #1826, opened 2022-05-03
Improved Discord notifications

JasirZaeem:
Followup to #949 
## Use discord ids of connected users to add and improve notifications
- ### On challenge submission

image

- ### On lesson completion

image

image

- ### On receiving a star

image

- ### On a comment on user's submission
Sent dm
image

- ### On  acceptance/rejection of user's submission
Sent dm
image
image

The title of the direct message is the reviewer's or commenter's name on c0d3.
The message either mentions the c0d3 user or discord username/nickname if their discord id is available, this let's other's easily know them on discord. The mention can also be clicked on to dm users.
image


graphql/resolvers/submissionController.test.js
```diff
@@ -19,20 +22,28 @@ import {
 } from './submissionController'
 import * as Sentry from '@sentry/node'
 
+// jest.mock('../getDiscordMessageUserIdString.ts', () => ({
```
flacial, line 25: I think you forgot to remove this.

constants/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export const ADMIN_PATH = '/admin'
+export const DOCS_PATH = '/docs'
+export const DISCORD_PATH = 'https://discord.gg/c0d3'
+export const LOGIN_PATH = '/login'
+export const SIGNUP_PATH = '/signup'
+
+// URL Constants
+export const DEPLOYMENT_URL = process.env.VERCEL_URL
+  ? `https://${process.env.VERCEL_URL}`
+  : `http://localhost:${process.env.PORT ?? 3000}`
```
anthonykhoa, line 16: Is there a need to explicitly check for null or undefined using `??` operator? Seems like it may be unneeded additional complexity. If the OR operator `||` can be used here instead it would make things simpler

JasirZaeem: Yep, you are right `||` would be better in this case, I have updated it.

graphql/resolvers/addComment.test.js
```diff
@@ -0,0 +1,178 @@
+/**
```
flacial, line 1: NIT: space between the test cases and variables would make the code a little bit easier to go through.

graphql/resolvers/addComment.ts
```diff
@@ -10,13 +19,70 @@ export const addComment = async (
...
+          challenge: true
+        }
+      }
     }
   })
+
+  const { author, submission } = comment
+
+  if (author.id !== submission.user.id && submission.user.discordId) {
+    const notificationEmbed: MessageEmbedOptions = {
```
flacial, line 59: Is it necessary to explicitly give it a type? TypeScript is capable of inferring its type and comparing the object's properties with `sendDirectMessage` first parameter type.

flacial (approved):

helpers/updateSubmission.ts
```diff
@@ -60,56 +70,90 @@ export const updateSubmission = async (
...
-    },
-    data: {
-      passedAt: new Date()
+    })
+  }
+
+  if (user.discordId && reviewer) {
+    const reviewerString = getDiscordMessageUserIdString(reviewer)
+    const reviewNotificationEmbed: MessageEmbedOptions = {
+      color: '#5440d8',
```
songz, line 123: should this also be a constant?

songz (approved):

graphql/resolvers/addComment.ts
JasirZaeem: Yeah ts will figure out if the type is sufficient when this object is passed to the function, but I have explicitly typed the object literal as `MessageEmbedOptions` to help with autocomplete and ides when reading/modifying the variable where it is declared.

SahilKalra98 (approved):

helpers/updateSubmission.ts
JasirZaeem: Yep, Done.

PR #1828, opened 2022-05-03
DOJO: Create Lessons side-nav component

### LINKED ISSUE #1733: DOJO - Create modules list component
[Component on Figma](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4548)

image

flacial:
Closes #1733 

This PR adds a component that takes a list of items (modules or challenges) and displays them. It allows the selection and adding of items. It'll be used to manage the modules and challenges on the admin lesson page.

<details><summary>Image</summary>

image

</details>

[Storybook](https://deploy-preview-1828--youthful-thompson-448332.netlify.app/?path=/story/components-adminlessonsidenav--basic)


scss/adminLessonSideNav.module.scss
```diff
@@ -0,0 +1,104 @@
...
+  max-width: 369.01px;
+  row-gap: 23px;
+  justify-content: center;
+}
+
+.container__heading {
+  font-weight: bold;
+  font-size: 18px;
+  font-family: Inter;
+  font-weight: 700;
```
flacial, line 30: Inter 700 font file doesn't exist. So the font won't display bold.

SlyBouhafs: Inter bold font has been added in #1832

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,137 @@
...
+            <span>{item.name}</span>
+            {withChevron}
+          </li>
+        )
+      }
+
+      return acc
+    }, [])
+
+  useEffect(() => setActive && setActive(-1), [lessonId])
```
songz, line 69: why do we need `useEffect` here?

flacial: When the current lesson changes, I think it's better not to set any active item by setting its value as `-1`. By setting the value to `-1`, it'll be in add-new-module mode.

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,137 @@
...
+  setActive,
+  active,
+  title,
+  lessonId
+}: Props) => {
+  const setAsActive = (id: number) => setActive && setActive(id)
+
+  const itemsList =
+    items &&
+    items.reduce((acc: JSX.Element[], item) => {
```
songz, line 43: I think `map` is more appropriate. if there are 5 items, there would be 5 elements

```js
const itemsList = (items || []).map(...)
```

components/admin/lessons/AdminLessonSideNav.tsx
songz: I think its better for the parent to be managing the state, the parent should be passing in `-1`. Then we don't have to do this right? I'm imaginging here's how the parent would use this:

```js

const AdminModulePage = () => {
  const [selected, setSelected] => useState(-1)
  
  const items = [{
    title: 'one',
  }, {title: 'two'}]
  
  return (
    <AdminLessonSideNav
      onClick = {(index) => setSelected(index)}
      selectedIndex= {selected}
    />
  )
 ```

File: components/admin/lessons/AdminLessonSideNav.tsx (hunk 3 above)
Comment (flacial) [reply components/admin/lessons/AdminLessonSideNav.tsx:43]: My initial approach was with `map`, but its array will end up having some `undefined` elements because it's filtering them. Using `filter` with `map` is O(a + b) or 2 loops with each one being O(n). The best solution I found is `reduce`.

File: components/admin/lessons/AdminLessonSideNav.tsx [hunk 4]
```diff
@@ -0,0 +1,129 @@
...
+  error?: ApolloError
+  active?: number
+  onClick: (item: Item) => void
+}
+
+const AdminLessonSideNav = ({
+  items,
+  onAddItem,
+  loading,
+  error,
```
songz, line 33: When do you see an error being passed into `AdminLessonSideNav`? I would imagine the parent would not render this on error or pass in empty lessons?

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,129 @@
...
+  loading?: boolean
+  error?: ApolloError
+  active?: number
+  onClick: (item: Item) => void
+}
+
+const AdminLessonSideNav = ({
+  items,
+  onAddItem,
+  loading,
```
songz, line 32: `AdminLessonSideNav` does not send any external requests, so I would not give it a loading prop.

components/admin/lessons/AdminLessonSideNav.tsx
songz: the filtering should be done in the parent and only pass in the relevant lessons into this component.

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,129 @@
...
+  error,
+  active,
+  title,
+  lessonId,
```
songz, line 36: the filtering should be done in the parent and only pass in the relevant lessons into this component, so `lessonId` is not necessary to pass in as a prop.

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,129 @@
...
+  active?: number
+  onClick: (item: Item) => void
+}
+
+const AdminLessonSideNav = ({
+  items,
+  onAddItem,
+  loading,
+  error,
+  active,
```
songz, line 34: `selectedIndex` is better, indicating the index of the array element that is selected.

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,129 @@
...
+  title,
+  lessonId,
+  onClick
```
songz, line 37: more of a nitpick here,`onSelect` is a better name here, because on mobile devices there's no click. And also for accessibility there's no click either most ppl use keyboard tab select.

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,129 @@
...
+      return acc
+    }, [])
+
+  const renderItems = () => {
+    if (error) {
+      return (
+        <div className={styles.container__error}>
+          <AlertFillIcon />
+          <span className={styles.container__heading}>
+            Failed to get the modules
```
songz, line 69: this component is not sending any requests, so it is confusing for it to have a state that says `Failed to get modules` because its not even getting the modules.

components/admin/lessons/AdminLessonSideNav.tsx
```diff
@@ -0,0 +1,129 @@
...
+        </ol>
+
+        <button onClick={onClick} className={styles['container--button']}>
+          ADD NEW MODULE
+        </button>
+      </>
+    )
+  }
+
+  return <div className={styles.container}>{renderItems()}</div>
```
songz, line 126: a few things going on here that feels wrong.

1. I would avoid running a function that returns a component inside a component, it is a bit of an antipattern because a component is a function that returns a component, you could have broken this into a separate component like `<Content ... />` or something. But I wouldn't recommend it for this case because of the next point
2. You created a `renderItems` function to save repeating 1 line of code `<div className={styles.container}>`, which does not help much. It is cleaner to just do:

```
if (condition1) {
  return (
    <div className={styles.container}>
       ...
    </div>
  )
}
if (condition2) {
  return (
    ...
  )
}
...
```

songz (changes requested):

songz (approved):

PR #1850, opened 2022-05-12
fix: Connect to discord modal appears for connected users

### LINKED ISSUE #1527: I keep on seeing the Connect to Discord popup
When i go to c0d3.com I sometimes see that Connect to Discord popup that has options to either `No thanks, I'll study on my own.` or `Connect now`.. even though I clicked one of these options awhile back already.
I am able to reproduce this in production by clearing localstorage `localStorage.clear()`
In our curriculum we even have a section on localStorage. This means that there is a high potential for students to come across this issue because they might try out localstorage functions in t heir browsers.

Additionally, i expect(`I havent tried yet`) that if a person tries to user another browser, they'll come across the issue as well. 

I don't know exact specifics of implementation, but if we are relying on localStorage in order to store persisting data, then that is a bad idea. Its bad because localStorage is dependent on the user. For saving persisting data we should utilize the backend.

flacial:
Closes #1527 

The issue cause is that after the user connects to discord, the query gets the user's data from the cache instead of the server where it's the most updated. 

This PR sets the query `fetchPolicy` to `network-only` so it'll always get the most updated data from the server instead of the cache. This will prevent any outdated data.

[Article that helped](https://medium.com/@galen.corey/understanding-apollo-fetch-policies-705b5ad71980)


anthonykhoa:
Lets say I have already connected to discord. I then log out of c0d3.com, and then log back in. Will the changes in this PR prevent the `connect to discord` modal from popping back up?

flacial:
> Lets say I have already connected to discord. I then log out of c0d3.com, and then log back in. Will the changes in this PR prevent the `connect to discord` modal from popping back up?

Yes, the modal won't appear. I tested it.

> Also, how is c0d3 currently identifying whether the user has connected to discord before or not?

It identifies it based on whether the `discordRefreshToken` exists or not. [Code](https://github.com/flacial/c0d3-app/blob/dfba87fd65ea5fa72378e83925dfe5337053e1ef/graphql/resolvers/session.ts#L16)

SlyBouhafs (approved):

songz (approved):

PR #1873, opened 2022-05-19
Next-Auth: Add Credentials provider

### LINKED ISSUE #1870: Next-Auth: Add Credentials Provider
Add [Credentials provider](https://next-auth.js.org/providers/credentials) to start login/signup flows.

flacial:
Closes #1870 

This PR adds the Credentials provider, so it can be possible to call `signIn('credentials')` and start login/signup flow.

After merging this PR, the [jwt](https://next-auth.js.org/configuration/callbacks#jwt-callback) and [session](https://next-auth.js.org/configuration/callbacks#session-callback) callbacks will be added to make setting the user object in the session as the user from db possible.


SlyBouhafs (approved):

helpers/nextAuth.ts
```diff
@@ -3,14 +3,74 @@ import { Account, User } from 'next-auth'
...
+            username,
+            email,
+            // Will be uncommented when signup takes password as argument
+            // password,
+            firstName,
+            lastName
+          },
+          context
+        )
+      : isLoginFlow && (await login(undefined, { username, password }, context))
```
songz, line 44: nit: for complex logic like this one, it is better for readability if you explicitly used `if/else` instead of ternary operator. Keep Ternary operator to simple and obvious if/else use cases.

helpers/nextAuth.ts
```diff
@@ -3,14 +3,74 @@ import { Account, User } from 'next-auth'
...
+
+    if (!user) return null
+
+    const dbUser = await prisma.user.findFirst({
+      where: { id: user.id }
+    })
+
+    // What we return here is passed to jwt and session callbacks
+    return dbUser
```
songz, line 53: you can just `return prisma.user.findFirst(....)` right? No need to create a `dbUser` variable its not used for anything.

songz (approved):

helpers/nextAuth.ts
flacial: I chose immutability over readability. We could argue that even if it is mutable, it's not mutated in any other place than the `if/else`.

Issue #1848, opened 2022-05-12
DOJO: Create the modules page


flacial:
The components needed by the page were added. It's time to connect them together and [create it](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4496).

PR #1886, opened 2022-05-23
DOJO: Create modules page

flacial:
Closes https://github.com/garageScript/c0d3-app/issues/1848

This PR creates the [modules page](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4532) and creates the Lessons' page folder. By merging this PR, it'll be possible to create and update the modules.

- You might notice the design is different from the one in Figma. This will be fixed in a different PR.
- This page is called `modules`, but it's the new lessons page. It'll be renamed and refactored once more panels are added. It only contains the modules' panel for now.

To test it, access https://c0d3-app-git-fork-flacial-1848-dojo-create-the-cd28db-c0d3-prod.vercel.app/admin/lessons/1/modules


pages/admin/lessons/dojo/modules.tsx
```diff
@@ -0,0 +1,126 @@
+import { gql, useQuery } from '@apollo/client'
```
songz, line 1: path should be: `/pages/admin/lessons/:lessonId/modules`

songz: We don't need a `dojo` in the url path

pages/admin/lessons/dojo/modules.tsx
```diff
@@ -0,0 +1,126 @@
...
+    </div>
+  )
+}
+
+const Lessons = ({ data }: GetAppProps) => {
+  const { lessons } = data
+  const { data: modulesData, refetch } = useQuery<{ modules: Modules }>(MODULES)
+
+  const [lesson, setLesson] = useState({ title: '', id: -1 })
+  const [modules, setModules] = useState(sortBy(modulesData?.modules, 'name'))
```
songz, line 77: ooo we need a `order` for sorting these modules. Could we create a PR to update Module model to include `order` (use lesson&challenge model) as reference.

pages/admin/lessons/dojo/modules.tsx
```diff
@@ -0,0 +1,126 @@
...
+const ModulesTab = ({
+  filteredModules,
+  lessonId,
+  refetch
+}: ModulesTabProps) => {
+  const [selectedIndex, setSelectedIndex] = useState(-1)
+  const onAddItem = () => setSelectedIndex(-1)
+  const onSelect = (item: Module) => setSelectedIndex(item.id)
+
+  useEffect(() => setSelectedIndex(filteredModules[0]?.id), [lessonId])
```
anthonykhoa, line 50: For the sake of consistency, can we stick to using either `get` or `optional chaining`?
Also, can you try setting a default value as empty object for filteredModules in this function's arguments? That should avoid the issue of having to use optional chaining or lodash `get` here. I'm not sure if it can be done without typescript issues tho

pages/admin/lessons/dojo/modules.tsx
```diff
@@ -0,0 +1,126 @@
...
+          module={filteredModules.find(module => module.id === selectedIndex)}
+        />
+      </div>
+    </div>
+  )
+}
+
+const Lessons = ({ data }: GetAppProps) => {
+  const { lessons } = data
+  const { data: modulesData, refetch } = useQuery<{ modules: Modules }>(MODULES)
```
anthonykhoa, line 74: Is it possible to set a default value for `modulesData` so that we don't have to use optional chaining or lodash `get` ?

anthonykhoa: Same with the lessons variable, can you look into trying to set a default value for that as well

anthonykhoa: nevermind about lessons variable, it looks like we only extract a property from it one time in this component

anthonykhoa: Hmm actually i'm not sure if its appropriate to set a default value for `modulesData` either since we only grab `modules.name` from it. I'll resolve this thread and leave it up for you to decide

pages/admin/lessons/dojo/modules.tsx
flacial: Removing the rest of breadcrumb items will be done in a different PR

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const [modules, setModules] = useState(
+    sortBy(get(modulesData, 'modules'), 'order')
+  )
+
+  const filteredModules = useMemo(
+    () => modules.filter(module => module.lesson.id === lesson.id),
+    [lesson.id, modules]
+  )
+
+  useEffect(() => {
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:97]: this `useEffect` to setModules by sorted order is uncesseray, because it is initialized to sorted order in line 89.

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+    setModules(sortBy(get(modulesData, 'modules'), 'order'))
+  }, [get(modulesData, 'modules')])
+
+  useEffect(() => {
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:101]: why does this need to be inside `useEffect`? If lesson data is passed in as prop and lessonId is retrieved in the url, you can just get current lesson right away

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+}
+
+const Lessons = ({ data }: GetAppProps) => {
+  const router = useRouter()
+  const { pageName, lessonId } = router.query
+
+  const { lessons } = data
+  const { data: modulesData, refetch } = useQuery<{ modules: Modules }>(MODULES)
+
+  const [lesson, setLesson] = useState({
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:84]: you don't need this as a state, because lessonId is coming from router

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonId]/[pageName]/index.tsx:97]: I think line 89 is dead code because, at the moment, the module query didn't resolve yet. So, it could be turned into:

`useState(undefined)`

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonId]/[pageName]/index.tsx:101]: I agree. `lessons` should be out of the `useEffect` dependencies, but the `lessonId` has to stay because in the first couple of renders, `lessonId` is undefined. It seems like `useRouter` doesn't immediately extract the data from the URL.

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+    [lesson.id, modules]
+  )
+
+  useEffect(() => {
+    setModules(sortBy(get(modulesData, 'modules'), 'order'))
+  }, [get(modulesData, 'modules')])
+
+  useEffect(() => {
+    if (lessons) {
+      const lessonFromParam = lessons.find(e => e.id === toNumber(lessonId))
```
Comment (flacial) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:103]: Should be `e.order === toNumber(lessonId)`

kyoumei (approved):

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
Comment (songz) [reply pages/admin/lessons/[lessonId]/[pageName]/index.tsx:103]: id is correct, we are getting the selected lesson (by id). I think order is used to sort the lessons in order, not really used for id comparison

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonId]/[pageName]/index.tsx:101]: I removed `lessons` but it seems it doesn't resolve immediately. It has to exist

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonId]/[pageName]/index.tsx:84]: Yes, but it's needed by the `Breadcrumbs` component

Comment (songz) [reply pages/admin/lessons/[lessonId]/[pageName]/index.tsx:84]: I meant you could do this:

```js
const lesson = lessons.find(e => e.id === toNumber(lessonId)) || {title: '', id: -1}
```

You don't need a `lesson` state.

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const { pageName, lessonId } = router.query
+
+  const { lessons } = data
+  const { data: modulesData, refetch } = useQuery<{ modules: Modules }>(MODULES)
+
+  const [lesson, setLesson] = useState({
+    title: '',
+    id: -1
+  })
+  const [modules, setModules] = useState(
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:88]: I don't think you need a state for `modules` you could just do:

```js
const modules = sortBy(get(modulesData, 'modules'), 'order')
```

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </header>
+        <section>
+          <AdminLessonNav
+            tabs={[
+              {
+                tabName: 'modules',
+                urlPageName: 'modules'
+              }
+            ]}
+            render={navItem => {
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:149]: This is starting to look like a component, so let's try to pass in a component here instead. Generally we always try to render components instead of running functions.

```js

const SideNavComponent = ({navItem}) => {
  const isSelected = navItem.urlPageName === pageName
  const className =
    navStyles[
      isSelected
        ? 'lessons_tabsNav__nav__item'
        : 'lessons_tabsNav__nav__item--inactive'
    ]
  
  return (
    <Link href={`${MAIN_PATH}/${lesson.id}/${navItem.urlPageName}`}>
      <a className={`${className} nav-pills`}>
        {toUpper(navItem.tabName)}
      </a>
    </Link>
  )
  }
  
return (
      <AdminLayout data={data}>
      <main className={styles.container}>
        <header>...</header>
        <section>
          <AdminLessonNav
            tabs={[
              {
                tabName: 'modules',
                urlPageName: 'modules'
              }
            ]} 
            Component={SideNavComponent}
            />
            ...
)
  )
  ```
  
  And then to render this component, make the change in this file: https://github.com/garageScript/c0d3-app/blob/master/components/admin/lessons/AdminLessonNav.tsx#L15
  
```js
  const AdminLessonNav: React.FC<Props> = ({ tabs, Component }: Props) => {
  return (
    <div>
      <div className={styles.lessons_tabsNav}>
        <div className={styles.lessons__tabsNav__nav}>
          {tabs.map((tab, i) => (
            <Component key={i} navItem={tab}></Component>
          ))}
        </div>
      </div>
    </div>
  )
}
```

songz (changes requested):

flacial:
bug: Lesson ID is wrong when admin switch lessons. We should use the slug property instead of ID.

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+  return (
+    <AdminLayout data={data}>
+      <main className={styles.container}>
+        <header>
+          <h1>Lesson Info</h1>
+          <Breadcrumbs
+            lesson={lesson}
+            setLesson={lesson => {
+              router.push(
+                `${MAIN_PATH}/${extractLessonOrder(lesson.slug)}/${pageName}`
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:149]: if we are now using lesson.slug, should the file name be `/pages/admin/lessons/[lessonSlug]/[page...`

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+  content: string
+  lesson: { id?: number }
+  order: number
+}
+type Modules = Module[]
+
+type ReqParam = string | string[] | undefined
+
+type ContentProps = {
+  pageName: ReqParam
```
Comment (anthonykhoa) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:45]: Looks like the only place `ReqParam` used is here. Can we make pageName an optional field? That way you dont have to explicitly define ReqParam as having possibility of undefined yourself

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const router = useRouter()
+  const { pageName, lessonId } = router.query
+
+  const { lessons } = data
+  const { data: modulesData, refetch } = useQuery<{ modules: Modules }>(MODULES)
+
+  const lesson = useMemo(() => {
+    if (lessons) {
+      const lessonFromParam = lessons.find(
+        e => toNumber(extractLessonOrder(e.slug)) === toNumber(lessonId)
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:96]: shouldn't we be comparing the slugs of each lesson? `e.slug === lessonSlug`

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+          module={modules.find(module => module.id === selectedIndex)}
+        />
+      </div>
+    </div>
+  )
+}
+
+const Lessons = ({ data }: GetAppProps) => {
+  const router = useRouter()
+  const { pageName, lessonId } = router.query
```
Comment (songz) [inline pages/admin/lessons/[lessonId]/[pageName]/index.tsx:88]: should be `const {pageName, lessonSlug}`...

pages/admin/lessons/[lessonId]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonId]/[pageName]/index.tsx]: But `lessonSlug` is a number that comes from the parameter.

pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const router = useRouter()
+  const { pageName, lessonSlug } = router.query
+
+  const { lessons } = data
+  const { data: modulesData, refetch } = useQuery<{ modules: Modules }>(MODULES)
+
+  const lesson = useMemo(() => {
+    if (lessons) {
+      const lessonFromParam = lessons.find(
+        e => toNumber(extractLessonOrder(e.slug)) === toNumber(lessonSlug)
```
Comment (songz) [inline pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:94]: Keep it simple: `find(e => e.slug === lessonSlug)`

pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
```diff
@@ -0,0 +1,177 @@
...
+  }, [lessons, lessonSlug])
+
+  const modules = useMemo(
+    () => sortBy(get(modulesData, 'modules'), 'order'),
+    [modulesData]
+  )
+
+  const filteredModules = useMemo(
+    () => modules.filter(module => module.lesson.id === lesson.id),
+    [lesson.id, modules]
```
Comment (songz) [inline pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:115]: I think we could combine these 2 `useMemo`

```js
const filteredModules = useMemo( () => {
  return get(modulesData, 'modules', []).filter(module => module.lesson.id === lesson.id).sortBy(get(modulesData, 'modules'), 'order')
 }, [lesson.id, modluesData])
 ```

pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:94]: `e.slug` is `js0` and `lessonSlug` is `0`. We need to extract the last value from `e.slug` and turn it into a number to be able to compare it with `lessonSlug`.

Comment (songz) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:95]: 1. Using the whole slug makes sense to me
2. Using the lessonId makes sense to me

Using a part of the slug does not make sense to me, because it seems like it will be adding an additional constraint and potential for bugs in the future: All lesson slugs **must contain a number** from now on. Like in the future if we added supplemental lesson like `Dynamic Programming` then I think this implementation will be an issue.

Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:95]: Yeah, this makes sense. It'll actually be inconsistent with the way we've it in `/curriculum/jsX`

Comment (songz) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:95]: seems like our existing implementation just checks the slug directly instead of trying to convert to number: https://github.com/garageScript/c0d3-app/blob/master/pages/curriculum/%5BlessonSlug%5D.tsx#L36

songz (approved):

Issue #1891, opened 2022-05-23
Improve the website accessibility metrics


flacial:
Fix the critical issues from the [accessibility checker](https://www.accessibilitychecker.org/audit/?website=c0d3.com&flag=us) to increase the metric from 64% to 95%.

PR #1900, opened 2022-05-25
fix: Low accessibility metrics

flacial:
Closes https://github.com/garageScript/c0d3-app/issues/1891

This PR replaces `h3` headings with a `p` element and set the lang attribute to the `html` element as `en`.

The reason behind these changes can be found here:

- https://www.w3.org/WAI/tutorials/page-structure/headings/
- https://dequeuniversity.com/rules/axe/3.5/html-has-lang

Before:

image

After:

image

### DIFF
--- a/components/LandingPage.tsx
+++ b/components/LandingPage.tsx
@@ -66,7 +66,7 @@ const LandingPage: React.FC = () => {
               alt="Modern curriculum"
               width="163"
             />
-            <h4 className="mt-3 fw-bold">Modern Curriculum</h4>
+            <p className="mt-3 fw-bold h3">Modern Curriculum</p>
             <p className="mt-4 fw-light">
               First the curriculum will help you achieve mastery in JavaScript.
               Using JavaScript, you will learn about different branches of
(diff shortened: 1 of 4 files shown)


next.config.js
```diff
@@ -30,6 +30,10 @@ const moduleExports = {
   pageExtensions: ['tsx', 'js', 'jsx', 'mdx', 'ts'],
   images: {
     domains: ['cdn.discordapp.com']
+  },
```
anthonykhoa, line 33: what does this do?

flacial: It's used to config the languages the app will support and set a default one that will be used for all the routing.

components/LandingPage.tsx
```diff
@@ -98,7 +98,7 @@ const LandingPage: React.FC = () => {
               alt="Work Experience"
               layout="fixed"
             />
-            <h4 className="mt-3 fw-bold">Work Experience</h4>
+            <p className="mt-3 fw-bold h3">Work Experience</p>
```
anthonykhoa, line 101: Does using `p` instead of a header for this file cause any change in UI?

flacial: No, since it gets the `h3` header css from the `h3` classname.

SlyBouhafs (approved):

songz (approved):

anthonykhoa (approved):

PR #1902, opened 2022-05-28
fixed PTMono-Bold font file name

ggwadera:
This font was not loading correctly due to the wrong file name. Fixed it so it can load as expected.

image

This is what the navbar looked like with my default system font

image


flacial (approved):
Great catch. We thought it had something to do with the cache.

SlyBouhafs (approved):

songz (approved):

PR #1944, opened 2022-06-07
refactor: Turn `render` prop into a functional component

flacial:
Closes https://github.com/garageScript/c0d3-app/issues/1942


SlyBouhafs (approved):

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -1,18 +1,22 @@
-import React from 'react'
+import React, { FunctionComponent } from 'react'
 import styles from '../../../scss/adminLessonNav.module.scss'
 
 type Props = {
   tabs: { tabName: string; urlPageName: string }[]
```
songz, line 5: Looking at the current state of this component, I'm noticing that `tabName` and `urlPageName` is not used for any logic within this component, so could we just have an object that could be anything?

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -1,18 +1,22 @@
...
+const AdminLessonNav: React.FC<Props> = ({ tabs, Component }: Props) => {
   return (
     <div>
       <div className={styles.lessons_tabsNav}>
         <div className={styles.lessons__tabsNav__nav}>
           {tabs.map((tab, i) => (
-            <React.Fragment key={i}>{render(tab)}</React.Fragment>
+            <Component
+              key={i}
+              tabName={tab.tabName}
```
songz, line 17: Adding on to the comment above, could we do `<Component key={i} tab={tab} />` This makes more sense to me because `tabName` and `urlPageName` is not used for any logic in here so we can make it generic

components/admin/lessons/AdminLessonNav.tsx
```diff
@@ -1,18 +1,22 @@
-import React from 'react'
```
songz, line 1: I think `AdminLessonSideNav` is a better name for this component because its a bit more clear. Whenever I see `AdminLessonNav` I think about the horizontal nav.
Screen Shot 2022-06-07 at 2 31 39 PM

PR #1957, opened 2022-06-09
Feat: Add Exercise preview component

### LINKED ISSUE #1819: DOJO: Create exercise preview component
Create the following [component](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=311%3A2300). 

image

SlyBouhafs:
This PR closes #1819 
It will add the Exercise preview component which will allow us to preview Exercises once the Exercise editor page is created.

Screen Shot 2022-06-09 at 17 27 52


components/ExercisePreview.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import styles from '../scss/exercisePreview.module.scss'
+
+type exerciseProps = {
```
flacial, line 4: NIT: What about naming it `Exercise` since it's not a prop for a component but a type of prop.

components/ExercisePreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <div className={`card shadow-sm ${classesList}`}>
+      <div className="card-body pt-3 pb-3 ">
+        <p className="card-text">
+          <p className={`${styles['exercise__p__bold']}`}>Problem</p>
+          <pre className={`${styles['exercise__pre__bg']} rounded px-4 py-4`}>
+            {exercise.question}
+          </pre>
+          <p className={`${styles['exercise__p__bold']}`}>Choose an answer</p>
+          {exercise.answers.map((answer, index) => (
```
flacial, line 28: What if there are 100 answers? Would it still look good?

SlyBouhafs: I guess with 100 answers it wouldn't look good at all, i guess i should probably give it a minimum height.

songz: would it make sense to call these options instead of an answer? 

`Choose an answer` <- this prompt is fine.

But `exercise.answers.map` <- this could cause naming issues because every exercise has one correct answer, so when the user picks from this list each element is technically an option for the user, not the "answer"

SlyBouhafs: > But `exercise.answers.map` <- this could cause naming issues because every exercise has one correct answer, so when the user picks from this list each element is technically an option for the user, not the "answer"

Yeah, sure! I'll use `options`.

songz (approved):

Issue #1935, opened 2022-06-06
Next-Auth: Replace GraphQL login/signup with next-auth


flacial:
Replace:

- Logout method to use next-auth `signOut`
- Login/Signup pages to use next-auth `signIn`

PR #1985, opened 2022-06-14
Next-Auth: Replace login/signup with next-auth

flacial:
Closes https://github.com/garageScript/c0d3-app/issues/1935

### Changed files in detail

- `success.tsx`: We were using a set of middlewares to populate the user from. We now use the next-auth [`getSession`](https://next-auth.js.org/getting-started/client#getsession) method to retrieve the session.
- `login.tsx` and `signup.tsx`: We used the Apollo mutations `signup` and `login` to start the flow. This has been changed with the next-auth [`signIn`](https://next-auth.js.org/getting-started/client#signin) method.
- `logoutContainer`: We used the Apollo mutation `logout` to log out the user. We now use [`signOut`](https://next-auth.js.org/getting-started/client#signout) method by next-auth.
- `nextAuth` (session callback): Updated the `session` callback to set the `session.user` as the user from the database. This callback runs every time we call `getSession` or `useSession`. The reason is that the data in the session stays outdated if one of them got updated from the client, such as if the user connected to discord.

### Notes

- Please try to login/signup, connect to discord and login with discord. Test it from [here](https://c0d3-172gxcn3g-c0d3-prod.vercel.app)
- There are a lot of changed files. Do you see how it can be broken down into small parts? I tried to think of it, but they're connected.

### Engineering changes 🔧 

- Use `getSession` and `useSession` to get the user's session.


songz:
> Do you see how it can be broken down into small parts? I tried to think of it, but they're connected.

I would have sent up a PR for adding password to our Signup flow. Then after that a PR for switching to nextAuth.

flacial:
> > Do you see how it can be broken down into small parts? I tried to think of it, but they're connected.
> 
> I would have sent up a PR for adding password to our Signup flow. Then after that a PR for switching to nextAuth.

Done in https://github.com/garageScript/c0d3-app/pull/2121

songz (approved):

Issue #1990, opened 2022-06-16
DOJO(admin) Manage exercises tab


flacial:
Create this page https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4588

Things need to be done before creating the page:

- ~~Create the [Exercise card](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4604)~~
- Create a [search/filtering input](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=906%3A11307)
- ~~Add `flagged` and `flaggedReason` columns to the `Exercise` model~~

evo777:
Figma is good.

flacial:
Filtering box and pagination are what's remaining.

PR #1993, opened 2022-06-17
feat: Add flag columns to Exercise

flacial:
Related to #1990

This PR adds the `flagged`, `flaggedBy`, and `flaggedReason`. `flagged` refers to exercises that were flagged by students or mentors to be reviewed by the Admin. `flaggedReason` is the reason for the flag. `flaggedBy` is the user who flagged it.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -176,18 +176,18 @@ model Module {
 }
 
 model Exercise {
-  id          Int      @id @default(autoincrement())
-  createdAt   DateTime @default(now()) @db.Timestamptz(6)
-  updatedAt   DateTime @updatedAt @db.Timestamptz(6)
-  author      User     @relation(fields: [authorId], references: [id])
+  id          Int       @id @default(autoincrement())
+  createdAt   DateTime  @default(now()) @db.Timestamptz(6)
+  updatedAt   DateTime  @updatedAt @db.Timestamptz(6)
+  author      User      @relation(fields: [authorId], references: [id])
   authorId    Int
-  module      Module   @relation(fields: [moduleId], references: [id])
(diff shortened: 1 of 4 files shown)


prisma/schema.prisma
```diff
@@ -187,6 +187,8 @@ model Exercise {
   answer      String
   testStr     String?
   explanation String?
+  flagged     Boolean?
```
JasirZaeem, line 190: instead of making flagged optional thereby giving it essentially 3 possible values (null, false, true) would be better to make it required with a default value of false
e.g. `flagged   Boolean @default(false)`

Alternatively, make the field flagged_at, optional and of the type date, where null means it is not flagged, and a timestamp means it is flagged and the timestamp is the time it was flagged at.
e.g. `flagged_at DateTime?`

songz: `flagged_at` is good suggestion thank you

flacial: Sounds good!

JasirZaeem:
The flaggedBy column you had proposed would be useful I suggest you add that in this PR as well.

flacial:
> The flaggedBy column you had proposed would be useful I suggest you add that in this PR as well.

I had some issues with having two columns with User type. I'll look into Prisma doc and add it.

prisma/schema.prisma
```diff
@@ -156,6 +156,7 @@ model User {
   modules                   Module[]
   exercises                 Exercise[]
 
+  Exercise Exercise[] @relation("flaggedBy")
```
JasirZaeem, line 159: Something like `flaggedExercises` will be a more suitable and descriptive name for this field.

JasirZaeem (approved):

songz (approved):

Issue #1975, opened 2022-06-13
DOJO(mentor): Create exercises page


flacial:
**Create the following [page](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1046%3A5849)**

Required components to be created:

- [x] [Preview card](https://github.com/garageScript/c0d3-app/issues/1819)

Components to be created within the page:

- [x] Inputs form
- [x] Dropdown to switch the lesson module

Misc:

- The page will be accessed through `/mentor/[lessonSlug]/[moduleSlug]/exercise` (could be discussed with engs)

evo777:
Good feature.

PR #1998, opened 2022-06-19
feat: Create Exercise card

flacial:
Related to #1975 
Depends on https://github.com/garageScript/c0d3-app/pull/2001

Merging this PR will create the Exercise card component that will have the following functionalities:

- Copy user discord username
- Copy user email
- Remove exercise flag (unflag)
- Remove exercise

Image:

image


components/admin/lessons/AdminLessonExerciseCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+`
+
+type FooterProps = {
+  exercise: Exercise
+  onRemove?: (id: number) => void
+  onUnflag?: (id: number) => void
+}
+
+const Footer = ({ exercise, onRemove }: FooterProps) => {
+  const [deleteExercise, { loading }] = useMutation(DELETE_EXERCISE)
```
anthonykhoa, line 76: Shouldn't there be a `useDeleteExercise` hook you can use from graphql/index file?

flacial: I thought so too, but it doesn't seem to exist. I'm not certain of the requirements for a resolver to be auto-generated as a hook.

anthonykhoa: From my understanding, its not that codegen gql generates resolvers as a hook, but rather codegen generates us hooks to use in place of `useMutation`. For these hooks, the functionality should be the same as `useMutation`, except you don't have to input the query yourself. 

Example of graphql/index code that inputs query for us(useGetApp):
https://github.com/garageScript/c0d3-app/blob/master/graphql/index.tsx#L2851

I think the problem might be that you haven't generated a `useDeleteExerciseMutation` hook yet. Please look into trying to use a `useDeleteExerciseMutation` hook that codegen gql generates for us.

flacial: > I think the problem might be that you haven't made codegen generate a useDeleteExerciseMutation hook yet.

I think it's because I didn't add the flagged fields to the `Exercise` type. Therefore, it couldn't generate the hook.

songz (changes requested):
set timestamps on server side

__tests__/__snapshots__/storyshots.test.js.snap
```diff
@@ -127,6 +127,134 @@ exports[`Storyshots Components/AdminLessonCard Basic 1`] = `
...
+      <span>
+        UNFLAG EXERCISE
+      </span>
+    </button>
+  </div>
+</div>
+`;
+
+=======
+>>>>>>> upstream/master
```
songz, line 257: this looks like an unresolved merge conflict

songz (approved):

components/admin/lessons/AdminLessonExerciseCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  return (
+    <div className={styles.card__section}>
+      <h3 className={styles.card__section__header}>{header}</h3>
+      <p className={styles.card__section__desc}>{paragraph}</p>
+      {explanation && (
+        <>
+          <button
+            className={styles['card__section--explanation']}
+            onClick={() => setShow(!show)}
```
JasirZaeem, line 62: When updating state based on previous value of the state it would be better pass a function the setState to get access to the latest value of the state and prevent hard to find bugs because of race conditions, batching, etc. https://beta.reactjs.org/apis/usestate#updating-state-based-on-the-previous-state

`setShow(prevShow => !prevShow)` something like this for example.

components/admin/lessons/AdminLessonExerciseCard.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <span
+            className={`${styles.card__header__badge} ${styles.card__header__badge__flagged}`}
+          >
+            FLAGGED
+          </span>
+        )}
+        <span
+          className={`${styles.card__header__badge} ${styles.card__header__badge__module}`}
+        >
+          {toUpper(exercise.module.name)}
```
JasirZaeem, line 28: nit: since this is purely for appearance we shoul consider doing such modifications using the css text-transform property instead of javascript.

components/admin/lessons/AdminLessonExerciseCard.test.js
```diff
@@ -0,0 +1,187 @@
...
+  it('Should not show the explanation if not provided', () => {
+    expect.assertions(1)
+
+    render(
+      <MockedProvider mocks={mocks}>
+        <Component user={user} exercise={{ ...exercise, explanation: '' }} />
+      </MockedProvider>
+    )
+
+    expect(screen.queryByText('Show explanation')).toBeFalsy()
```
JasirZaeem, line 185: nit: it might to be clearer to test that element does not exist
```javascript
expect(screen.getByText('Show explanation')).not.toBeInTheDocument()
```

JasirZaeem (approved):

PR #2000, opened 2022-06-20
fix: Remove parameters from the Breadcrumbs

### LINKED ISSUE #1997: Remove params from the Breadcrumb component items
Remove any param that comes after the dropdown:

image

flacial:
Closes #1997

After merging this PR, the params (e.g, `[lessonSlug]`) will no longer appear after the dropdown that allows switching between lessons.

image


anthonykhoa:
Please include an image that shows the change in UI due to these changes

components/Breadcrumbs.tsx
```diff
@@ -110,12 +110,19 @@ const Breadcrumbs = ({
...
-        {breadcrumbs.map((breadcrumb, index) => (
-          <React.Fragment key={`${breadcrumb.title}-${index}`}>
-            <Breadcrumb breadcrumb={breadcrumb} index={index} />
-            {lowerCase(breadcrumb.title) === 'lessons' && <LessonsDropdown />}
-          </React.Fragment>
-        ))}
+        {breadcrumbs
+          .slice(
+            0,
+            breadcrumbs.findIndex(
```
songz, line 116: If you don't plan to use something, don't set it in the first place. Therefore, I think this logic should be handled in line 49.

songz (approved):

components/Breadcrumbs.tsx
```diff
@@ -53,7 +53,12 @@ const Breadcrumbs = ({
       }
     })
 
-    setBreadcrumbs(pathArray)
+    setBreadcrumbs(
+      pathArray.slice(
+        0,
+        pathArray.findIndex(breadcrumb => breadcrumb.title === 'lessons') + 1
+      )
+    )
```
songz, line 61: since you are making a change to this file, why do we need a state for breadcrumbs? Why not just:

```js
const Breadcrumbs = ({...}) => {
  const [_, ...urlPath] = router.asPath.split('/')
  const pathArray = urlPath.map((path, i) => {
      return {
        title: path,
        href: `/${urlPath.slice(0, i + 1).join('/')}`
      }
    })
   const validPaths = ...
```

anthonykhoa:
Not blocking, but I wanted to discuss the potential of `breadcrumbs` to be used in other c0d3 pages in the future. I think breadcrumbs is a really nice feature to have for users, and I can see it being re-used for other projects in the future when there are similar situations as this one where there is a hierarchical structure(`e.g home -> admin -> lessons -> foundations...`).

@flacial @songz Any thoughts on the potential of breadcrumbs to be used for other pages in the future? IMO if the potential is not high, then lets keep this component specific to the Admin Lessons page. But if there is a high potential, I think there should be a change made to make this component more generic so as to make it less specific to the Admin Lessons page. The reasoning is that if we keep this component specific to the Admin Lessons page, then more tech debt will build up from any other updates that this component receives.

anthonykhoa (approved):
Looks good, before merging just follow up with adding a comment in the code as requested by @songz

songz:
> Not blocking, but I wanted to discuss the potential of `breadcrumbs` to be used in other c0d3 pages in the future. I think breadcrumbs is a really nice feature to have for users, and I can see it being re-used for other projects in the future when there are similar situations as this one where there is a hierarchical structure(`e.g home -> admin -> lessons -> foundations...`).
> 
> @flacial @songz Any thoughts on the potential of breadcrumbs to be used for other pages in the future? IMO if the potential is not high, then lets keep this component specific to the Admin Lessons page. But if there is a high potential, I think there should be a change made to make this component more generic so as to make it less specific to the Admin Lessons page; the reasoning is that if we keep this component specific to the Admin Lessons page, then more tech debt will build up from any other updates that this component receives.

No need to make it generic for now because only admin page uses it. When the designer creates other pages that uses breadcrumbs, we can generalize it at that time.

anthonykhoa:
> No need to make it generic for now because only admin page uses it. When the designer creates other pages that uses breadcrumbs, we can generalize it at that time.

What I'm getting from this reply is that the potential for breadcrumbs to be used in the future is currently unknown; therefore, we will wait to see if the designer will make more breadcrumb designs before concluding that this component needs to be generic. Is this a correct assessment?

components/Breadcrumbs.tsx
```diff
@@ -41,22 +41,21 @@ const Breadcrumbs = ({
...
+  const [_, ...urlPath] = router.asPath.split('/')
+
+  const pathArray = urlPath.map((path, i) => {
+    return {
+      title: path,
+      href: `/${urlPath.slice(0, i + 1).join('/')}`
+    }
+  })
+
+  const breadcrumbs = pathArray.slice(
```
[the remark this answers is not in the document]
flacial: Yeah, I agree!

Issue #1959, opened 2022-06-09
DOJO(admin/page): Create lessons home page


flacial:
Create the following [page](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4488).

To resolve this issue, you must create the following component(s):

- [Lesson card](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=204%3A4491)

After creating the components, in the next PR, create the page. ADD NEW LESSON and EDIT THE LESSON buttons won't work until the next [pages](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=906%3A4589) are implemented.

evo777:
Nice.

flacial:
Duplicate of https://github.com/garageScript/c0d3-app/issues/1656

HS-90:
new lessons page merged under path '/admin/lessons/new', but im leaving the issue open for now since we should update/finalize the links in the page once other related admin pages like exercises page, pending questions, etc. are finished.

flacial:
Closed in https://github.com/garageScript/c0d3-app/pull/2538

PR #2023, opened 2022-06-28
Dojo(admin) Create lessons home page. Part 2: Lessons Page

HS-90:
**Title:** enhancement : Create a lessons homepage with Lesson Card components for Dojo(Admin).

**Description:**

Related to #1959

Create a lessons homepage with the ability to Add New Lessons, Edit Lessons, view questions from students for each lesson.

In this PR. The Lessons page is created. Please see screenshots below. 

Included in the PR is an alteration to the `AdminLessonCard` component CSS. The description section of the card was assigned a `min-height` to make each card look more uniform when aligned against eachother in the Lessons page.

**EDIT 7/7/22:** In conjunction with this PR, a seperate PR for the generated query `getFlaggedExercises` has been made #2055 . In addition, a custom `Lesson` type has been included in both  `AdminLessonCard` and the new lessons page. This was done so that the data types are consistent when passing them to each component. 

New admin lessons page screenshot:
image


__tests__/pages/admin/lessons/new/lessons.test.js
```diff
@@ -0,0 +1,160 @@
...
+    const { container, getByText, getAllByText } = render(
+      <MockedProvider mocks={mocks.lessonsLoaded} addTypename={false}>
+        <Lessons />
+      </MockedProvider>
+    )
+
+    // Used to wait for the query response to arrive
+    await act(async () => await new Promise(res => setTimeout(() => res(), 0)))
+
+    await waitFor(() => expect(container).toMatchSnapshot())
```
flacial, line 140: `waitFor` + `.toMatchSnapshot` result in unexpected snapshot changes. We should look for an alt.

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+              />
+            </div>
+          )
+          return prev
+        }, [])
+      )
+    }
+  })
+
+  React.useEffect(() => {
```
anthonykhoa, line 72: Can we deconstruct useEffect from the React object like how we do it for useState on line 1

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+    lessons && getExercises()
+  }, [lessons, exercisesData])
+
+  return (
+    <AdminLayout data={data} title="Admin Lessons Homepage">
+      <div className={styles.heading}>
+        <span className={styles.lessonText}>Lessons</span>
+        <span>
+          <Link href={'../../admin/lessons'}>
```
anthonykhoa, line 81: brackets are unnecessary here

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+}
+
+type Exercises = Exercise[]
+
+const Lessons: React.FC<GetAppProps> = ({ data }) => {
+  const { lessons } = data
+  const [lessonCards, setCards] = useState<[]>([])
+  const [getExercises, { data: exercisesData }] = useLazyQuery<{
```
anthonykhoa, line 35: Please use a hook that gql code-gen will generate for us. They generate `lazy` hooks as well.

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+    exercises: Exercises
+  }>(EXERCISES, {
+    onCompleted: async () => {
+      const exerciseMapping = await exercisesData?.exercises.reduce(
+        (prev: any, cur) => {
```
anthonykhoa, line 40: Typing as `any` defeats the purpose of using typescript

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+            prev[cur.module?.lesson?.title]++
+          }
+
+          return prev
+        },
+        {}
+      )
+
+      setCards(
+        lessons?.reduce((prev: any, cur: any, i) => {
```
anthonykhoa, line 55: same comment about `any` here

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+          prev.push(
+            <div key={i} className={styles.lessonCard}>
+              <AdminLessonCard
+                lesson={cur}
+                pendingFlaggedQuestions={
+                  exerciseMapping[cur.title] ? exerciseMapping[cur.title] : 'No'
```
anthonykhoa, line 61: `exerciseMapping[cur.title]` is repeated twice, make it into a variable

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+      title: string
+    }
+  }
+}
+
+type Exercises = Exercise[]
+
+const Lessons: React.FC<GetAppProps> = ({ data }) => {
+  const { lessons } = data
+  const [lessonCards, setCards] = useState<[]>([])
```
anthonykhoa, line 34: It is confusing for `lessonCards` and `setCards` to be different naming convention. Please use same naming convention for both, example `const [lessonCards, setLessonCards] = ...`

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+            }
+            prev[cur.module?.lesson?.title]++
+          }
+
+          return prev
+        },
+        {}
+      )
+
+      setCards(
```
anthonykhoa, line 54: We should not be setting jsx in a react state. State should only contain the minimal amount of data needed to represent your UI's state. Use data from the state in order to perform correct logic for what you want to render.

React is based on the concept of a function rendering elements that are the result of its current values. If you store jsx in a state, thats an anti-react pattern since what you are rendering doesn't depend on current values.

Storing jsx in a state also creates the risk of having stale elements that don't correspond to the current set of properties; although the risk of `stale element` problem doesn't appear to happening in this PR, it is still an increased complexity. The increased complexity of `storing jsx in a state`doesn't offer any benefits over simply storing data in a state to be used in rendering logic. It just makes things more complex thus harder to maintain

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+      flaggedAt
+      module {
+        lesson {
+          title
+        }
+      }
+    }
+  }
+`
+type Exercise = {
```
anthonykhoa, line 21: this should be declared in the queries folder 
https://github.com/garageScript/c0d3-app/tree/4e30845fb7b47ceca0a3e3c2733a61fb723d4776/graphql/queries

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { useState } from 'react'
+import { gql, useLazyQuery } from '@apollo/client'
+import { AdminLessonCard } from '../../../../components/admin/lessons/AdminLessonCard'
+import { withGetApp, GetAppProps } from '../../../../graphql'
+import { AdminLayout } from '../../../../components/admin/AdminLayout'
+import Link from 'next/link'
+import styles from '../../../../scss/adminLessonPage.module.scss'
+
+const EXERCISES = gql`
```
anthonykhoa, line 9: this should be declared in the queries folder
https://github.com/garageScript/c0d3-app/tree/4e30845fb7b47ceca0a3e3c2733a61fb723d4776/graphql/queries

pages/admin/lessons/new/index.tsx
anthonykhoa, line 21: Looks like we already have a type Exercise
https://github.com/garageScript/c0d3-app/blob/4e30845fb7b47ceca0a3e3c2733a61fb723d4776/graphql/index.tsx#L75

Is is possible use this type instead?

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+            </div>
+          )
+          return prev
+        }, [])
+      )
+    }
+  })
+
+  React.useEffect(() => {
+    lessons && getExercises()
```
anthonykhoa, line 73: What is the reason this useEffect is needed?

anthonykhoa: I'm guessing this is to handle the use case of the `lessons still loading` from getApp query. If that is the case, please try to use AdminLayout component to handle this case. It should have sufficient logic to handle this use case

__tests__/pages/admin/lessons/new/lessons.test.js
```diff
@@ -0,0 +1,160 @@
...
+import Lessons from '../../../../../pages/admin/lessons/new/index'
+import { MockedProvider } from '@apollo/client/testing'
+import { act, render, waitFor } from '@testing-library/react'
+import dummyLessonData from '../../../../../__dummy__/lessonData'
+import dummySessionData from '../../../../../__dummy__/sessionData'
+import dummyAlertData from '../../../../../__dummy__/alertData'
+import '@testing-library/jest-dom'
+import GET_APP from '../../../../../graphql/queries/getApp'
+
+const EXERCISES = gql`
```
anthonykhoa, line 12: This should be in the queries folder(look for my other comment about this, I provide link in it)

__tests__/pages/admin/lessons/new/lessons.test.js
```diff
@@ -0,0 +1,160 @@
...
+    expect.assertions(5)
+
+    const { container, getByText, getAllByText } = render(
+      <MockedProvider mocks={mocks.lessonsLoaded} addTypename={false}>
+        <Lessons />
+      </MockedProvider>
+    )
+
+    // Used to wait for the query response to arrive
+    await act(async () => await new Promise(res => setTimeout(() => res(), 0)))
```
anthonykhoa, line 138: The `res` function call in setTimeout doesn't need be explicitly called, just pass it as first parameter to setTimeout
```
await act(async () => await new Promise(res => setTimeout(res, 0)))
```

__tests__/pages/admin/lessons/new/lessons.test.js
```diff
@@ -0,0 +1,160 @@
...
+  test('should not render any components if lessons data does not load', async () => {
+    expect.assertions(3)
+
+    const { container, getByText, getAllByText } = render(
+      <MockedProvider mocks={mocks.lessonsNotLoaded}>
+        <Lessons />
+      </MockedProvider>
+    )
+
+    await act(async () => await new Promise(res => setTimeout(() => res(), 0)))
```
anthonykhoa, line 155: same comment as my other one for this file

pages/admin/lessons/new/index.tsx
flacial, line 40: `prev` could be `{ [key: string]: number }` instead of `any`. `cur` is the `Exercise`.

__tests__/pages/admin/lessons/new/lessons.test.js
```diff
@@ -0,0 +1,160 @@
...
+    expect(() => getAllByText('No Pending Questions')).toThrow()
+    expect(() => getByText('Arrays')).toThrow()
+    expect(container).toMatchSnapshot()
```
SlyBouhafs, line 158: You should avoid using `toMatchSnapshot`.

https://github.com/garageScript/c0d3-app/wiki/Engineering-Practices#testing

__tests__/pages/admin/lessons/new/lessons.test.js
```diff
@@ -0,0 +1,160 @@
...
+    )
+
+    // Used to wait for the query response to arrive
+    await act(async () => await new Promise(res => setTimeout(() => res(), 0)))
+
+    await waitFor(() => expect(getByText('3 Pending Questions')).toBeTruthy())
+    expect(getByText('2 Pending Questions')).toBeTruthy()
+    expect(getAllByText('No Pending Questions')).toBeTruthy()
+    expect(getByText('Arrays')).toBeTruthy()
+    expect(container).toMatchSnapshot()
```
SlyBouhafs, line 144: You should avoid using `toMatchSnapshot`.

You can read more about testing practices here: https://github.com/garageScript/c0d3-app/wiki/Engineering-Practices#testing

scss/adminLessonPage.module.scss
```diff
@@ -0,0 +1,38 @@
...
+  font-weight: 400;
+}
+
+.button {
+  width: 214px;
+  height: 49px;
+  font-weight: 700;
+}
+
+.container__lessonCard__outer {
```
SlyBouhafs, line 26: `.container__lessonCard__outer` and `.container__lessonCard__inner` have the same properties. Should we group them together instead?

pages/admin/lessons/new/index.tsx
HS-90: Can i get some additional guidance on how `AdminLayout` can handle this? Im already passing in the `data` that `lessons` was deconstructed from to the `data` prop in `AdminLayout`

anthonykhoa: @HS-90 Adminlayout returns a loading spinner when data is still loading. It will only return its child elements when data is done loading

https://github.com/garageScript/c0d3-app/blob/3a064c05b09b9ec65049904ddef3457a48e4b9cd/components/admin/AdminLayout.tsx#L36

pages/admin/lessons/new/index.tsx
anthonykhoa: Actually, instead of a making variable, it would be better to simplify this to `exerciseMapping[cur.title] || 'No'`, since your use of ternary operator is doing what the logical operator `||` already does

pages/admin/lessons/new/index.tsx
anthonykhoa: If an  array of objects is the minimum data needed to render `lessonsCardsComponents`(as per your example), then yeah that sounds right

pages/admin/lessons/new/index.tsx
HS-90: @anthonykhoa is there an example in the codebase for this? I've looked at the AdminLayout file and see the logic, but I'm still having a hard time applying it.

anthonykhoa: I am not sure about other instances(I haven't contributed in awhile, so i'm less familiar with the codebase),  but you can try searching `AdminLayout` and see all the files that use it. Looks like there are only 7 results, so there's not much to have to peruse through:
https://github.com/garageScript/c0d3-app/search?q=adminlayout

`it doesn't wait for the lessons data to load on a reliable basis unless we apply the useEffect`. <--
That is because this `Lesson` component is not a child element of AdminLayout, instead it just uses the component in its return statement.

What I would try doing(and optimally, I would try to think of multiple proposals on how to achieve a solution to compare pros/cons) is:

1) Create a new component `component X`(needa change name of course) to do all that is needed to do after `lessons data` is fully loaded. This way, the new component will not have to worry about lessons data not being loaded. 

2) Use component `X` as a child component of AdminLayout component
```
<AdminLayout>
  <X />
</AdminLayout>
```

With this method, you can don't have to use a lazy hook either

anthonykhoa: Something else that might work is:
1) Create new component `component X`. Except this time, you use a codegen graphql hook `withGetExercises`(im just guessing the name, idk what was actually generated)

2) Use Component `X` as a child component of AdminLayout component
```
<AdminLayout>
  {withGetExercises()(X)}
</AdminLayout>
```

Also,I haven't made POCS for any of these proposals I gave, so you would have to test to confirm if they will work

anthonykhoa: Alternatively, you can turn adminlayout itself into a container

example container
https://github.com/garageScript/c0d3-app/blob/master/containers/withQueryLoader.tsx

HS-90: > I am not sure about other instances(I haven't contributed in awhile, so i'm less familiar with the codebase), but you can try searching `AdminLayout` and see all the files that use it. Looks like there are only 7 results, so there's not much to have to peruse through: https://github.com/garageScript/c0d3-app/search?q=adminlayout
> 
> `it doesn't wait for the lessons data to load on a reliable basis unless we apply the useEffect`. <-- That is because this `Lesson` component is not a child element of AdminLayout, instead it just uses the component in its return statement.
> 
> What I would try doing(and optimally, I would try to think of multiple proposals on how to achieve a solution to compare pros/cons) is:
> 
> 1. Create a new component `component X`(needa change name of course) to do all that is needed to do after `lessons data` is fully loaded. This way, the new component will not have to worry about lessons data not being loaded.
> 2. Use component `X` as a child component of AdminLayout component
> 
> ```
> <AdminLayout>
>   <X />
> </AdminLayout>
> ```
> 
> With this method, you can don't have to use a lazy hook either

This method worked. It actually made my old method look a little over-complicated/over-engineered. I want to try the  other methods you suggested before updating this PR, but this was a good way to do it. Also good learning experience for me.

songz (approved):

pages/admin/lessons/new/index.tsx
```diff
@@ -1,94 +1,91 @@
...
-        },
-        {}
-      )
+const LoadedLessonCards = ({
+  lessonsData,
+  exercisesData
+}: LoadedLessonCardsProps) => {
+  const exerciseMapping = exercisesData?.exercises.reduce(
+    (acc: { [key: string]: number }, cur) => {
+      const curTitle = _.get(cur, 'module.lesson.title', undefined)
```
flacial, line 43: Here `get` was used, but in line `41` `?.` was used.

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const LoadedLessonCards = ({
+  lessonsData,
+  exercisesData
+}: LoadedLessonCardsProps) => {
+  const exerciseMapping = exercisesData?.exercises.reduce(
+    (acc: { [key: string]: number }, cur) => {
+      const curTitle = _.get(cur, 'module.lesson.title', undefined)
+      if (cur?.flaggedAt && curTitle) {
+        if (!acc[curTitle]) {
```
anthonykhoa, line 45: Can we simplify this down to
```
acc[curTitle] = (acc[curTitle] || 0) + 1
```

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  challenges: Array<{
+    id: number
+    title: string
+    description: string
+    order: number
+  }>
+}
+
+type LoadedLessonCardsProps = {
+  lessonsData: Lesson[] | undefined
```
anthonykhoa, line 33: undefined is unnecessary, just use typescript `?`

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  exercisesData: GetFlaggedExercisesQuery | undefined
```
anthonykhoa, line 34: same here

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  })
+
+  return (
+    <div className={styles.container__lessonCard}>{lessonCardComponents}</div>
+  )
+}
+
+const Lessons: React.FC<GetAppProps> = ({ data }) => {
+  const { lessons } = data
+  const { data: exercisesData } = useGetFlaggedExercisesQuery()
```
anthonykhoa, line 74: this lessons component doesnt use exercisesData,it just passes it down to a child component as a prop, loadedlessoncards. It makes more sense to for the useGetFlaggedExercisesQuery call to go in loadedlessoncards component, since thats where its return value is actually used

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  withGetApp,
+  GetAppProps,
+  useGetFlaggedExercisesQuery,
+  GetFlaggedExercisesQuery
+} from '../../../../graphql'
+import { AdminLayout } from '../../../../components/admin/AdminLayout'
+import Link from 'next/link'
+import styles from '../../../../scss/adminLessonPage.module.scss'
+
+type Lesson = {
```
anthonykhoa, line 14: dont we already have a Lesson type? It should be in the graphql/index file

components/admin/lessons/AdminLessonCard.tsx
```diff
@@ -1,11 +1,28 @@
...
+    title: string
+    description: string
+    order: number
+  }>
+}
 
 type Props = {
   lesson: Lesson
-  pendingFlaggedQuestions: number
+  pendingFlaggedQuestions: number | string
```
anthonykhoa, line 25: Why is it number or string?

pages/admin/lessons/new/index.tsx
HS-90: The `Lesson` type from `graphql `doesn't match the lessons object type from the `getApp` query so it will throw a type-error . It needs to match the type shown here: https://github.com/garageScript/c0d3-app/blob/a5ad3d3b10112324ae9cc35c080c57706c1ba2a4/graphql/index.tsx#L697 

,but without the array. So my current solution was to create a Lessons object mirroring what's shown in the link.

components/admin/lessons/AdminLessonCard.tsx
HS-90: So when there are no questions, it displays 'No'. I can changed it to just say `0` if that is what's preferred, then we can just leave it as a number type

anthonykhoa: Yeah lets do that. For this specific context, It seems weird to me that this field can be either be number or string. Lets simplify things and change it so that it can only be number

anthonykhoa: If you want `No` to be displayed when there are no pending questions, then do this on line 40:
```
{pendingFlaggedQuestions || 'No'} Pending Questions
```

This will reduce complexity because we are keeping the type as `number`. Future developers who touch this code won't have to think about whether to pass a number or string to `AdminLessonCard` -- they'll just simply pass a number.

pages/admin/lessons/new/index.tsx
anthonykhoa: Hmm looks the same to me. The Lesson type in the index file has currentUsers and users as fields, but those are optional fields so that shouldn't throw any errors
https://github.com/garageScript/c0d3-app/blob/a5ad3d3b10112324ae9cc35c080c57706c1ba2a4/graphql/index.tsx#L90

So when you use this Lesson type instead of the one you created, you say there is a type error?

anthonykhoa: And when you use the Lesson type from index file , you use it as `Lesson[]` right?

pages/admin/lessons/new/index.tsx
HS-90: Im going to change both to `?. `optional chaining. Line 41 was giving me issues with lodash("`error: object is possibly undefined"`)

pages/admin/lessons/new/index.tsx
HS-90: image
i use `Lesson[]` like shown above when when using the `Lesson` type directly imported from `graphql/index`. 

These are the main errors showing :
 `Type '{ __typename?: "Challenge" | undefined; id: number; title: string; description: string; order: number; }[]' is not assignable to type 'Challenge[]'.`
   `Property 'lessonId' is missing in type '{ __typename?: "Challenge" | undefined; id: number; title: string; description: string; order: number; }' but required in type 'Challenge'.`

It appears the `Challenge` field from the imported `Lesson` type is not compatible with the `Lesson` type defined in the `GetAppQuery`
image

components/admin/lessons/AdminLessonCard.tsx
HS-90: I ended up going with just showing 0.

pages/admin/lessons/new/index.tsx
flacial: > Im going to change both to `?. `optional chaining. Line 41 was giving me issues with lodash("`error: object is possibly undefined"`)

I think to solve it, you could do 
```ts
const exercises = _.get(exercisesData, 'exercises')

const exerciseMapping = exercises && exercises.reduce...
```

anthonykhoa: @HS-90 i dont think its a problem with lodash, I think its a problem with how you set up your logic.

I initially speculated the problem might have been that lodash `_.get` did not cases where first parameter has value of undefined. But then I went to lodash.com and tried the following in my console:
```
a = undefined
b = _.get(a, 'ss', 'potato')
console.log(b) // logs potato
```

So it looks like _.get can handle when the first parameter has a value of undefined. You shouldn't have to do @flacial 's suggestion either of `exercises && exercises.reduce`. Thats an an anti pattern since we already have `_.get` to set default values for nested properties that are not found
If the problem is with `exercises` possibly being undefined, just set its default value as `[]`
```
const exercises = _.get(exercisesData, 'exercises', [])
```

pages/admin/lessons/new/index.tsx
anthonykhoa: I don't have much time to think of further proposals, but here are two proposals without needing to rewrite the entire Lesson type in this file:

1) Do what @flacial suggested and add lessonId to getAppQuery
2) Try to use typescript utils to delete fields you don't want

Example:
```
interface Todo {
  title: string;
  description: string;
  completed: boolean;
  createdAt: number;
  cat: Cat
}

interface Cat {
  nose: string
  removeMePls: string
}

type removedCat = Omit<Cat, "removeMePls">;


type TodoInfo = Omit<Todo, "cat"> & { cat: removedCat }

const todoInfo: TodoInfo = {
  title: "Pick up kids",
  description: "Kindergarten closes at 5pm",
  completed: false,
  createdAt: 1615544252770,
  cat: {
    nose: 'lol'
  },
};
```

I'm not super familiar with typescript, so i'm not sure this is how to best use typescript for this situation. There are maybe other ways more appropriate to implement for this situation than my given proposal.

HS-90: @anthonykhoa I ended up going with this stackoverflow: https://stackoverflow.com/questions/47914536/use-partial-in-nested-property-with-typescript

Basically modify the `Lessons` type from graphql to be able to accept partial keys in the `AdminLessonCard` file.
image

pages/admin/lessons/new/index.tsx
HS-90: If you set the default value to empty array `[]` while using `reduce`, Typescript will throw an error https://github.com/microsoft/TypeScript/issues/36390#issuecomment-852533093 

However, `forEach` works in this case. So I set it up a little differently than before:

image

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+const LoadedLessonCards = ({ lessonsData }: LoadedLessonCardsProps) => {
+  const { data: exercisesData } = useGetFlaggedExercisesQuery()
+  const exerciseMapping: { [key: string]: number } = {}
+  const exercises = _.get(exercisesData, 'exercises', [])
+
+  exercises.forEach(e => {
+    const curTitle = _.get(e, 'module.lesson.title', undefined)
+    if (e?.flaggedAt && curTitle)
```
flacial, line 27: This should be `_.get`

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      exerciseMapping[curTitle] = (exerciseMapping[curTitle] || 0) + 1
+  })
+
+  const lessonCardComponents = lessonsData?.map((e, i) => {
```
flacial, line 31: This should be `lessonsData &&`

pages/admin/lessons/new/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+type LoadedLessonCardsProps = {
+  lessonsData?: LessonType[]
+}
+
+const LoadedLessonCards = ({ lessonsData }: LoadedLessonCardsProps) => {
+  const { data: exercisesData } = useGetFlaggedExercisesQuery()
+  const exerciseMapping: { [key: string]: number } = {}
+  const exercises = _.get(exercisesData, 'exercises', [])
+
+  exercises.forEach(e => {
```
flacial, line 25: `reduce` or `map` didn't work?

songz (approved):

flacial (approved):

Issue #1088, opened 2021-08-28
MDX component improvements 


tomrule007:
Just quickly going over the difference between our self hosted content and the old notion docs I saw some style differences that I think could be improved. 

1) Spoiler collapsed state should have some bottom margin to better separate itself from the content below it.

![spoilerMargin](https://user-images.githubusercontent.com/17365077/131230898-8dd682f1-26f0-475b-8d38-0d4d08f843ba.png)

2) Spoiler expanded state should be indented to show it belongs to that spoiler

![spoilerDetailsIndent](https://user-images.githubusercontent.com/17365077/131230925-351fe1ec-12ab-4341-abb9-a73d99defadd.png)

3) Code block should not have so much bottom margin. Somehow the code blocks are double wrapped in `<pre>` tags that each have their own bottom margins. (The top margin might want to be tweaked as well)

![codeblockPadding](https://user-images.githubusercontent.com/17365077/131230950-7d996e61-25be-4130-9e07-ca9bb58dc3ea.png)

4) Quote block lightbulb icon should not wrap to its own line on small screens. Smaller icon maybe better as well?

![quoteIconOnSingleLine](https://user-images.githubusercontent.com/17365077/131230978-4f6dd569-b93a-46dd-b8a7-7ead6f000111.png)

tomrule007:
Also you might be able to spot more improvements by comparing our notion docs to the self hosted ones
https://www.c0d3.com/curriculum/js0/primitive_data_and_operators
https://www.notion.so/JS-0-Foundations-a43ca620e54945b2b620bcda5f3cf672

flacial:
A redesign is currently under work for this page.

PR #2088, opened 2022-07-14
WIP(1): Update lesson content styles

flacial:
Related to #1088 

This PR solves:

- "Spoiler collapsed state should have some bottom margin to better separate itself from the content below it."
  - Before
  image
  - After 
  image

- "Spoiler expanded state should be indented to show it belongs to that spoiler"
  - Before
  image
  - After 
   image


SlyBouhafs:
I think the spacing and indent is a little bit too big, look how notion does it:


Screen Shot 2022-07-14 at 14 19 51
Screen Shot 2022-07-14 at 14 20 02

The bottom margin is much more smaller and the indent aligns perfectly with the "A" of Answer.

components/mdx/Spoiler/spoiler.module.scss
```diff
@@ -0,0 +1,3 @@
+.content {
+  margin-left: 0.9rem;
```
SlyBouhafs, line 2: This should be `1rem` instead.

flacial: `1rem` goes a bit away from the letter "A", I wanted it to be close to perfection.

JasirZaeem (approved):

Issue #2081, opened 2022-07-13
DOJO(admin/page): Edit/Add challenges


flacial:
Create the following [page mock](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=311%3A2144).

The page is very similar to the [modulesTab](https://github.com/garageScript/c0d3-app/blob/a5ad3d3b10112324ae9cc35c080c57706c1ba2a4/pages/admin/lessons/%5BlessonSlug%5D/%5BpageName%5D/index.tsx#L64-L81). So, you might use the same components or refactor it, so you won't have to write the same code twice.

flacial:
Resolved

PR #2095, opened 2022-07-16
feat: Refactor AdminLessonInputs to handle challenges

flacial:
Related to #2081 

This PR makes it possible for the `AdminLessonInputs` to add/update a challenge by refactoring the code and adding the mutations required to perform the mentioned actions.

Side changes:
- Create a folder for `AdminLessonInputs`
- Rename `AdminModuleInputs` to `AdminLessonInputs`


JasirZaeem:
For moving existing files into their new respective component folders, and/or renaming them, it would be better to use `git mv` so that git can keep track of their commit histories across renaming and moving . It seems git is unable to properly track history for changed files in this PR. Retaining correct history will help to solve future bugs and find out reasons of past decisions.

components/admin/lessons/AdminLessonInputs/AdminLessonInputs.test.js
```diff
@@ -0,0 +1,732 @@
+import React from 'react'
```
JasirZaeem, line 1: Use `git mv` to retain commit history.

components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx
```diff
@@ -0,0 +1,286 @@
+import React, { useEffect, useState } from 'react'
```
JasirZaeem, line 1: Use `git mv` to retain commit history.

PR #2136, opened 2022-07-30
Cross-Env for windows commands

### LINKED ISSUE #1976: Add a `yarn test` script for Windows
Some of our engineers use Windows, and setting environment variables in PowerShell or CMD is different from Linux.
`TZ=UTC` has to become `set TZ=UTC` on Windows.

To resolve this issue, add scripts with the following format `action:windows` for each script that sets an env variable.

Example:

- `test:windows: "set TZ=UTC && yarn test --coverage --runInBand"`

HS-90:
**Title:** enhancement : Cross-Env for windows commands

**Description:**
Closes #1976 

This PR will solve the issue of certain scripts not working on Windows. Cross-env allows for a single command line assigned to scripts to work seamlessly between operating systems. Previously, windows users would have to manually adjust an individual script due to windows setting environment variables differently.

Taking `yarn test `as an example:

Windows users would have to change the `test` script in their local machines from `TZ=UTC jest --coverage --runInBand ` to `set TZ='UTC' & jest --coverage --runInBand`. 

With cross-env, we just need to set it to the following for all users, and Windows/Mac/Linux users should all have their environment variables set properly:

`cross-env TZ=UTC jest --coverage --runInBand`


flacial (approved):
This is amazing!

anthonykhoa (approved):
nice PR description

SlyBouhafs (approved):

songz:
`cross-env` should be a dev dependency, not dependency.

PR #2166, opened 2022-08-08
withUserContainer helper

### LINKED ISSUE #2017: Create withUserContainer helper method
Protip for doing this:

1. Create a helper function, we usually call them `containers`

```js
export const withUserContainer = (cb) => {
  return async (
     _parent: void,
     arg: ...,
     { req }: Context
   ): Promise... => {
         const { id } = arg
         const userId = req.user?.id
         if (!authorId) throw new Error('No User')
         return cb(_parent, arg, {req, userId} )
}
```

Then now you can use this container for every resolver you would wish to make sure userId exists:

```js

export const flagExercise = withUserContainer(async () => {} )

export const removeExerciseFlag = withUserContainer( async () => {} )
```

_Originally posted by @songz in https://github.com/garageScript/c0d3-app/pull/2012#discussion_r906688879_

HS-90:
**Title:** enhancement : withUserContainer helper

**Description:** closes #2017

This PR will create a helper method to use with any resolvers that checks for user id. For example: 
image

Instead of calling the above in every single resolver that needs it, we wrap it in the helper function instead. Although some exceptions do exist. For example `addExercises` needs the `authorId` in it's resolver function to update the prisma database, in which case we would have to create a variable for `authorId` and check for existence anyways like below.
image


flacial (approved):

SlyBouhafs (approved):

graphql/resolvers/exerciseCrud.ts
```diff
@@ -75,26 +73,24 @@ export const updateExercise = async (
...
   const exercise = await prisma.exercise.findUnique({
     where: {
       id
     }
   })
 
-  if (!isAdmin(req) && exercise?.authorId !== authorId) {
+  const authorId = req.user?.id
+
+  if (!isAdmin(req) && exercise?.authorId !== authorId)
```
songz, line 92: `withAdminContainer` and then we don't need the `isAdmin` check

helpers/withUserContainer.ts
```diff
@@ -0,0 +1,15 @@
+import { Context } from '../@types/helpers'
+
+export const withUserContainer =
```
anthonykhoa, line 3: lets move this container into the container folder:
https://github.com/garageScript/c0d3-app/tree/master/containers

anthonykhoa: Also, please make a followup PR to make another container to check if a user is an admin

helpers/withUserContainer.ts
```diff
@@ -0,0 +1,15 @@
...
+  <Type, ArgsType>(
+    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type
+  ) =>
+  async (_parent: void, args: ArgsType, ctx: Context) => {
+    const { req } = ctx
+
+    const authorId = req.user?.id
```
anthonykhoa, line 10: i believe the team decided to use lodash get instead of optional chaining for this repo

graphql/resolvers/exerciseCrud.ts
HS-90: @songz So in this example, I been getting stuck how to handle when you need to also check for the ` && exercise?.authorId !== authorId` on line 92 or something similar when the resolver needs to query something in prisma, for example lines 84-88:
```
 const exercise = await prisma.exercise.findUnique({
    where: {
      id
    }
  })
```
That part of the check can't be captured universally with the container, since it is unique to a specific resolver. 

The if statement on line 92 is also an `&&` operator. So checking the conditions individually with a separate if statement for each wont work.

anthonykhoa: IMO it's fine and would be a good choice to keep `if (exercise?.authorId !== authorId)` in this file since its specific to this resolver.

songz: oh this is my bad, deleting exercise seems like it can be used for both admins **and** regular users, so `withAdminContainer` wouldn't be appropriate here.

songz: it would be more appropriate like `updateLesson` or `deleteModule` etc, which you could just do in a separate PR.

songz (approved):

graphql/resolvers/exerciseCrud.ts
HS-90: @songz Hey no worries. Seems you need to either be an admin, or the regular user who created the exercise to be able to delete it, which makes sense. 

I applied some changes based on @anthonykhoa  comments and think it's good to merge now. Then I can make a separate PR for the AdminContainer

anthonykhoa (approved):

containers/withUserContainer.ts
```diff
@@ -0,0 +1,15 @@
...
+import _ from 'lodash'
+
+export const withUserContainer =
+  <Type, ArgsType>(
+    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type
+  ) =>
+  async (_parent: void, args: ArgsType, ctx: Context) => {
+    const { req } = ctx
+
+    const authorId = _.get(req, 'user.id', undefined)
```
anthonykhoa, line 11: undefined is unnecessary because lodash get will automatically return undefined when the looked up property(user.id) is not found in the object(req)

PR #2160, opened 2022-08-06
feat(components): Update Alert component styles

flacial:
Closes #2145 

# Problem

Bootstrap `alert-danger` class need the base class `alert` for the element to have a background color. What's currently happening, We're applying the color `alert-danger` without the base styles from `alert` which doesn't seem to be working.

image

# Solution

Add `alert` class to the `div` containing the alert component elements.
- From `alert-danger` to `alert alert-danger`.

## Other changes

- Align the items to the center
- Change `img` element to `next/image` because `next/image` has better performance and automatic image optimizations.
- Update info type styles (top is old style, bottom is new styles)
infoAlert
- Update urgent type styles
urgentAlert

PR #2196, opened 2022-08-16
feat: Create QueryInfo component

### LINKED ISSUE #2181: Add QueryStateMessage to our components library
Add the following component to our components library so it can be used to show the state of queries on other pages.

https://github.com/garageScript/c0d3-app/blob/3b01714049c8b8d76c0bbbe6cf97804356c460e4/components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx#L151-L189

flacial:
Closes #2181 
Related to #1975

# Problem

In the admin pages where we create data or the [mentor page](#1975), we want to display a message that describe the state of the request. Whether it's sent, loading, or encountered an error.  

This feature is already implemented in one of our components called [AdminLessonInputs](https://github.com/garageScript/c0d3-app/blob/15a241b8eada0016eea7cab21620184b1dc311b4/components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx#L151). 

# Solution

For this component to be used in other pages, this PR extract the code from the AdminLessonInputs component, makes it generic, and create a file for it.  The component go away after 4 seconds

### Screenshots

Basic:

image

Error:

image

Loading:

image

---

[Storybook](url) (waiting for storybook link)


components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx
```diff
@@ -148,49 +147,31 @@ const AdminModuleInputs = ({
...
+        }}
+        DataMessage={() => {
+          const updateModule = get(data, 'updateModule')
+          const addModule = get(data, 'addModule')
+
+          return (
+            <span>
+              {updateModule ? 'Updated' : 'Added'} the module{' '}
+              <strong>
+                {get(addModule, 'name') || get(updateModule, 'name') || ''}
```
anthonykhoa, line 168: The last OR is unnecessary
```
{get(addModule, 'name') || get(updateModule, 'name', ''}
```

components/QueryInfo/QueryInfo.tsx
```diff
@@ -0,0 +1,54 @@
...
+  loading: boolean
+  error: string
+  texts: {
+    loading: string
+    errorTitle: string
+  }
+  DataMessage?: () => JSX.Element
+}
+
+const QueryInfo = ({ data, loading, error, texts, DataMessage }: QueryInfo) => {
```
anthonykhoa, line 17: 1) Will these notification messages stay on the screen forever, or will they go away on their own?

2) Is it possible to close these notification messages that pop up? As in, is there an `X` button somewhere I can click to close the notification?

components/QueryInfo/QueryInfo.tsx
```diff
@@ -0,0 +1,54 @@
+import { CheckCircleIcon, AlertFillIcon } from '@primer/octicons-react'
+import React from 'react'
+import { Spinner } from 'react-bootstrap'
+import styles from '../../scss/queryInfo.module.scss'
+
+type QueryInfo = {
+  data: any
```
anthonykhoa, line 7: Can you remove this use of `any`?

anthonykhoa: I'm not super familiar with typescript, but from my knowledgeI think generics should help
https://www.typescriptlang.org/docs/handbook/2/generics.html

components/QueryInfo/QueryInfo.tsx
flacial: It's not possible to close them and they'll remain forever on the screen. I think it's better to hide them after some seconds.

components/QueryInfo/QueryInfo.tsx
flacial: > I'm not super familiar with typescripbut from my knowledge, I think generics should help https://www.typescriptlang.org/docs/handbook/2/generics.html

I wasn't sure how the generic type for a prop work, so I googled a bit and ended up with what's in the code. According to the code examples I've seen and based on my understanding, `T` stands for the prop type. I'm not certain though, there isn't a clear explanation of what `T` is.

components/QueryInfo/QueryInfo.tsx
anthonykhoa: Can we the message go away after a few seconds in this PR too?

songz (approved):

components/QueryInfo/QueryInfo.tsx
flacial: I think the only message that should go away after some seconds is the success one. The loading message go away after the data or error resolves. 

The error message is important and it should remain on screen until the admin switch to a different lesson/module/challenge to update/add (different issue in this [component](https://github.com/garageScript/c0d3-app/blob/7c98313c745a0997d49adda34102a79e0ee3d6ba/components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx#L74-L84) that's related to this).

anthonykhoa: Why is the error message notification important to keep on screen? We shouldn't rely on info from the notification message, that's only there to notify us that something went wrong, so I don't see a reason to keep it forever on the screen. A developer can gain more details by looking at the actual request(in network tab of devtools) that's not working as expected.

anthonykhoa: Also, what if we want to implement this for features that are public-facing? It would be bad experience for users to have a popup notification appear and stay on screen forever until they switch pages

components/QueryInfo/QueryInfo.tsx
```diff
@@ -0,0 +1,60 @@
...
+      </div>
+    )
+  }
+
+  if (error) {
+    return (
+      <div className={styles.error}>
+        <AlertFillIcon />
+        <span>
+          {texts.errorTitle}: {error}
```
anthonykhoa, line 38: This component seems like it has high potential to be used in a public-facing feature in the future. In that case, we shouldn't display the actual error coming from the request, that would be bad user experience for the user to see error logs that are for the developers to see

anthonykhoa:
I remember we have the Alerts component. It already has proper styling and even has the `x` button. We should be using the Alerts component for this use case, and make updates to it as needed. Instead of creating divs with custom css and using span elements, it would be better to use the Alerts component instead.

https://github.com/garageScript/c0d3-app/blob/abede305660455eddd75c1877893fee6853ddeae/components/Alert/Alert.tsx

Screen Shot 2022-08-18 at 11 48 50 AM

flacial:
> 

I'll look into it.

flacial:
> I remember we have the Alerts component. It already has proper styling and even has the `x` button. We should be using the Alerts component for this use case, and make updates to it as needed. Instead of creating divs with custom css and using span elements, it would be better to use the Alerts component instead.
> 
> https://github.com/garageScript/c0d3-app/blob/abede305660455eddd75c1877893fee6853ddeae/components/Alert/Alert.tsx
> 
> Screen Shot 2022-08-18 at 11 48 50 AM

To use it within this component or completely revert this and use the Alert instead?

flacial:
> Use it within this component(QueryInfo), and update Alerts component as needed. I'm thinking it will make more sense to implement the `hide after X amount of time` optional feature on the Alerts component instead of doing so inside QueryInfo component. If styling updates are needed it should be done in the Alerts component as well

This https://github.com/garageScript/c0d3-app/pull/2160 PR needs to be merged and another will be created to add the missing styles for the Alert component.

flacial:
@anthonykhoa I'm confused on how the hiding-after-x-seconds logic. With the current logic, it hides it first time and then never displays it because the `error` which is a string doesn't change.

Here's a scenario where it doesn't show again:

1. Submit with invalid inputs
2. Error message shows and disappear after 4 seconds
3. Submit again with invalid inputs
4. Error message doesn't appear again because the state didn't change (code below)

https://github.com/garageScript/c0d3-app/blob/85ff8c03f2d31adde0e84fb27c7ad5870e0eac66/components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx#L114

flacial:
I kept the `loading` state element. I feel `Alert` component is used for displaying a finished "info" about something. Whereas the loading element shows the progress of a request.

anthonykhoa:
> @anthonykhoa I'm confused on how the hiding-after-x-seconds logic work. With the current logic, it hides it first time and then never displays it because the `error` which is a string doesn't change.
> 
> Here's a scenario where it doesn't show again:
> 
> 1. Submit with invalid inputs
> 2. Error message shows and disappear after 4 seconds
> 3. Submit again with invalid inputs
> 4. Error message doesn't appear again because the state didn't change (code below)
> 
> https://github.com/garageScript/c0d3-app/blob/85ff8c03f2d31adde0e84fb27c7ad5870e0eac66/components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx#L114
> 
> Do you know a better approach?

This feature should not be specifically implemented for queryInfo component, it should be implemented for the Alerts component instead. That way, if anyone needs to show a notification message(using Alerts component) that disappears after X amount of time, then they can do so. It can be done in another PR. Originally I wanted it done in this PR because the user experience would be bad if a user cant close a popup notification message. But since we are now using Alerts component that comes with an exit button, we can do the `hide after x seconds` in a followup PR

flacial:
> > @anthonykhoa I'm confused on how the hiding-after-x-seconds logic work. With the current logic, it hides it first time and then never displays it because the `error` which is a string doesn't change.
> > Here's a scenario where it doesn't show again:
> > 
> > 1. Submit with invalid inputs
> > 2. Error message shows and disappear after 4 seconds
> > 3. Submit again with invalid inputs
> > 4. Error message doesn't appear again because the state didn't change (code below)
> > 
> > https://github.com/garageScript/c0d3-app/blob/85ff8c03f2d31adde0e84fb27c7ad5870e0eac66/components/admin/lessons/AdminLessonInputs/AdminLessonInputs.tsx#L114
> > 
> > Do you know a better approach?
> 
> This feature should not be specifically implemented for queryInfo component, it should be implemented for the Alerts component instead. That way, if anyone needs to show a notification message(using Alerts component) that disappears after X amount of time, then they can do so. It can be done in another PR. Originally I wanted it done in this PR because the user experience would be bad if a user cant close a popup notification message. But since we are now using Alerts component that comes with an exit button, we can do the `hide after x seconds` in a followup PR

Alright. I'll implement the exit button feature then merge this PR and create another one for the `hide after x seconds`.

flacial:
Before proceeding, we must fix the [issue with dismissing an alert](https://github.com/garageScript/c0d3-app/issues/2215).

SlyBouhafs (approved):

PR #2199, opened 2022-08-17
feat(dojo/admin): Create Introduction page

### LINKED ISSUE #2173: DOJO(admin): Create Introduction tab
Create this [page](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1394%3A7694)

You'll use the [FormCard](https://github.com/garageScript/c0d3-app/blob/master/components/FormCard.tsx) component with a bit of modifications to make it work for the layout structure. When `SAVE CHANGES` is hit, it should call the [updateLesson](https://github.com/garageScript/c0d3-app/blob/5a99b96a5bac6fe0a485d4435f07f2ba522a573f/graphql/resolvers/lessonsController.ts#L23) resolver (its hook could be found within GraphQL index file).

Look into the `docFileUrl` inputs. I don't think they're supposed to be this many.

flacial:
Closes #2173 

Merging this PR will create the page where the admin can update the lesson's content such as the title or the description. It's similar to [this page](https://www.c0d3.com/admin/lessons) but redesigned

Page image:

image

The page is under `/admin/lessons/[lessonSlug]/introduction`. [Try it](https://c0d3-app-git-fork-flacial-2173-dojoadmin-creat-5c72b1-c0d3-prod.vercel.app/admin/lessons/js1/introduction)


pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
```diff
@@ -39,26 +53,76 @@ type Module = {
...
-  pageName?: string | string[]
+const IntroductionPage = ({ lesson }: { lesson: Lesson }) => {
+  const [updateLesson] = useUpdateLessonMutation()
+
+  const [formOptions, setFormOptions] = useState(
+    getPropertyArr(lesson, ['challenges', '__typename'])
+  )
+
+  // Update the Inputs values when the lesson changes
+  useEffect(
```
Comment (songz) [inline pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: why do you need `useEffect` here? Would this render when lesson change automatically?

SlyBouhafs (approved):

pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: It's possible to switch between lessons from the Breadcrumbs lessons dropdown menu. When the admin switches to a different lesson, the input value remain the same. To solve it, this `useEffect` has been added.

anthonykhoa:
please include image of the new page.

pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
Comment (anthonykhoa) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: I think there is something wrong with the implementation of logic. We should not be having to use useEffect for this use case. A proper re-render when props change should solve the issue

Can you try giving both child and parent components the same key value?

Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: > Can you try giving both child and parent components the same key value?

What do you mean?

Comment (anthonykhoa) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: A few possibilities I was considering:

1) I was thinking that since React sometimes relies on keys to decide when to rerender, then possible placing keys could solve the issue, so that is one possible fix. 
Try giving the IntroductionPage component a key value that changes along with lesson value:
```
return <IntroductionPage lesson={lesson} key={lesson.id} />
```

2) Another possibility for the cause of this issue is that maybe there is no state/prop change. What does the breadcrumb component do? Does it change any states? How does it achieve the effect of switching lessons? I see on line 227 we do `router.push`, so does that mean the breadcrumb component achieves the switching of lessons by redirection to a URL for a specific lesson?
https://github.com/garageScript/c0d3-app/pull/2199/files#diff-a754b3c648a6d2067946a27fc92b81619719063cde89810678ba5d9ce444231aR227

Does this mean that

3) Maybe the state change is still referencing the same object(thus same memory address, so react doesn't see it an update in value).

Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: The component re-renders when the lesson changes. When it re-renders, the initial state `formOptions` value `getPropertyArr(lesson, ['challenges', '__typename'])` doesn't get set again because the component got re-rendered. 

It only does when the component first mounts (as far as I tested). To workaround this, one solution is to set the state value again through `useEffect`.

If you tried to comment the `useEffect` code, went to the introduction tab, switched to another lesson, went to the modules tab, and finally back to the introduction tab, you'll notice the inputs have the data of the new lesson. 

This is because it got dismounted. The state initial value is now the new lesson.

Comment (anthonykhoa) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: I see, thanks for additional context. My first suggestion  to give IntroductionPage a key prop should work to solve this issue then.

Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: > I see, thanks for additional context. My first suggestion to pass IntroductionPage a key prop should work to solve this issue then.

What will this solve? We're passing it so the component re-renders? 

I applied the suggested changes and it still didn't update the state `formOptions` with the new lesson data.

Comment (anthonykhoa) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: Hmm looks like the issue is not to do with passing a key then. And yeah it would have been better. Key is a special react prop used by react to detect when a component has changed. It isn't supposed to be used for any other purpose

Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:64]: Actually, `key={lesson.id}` works. I didn't know the `key` prop could be used to reset the component's states. My first attempt was `lessonId={lesson.id}` which didn't work. But I tried now with the key prop and it worked. Super nice 😄

pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
```diff
@@ -39,26 +53,76 @@ type Module = {
...
-type ContentProps = {
-  pageName?: string | string[]
+const IntroductionPage = ({ lesson }: { lesson: Lesson }) => {
+  const [updateLesson] = useUpdateLessonMutation()
+
+  const [formOptions, setFormOptions] = useState(
+    getPropertyArr(lesson, ['challenges', '__typename'])
+  )
+
+  // Update the Inputs values when the lesson changes
```
Comment (anthonykhoa) [inline pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:63]: I think you forgot to remove these commented out lines

Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:63]: Oops, yeah.

bryanjenningz (approved):

bryanjenningz (approved):

flacial:
QueryInfo component is missing in this page

PR #2216, opened 2022-08-24
fix: Alert can't be dismissed 

### LINKED ISSUE #2215: fix: Can't dismiss an alert
# Problem

The [Alert component](https://github.com/garageScript/c0d3-app/blob/master/components/Alert/Alert.tsx) has a close button but it's not working. This is because of how the bootstrap classes are used. For the proper use, [refer to the bootstrap doc](https://getbootstrap.com/docs/4.1/components/alerts/#dismissing).

# Solution

We'll update the component to use the [Alert component](https://react-bootstrap.github.io/components/alerts/) from react-bootstrap library. This component makes it possible to dismiss an alert by passing a boolean.

---

Example code:

```tsx
<Alert
  variant={`${type === 'urgent' ? 'danger' : 'primary'}`}
  onClose={() => {
    setShow(false)
    onDismiss && onDismiss(id)
  }}
  dismissible={!!onDismiss}
>
  // ...
</Alert>
```

flacial:
Closes #2215 

# Problem 

The [Alert component](https://github.com/garageScript/c0d3-app/blob/master/components/Alert/Alert.tsx) can't be dismissed through clicking the close button.

# Solution

Replace its implementation with [react-bootstrap alert component](https://react-bootstrap.github.io/components/alerts/#dismissing).

Screen Recording


SlyBouhafs (approved):

components/Alert/Alert.tsx
```diff
@@ -21,16 +22,20 @@ const alertIconMap: { [type: string]: string } = {
...
-  return (
-    <div
-      className={`${styles.alert} d-flex justify-content-between mt-3 ${alertClasses}`}
-      role="alert"
+  return show ? (
+    <AlertBS
+      variant={`${type === 'urgent' ? 'danger' : 'primary'}`}
+      onClose={() => {
+        setShow(false)
+        onDismiss && onDismiss(id)
```
HS-90, line 36: Just out of curiosity and to confirm, does `onDismiss` handle whether or not the alert will be displayed for particular user in the future(if the user clicks the x button to dismiss)?

HS-90 (approved):

songz (approved):

components/Alert/Alert.tsx
flacial: No, it doesn't handle it. It gets called when the alert is closed. It's useful when we want to perform a certain action when the alert closes.

PR #2219, opened 2022-08-25
withAdminContainer function

### LINKED ISSUE #2017: Create withUserContainer helper method

HS-90:
**Title:** enhancement : withAdminContainer function

**Description:** closes #2017

This PR will create a container function that checks for existence of an admin user for certain resolvers.
It is the follow-up to this withUserContainer PR: #2166 

The container file consists of two helper functions: `withAdminContainer` and `withAdminUserContainer`. 

`withAdminContainer` checks if the user is an admin, and if not, will throw an error. Example use case below:
image

`withAdminUserContainer` checks both if the user exists, and whether or not the user is an admin:
image


containers/withAdminContainer.ts
```diff
@@ -0,0 +1,45 @@
...
+export const withAdminUserContainer =
+  <Type, ArgsType>(
+    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type,
+    errorMessage: string | undefined = undefined
+  ) =>
+  async (_parent: void, args: ArgsType, ctx: Context) => {
+    const { req } = ctx
+
+    //check if user exists
+    const authorId = _.get(req, 'user.id')
```
anthonykhoa, line 34: Is it possible to use withUserContainer here so we have less duplicate code?

containers/withAdminContainer.ts
```diff
@@ -0,0 +1,45 @@
+import { Context } from '../@types/helpers'
+import { isAdmin, isAdminOrThrow } from '../helpers/isAdmin'
+import _ from 'lodash'
+
+//use when only checking if user is admin
+export const withAdminContainer =
+  <Type, ArgsType>(
+    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type,
+    errorMessage: string | undefined = undefined
```
flacial, line 9: Can't this be optional with `errorMessage?: string`

containers/withAdminContainer.ts
```diff
@@ -0,0 +1,45 @@
...
+  ) =>
+  async (_parent: void, args: ArgsType, ctx: Context) => {
+    const { req } = ctx
+
+    //check if user exists
+    const authorId = _.get(req, 'user.id')
+    if (!authorId) throw new Error('No user')
+
+    //check if is admin
+    if (errorMessage) {
```
flacial, line 38: Thought: I think this to `44` could be replaced with a single call to `withAdminContainer`:

```ts
export const withAdminUserContainer =
  <Type, ArgsType>(
    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type,
    errorMessage: string | undefined = undefined
  ) =>
  async (_parent: void, args: ArgsType, ctx: Context) => {
    const { req } = ctx

    //check if user exists
    const authorId = _.get(req, 'user.id')
    if (!authorId) throw new Error('No user')

    //check if is admin
    return withAdminContainer(resolver)(_parent, args, ctx)()
  }
```

containers/withAdminContainer.ts
anthonykhoa: One idea i have to is wrap this entire container with withUserContainer

```
export const withAdminUserContainer = withUserContainer(...)
```

flacial (approved):

containers/withAdminContainer.ts
anthonykhoa: Hmm looks like this container also has duplicate code with the `withAdminContainer` too -- is it possible to use withAdminContainer here too?

anthonykhoa: Maybe something like this could work
```
withAdminUserContainer = (resolver, errorMessage) => {
  return withUserContainer(WithAdminContainer(resolver, errorMessage))
}
```

containers/withAdminContainer.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    return resolver(_parent, args, ctx)
+  }
+
+export const withAdminUserContainer =
+  <Type, ArgsType>(
+    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type,
+    errorMessage?: string
+  ) =>
+  async (_parent: void, args: ArgsType, ctx: Context) => {
```
anthonykhoa, line 30: Can you try

```
export const withAdminUserContainer =
  <Type, ArgsType>(
    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type,
    errorMessage?: string
  ) => withUserContainer(withAdminContainer(resolver, errorMessage))
```

containers/withAdminContainer.ts
```diff
@@ -0,0 +1,34 @@
...
+export const withAdminContainer =
+  <Type, ArgsType>(
+    resolver: (_parent: void, args: ArgsType, ctx: Context) => Type,
+    errorMessage?: string
+  ) =>
+  async (_parent: void, args: ArgsType, ctx: Context) => {
+    const { req } = ctx
+
+    //check if is admin
+    if (errorMessage) {
```
anthonykhoa, line 16: Lets simplify this to:

```
if (!isAdmin(req)) {
  throw new Error(errorMessage || 'User is not an admin')
}
```

After updating all resolvers that check for `admin` with the withAdminContainer and withAdminUserContainer, we can remove the isAdminOrThrow function since the containers in this file basically replace it

HS-90: @anthonykhoa I don't think we can remove it entirely because some checks that are unique to a specific resolver can't be replicated inside a container. for example the below from updateExercise:
```
 const authorId = req.user?.id

  if (!isAdmin(req) && exercise?.authorId !== authorId)
    throw new Error('Not authorized to change')

```

HS-90: The above code is checking if this person is either an admin or the original author/creator of the exercise(in this case, they dont have to be an admin). So unless there is a way we can capture the second part of the check `exercise?.authorId !== authorId` in a container, without making it too complicated, i think we should keep as is.

if we remove the `isAdmin` here, apply the `withAdminContainer` and changed it to just:  

`if (exercise?.authorId !== authorId)
    throw new Error('Not authorized to change')`

the container will throw an error and not allow the original author(if they are not an admin) to change the exercise.

anthonykhoa: > i think we should keep as is.

Can you explain the benefit of keeping the isAdmin function call check in this resolver? I don't understand why we are doing this check inside the resolver, since we made the withAdminContainer container so that we don't to do checks for admin status inside the resolvers anymore.

anthonykhoa: > So unless there is a way we can capture the second part of the check exercise?.authorId !== authorId in a container,

Why not just keep this check in the updateExercise resolver itself?

HS-90: When I was working on it, and tried to apply the `withAdminContainer` to the `updateExercise` resolver, I noticed tests were failing. Below compares the behavior we ideally want, vs what happens when we apply the container, and do the check (exercise?.authorId !== authorId) by itself while wrapped in a  `withAdminContainer`:

image

The way the resolver is set up, it should throw an error only if the user isnt an admin `&&` if the user isnt the author of the exercise. With the container, it will throw an error if the user isnt admin `||` if the user isnt the author.

anthonykhoa: So is it just because tests are failing? When we use withAdminContainer and keep `exercise?.authorId !== authorId` inside the updateExercise resolver, it still ensures that the user must be both an admin and also the author of the exercise in order to make an update to an exercise in updateExercise resolver, right?

anthonykhoa: Oh , i see what you are saying now

anthonykhoa: Lets continue using the isAdmin function in this container for now then. Just do the refactoring necessary so that you don't have the `isAdmin` and `throw new Error` twice in this resolver(lines 17 and 20).

containers/withAdminContainer.ts
HS-90: @anthonykhoa this didn't work as we need to pass in the` (_parent, args, ctx)` when calling the withUserContainer function. Since `withUserContainer` returns another function.
Like this: `withUserContainer(withAdminContainer(resolver, errorMessage))(_parent, args, ctx)`

anthonykhoa: Just to verify, you tried copying/pasting my example code and it didn't work?

HS-90: So the tests work:

but I keep getting this type error:
image

then I tried: 
image

getting the same error.

also tried:
image

anthonykhoa: try passing just resolver and errorMessage then

anthonykhoa: doing `...arguments as const` should work too

HS-90: passing in resolver and errorMessage worked, but `...arguments as const` gave the same error

anthonykhoa (approved):

containers/withAdminContainer.ts
```diff
@@ -0,0 +1,27 @@
+import { Context } from '../@types/helpers'
+import { isAdmin } from '../helpers/isAdmin'
+import { withUserContainer } from './withUserContainer'
+import _ from 'lodash'
```
anthonykhoa, line 4: Ah just noticed this -- i think we arent using anything from lodash in this file anymore, so it can be cleaned up

HS-90: ill make a new PR that fixes this.

PR #2247, opened 2022-09-01
Remove unnecessary forgot-password full page load

bryanjenningz:
Closes https://github.com/garageScript/c0d3-app/issues/2250

Before this change, when you got an error on the forgot-password page then clicked "Go Back", it would do a full page reload.
This change removes that full page load, so now it just goes back instantly when you click "Go Back".

How to test:
* Go to /forgotpassword.
* Type in a username that doesn't exist and hit ENTER so it causes an error to happen.
* Click the "Go Back" button and notice there is no full page reload.

### DIFF
--- a/__tests__/pages/__snapshots__/forgotpassword.test.js.snap
+++ b/__tests__/pages/__snapshots__/forgotpassword.test.js.snap
@@ -182,13 +182,12 @@ exports[`ForgotPassword Page Should render invalid user/email form on error 1`] 
           <p
             class="card-text"
           />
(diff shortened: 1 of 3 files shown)


flacial (approved):

HS-90 (approved):

flacial:
May you create and link an issue to it (`Closes #1234`)?

bryanjenningz:
> May you create and link an issue to it (`Closes #1234`)?

Done

PR #2248, opened 2022-09-02
Remove full page load link from confirm/[token] page

bryanjenningz:
Closes https://github.com/garageScript/c0d3-app/issues/2249

Before this change, there was a link that did a full page load on the confirm password change page (e.g. /confirm/123).
This change removes that full page load link and instead uses the Next.js Link component which improves performance by not doing a full page load (https://nextjs.org/docs/messages/no-html-link-for-pages).

How to test:
* Go to /confirm/123.
* Type in a new password twice and hit ENTER.
* Click the dashboard link and confirm that a full page load doesn't happen anymore and the page loads faster.

### DIFF
--- a/pages/confirm/[token].tsx
+++ b/pages/confirm/[token].tsx
@@ -2,6 +2,7 @@ import { useMutation } from '@apollo/client'
 import React from 'react'
 import { useRouter } from 'next/router'
 import { Formik, Form, Field } from 'formik'
+import Link from 'next/link'
 import UPDATE_PASSWORD from '../../graphql/queries/updatePassword'
 import NavLink from '../../components/NavLink'
 import Input from '../../components/Input'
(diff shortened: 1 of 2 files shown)


pages/confirm/[token].tsx
```diff
@@ -16,9 +17,9 @@ const initialValues = {
 
 const ConfirmSuccess: React.FC = () => (
   <Card type="success" title="Password has been set!">
-    <a className="btn btn-primary btn-lg mb-3" role="button" href="/curriculum">
```
Comment (bryanjenningz) [inline pages/confirm/[token].tsx:19]: It seems like screen readers should view this as a link rather than a button, so I removed `role="button"`.

flacial (approved):

flacial:
May you create and link an issue to it (`Closes #1234`)?

bryanjenningz:
> May you create and link an issue to it (`Closes #1234`)?

Done

songz (approved):

SlyBouhafs (approved):

Issue #2253, opened 2022-09-02
DOJO(main/page): Answering questions page


flacial:
[Create the main part](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1405%3A7764) of DOJO which is a page that the student will answer the questions through.

evo777:
This is a wonderful idea!

PR #2265, opened 2022-09-07
Add DOJO exercises nav card

bryanjenningz:
This pull request adds a simple nav card to the exercises page.

exercises-nav

How to test:
* Go to /exercises/js0
* Check that the nav card is there
* Check that the nav card links work


pages/exercises/[lessonSlug].tsx
```diff
@@ -25,12 +26,48 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
...
+        className="py-2 px-3 mx-1 fw-bold text-primary"
+      >
+        CHALLENGES
+      </NavLink>
+      <div className="py-2 px-3 mx-1 fw-bold bg-primary text-white rounded">
+        EXERCISES
+      </div>
+    </div>
+  )
+}
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:69]: Right now, the nav card is really simple and it's only being used in this file so I decided to keep it in this file instead of making it a separate component. It looks similar to the AppNav component (https://github.com/garageScript/c0d3-app/blob/abede305660455eddd75c1877893fee6853ddeae/components/AppNav/AppNav.tsx), so maybe we should move it into the components/AppNav folder and share some styling.

pages/exercises/[lessonSlug].tsx
```diff
@@ -25,12 +26,45 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
...
     <Layout title={currentLesson.title}>
+      <ExercisesNavCard lessonDocUrl={currentLesson.docUrl} lessonSlug={slug} />
       <h1>{currentLesson.title}</h1>
       {alerts && <AlertsDisplay alerts={alerts} />}
     </Layout>
   )
 }
 
+type ExercisesNavCardProps = {
+  lessonDocUrl: string | null | undefined
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:37]: The reason why the lessonDocUrl property has type `string | null | undefined` is because the `Lesson` type has its docUrl property as `docUrl?: string | null`. I made lessonDocUrl `string | null | undefined` instead of `lessonDocUrl?: string | null` so that people calling `ExerciseNavCard` are forced to pass in a `lessonDocUrl` prop.

pages/exercises/[lessonSlug].tsx
```diff
@@ -25,12 +26,45 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
...
+      )}
+      <NavLink
+        path={`/curriculum/${lessonSlug}`}
+        className="py-2 px-3 mx-1 fw-bold text-primary"
+      >
+        CHALLENGES
+      </NavLink>
+      <div className="py-2 px-3 mx-1 fw-bold bg-primary text-white rounded">
+        EXERCISES
+      </div>
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:63]: I'm not sure if we want to make this also a NavLink that points to the current page. I made it a div instead since it doesn't seem like there's a point to having a link to the current page, but maybe it makes sense to add a link to the current page. Let me know if you think I should add a link to the current page.

pages/exercises/[lessonSlug].tsx
Comment (flacial) [reply pages/exercises/[lessonSlug].tsx:66]: Can we make [AdminLessonNav](https://github.com/garageScript/c0d3-app/blob/master/components/admin/lessons/AdminLessonSideNavLayout.tsx) generic and use it here?

pages/exercises/[lessonSlug].tsx
Comment (flacial) [reply pages/exercises/[lessonSlug].tsx:63]: Yeah, I agree. It's unnecessary to point to the same page.

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:66]: That's a good idea. I'll make a separate issue and pull request which will make AdminLessonNav generic and easier to use. Once that pull request is merged, I'll use that generic component in this pull request.

Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:66]: Here's the issue that was created: https://github.com/garageScript/c0d3-app/issues/2268

pages/exercises/[lessonSlug].tsx
```diff
@@ -23,9 +24,21 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
   if (!currentLesson)
     return <Error code={StatusCode.NOT_FOUND} message="Lesson not found" />
 
+  const tabs = [
+    ...(currentLesson.docUrl
+      ? [{ text: 'lessons', url: currentLesson.docUrl }]
+      : []),
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:30]: This code doesn't look very good. This makes it so if the current lesson doesn't have a docUrl, it will not show a LESSONS NavCard link. I wasn't sure if there was a better way to do this or if anyone has any suggestions, feel free to comment ideas.

flacial (approved):

songz (approved):

PR #2272, opened 2022-09-08
Make AdminLessonNav more generic

### LINKED ISSUE #2268: Make AdminLessonNav generic so it can be reused
[AdminLessonNav](https://github.com/garageScript/c0d3-app/blob/master/components/admin/lessons/AdminLessonSideNavLayout.tsx) is currently only used in the admin page, but we want to have the same style of navigation tabs in the [DOJO exercises page](https://github.com/garageScript/c0d3-app/pull/2265), so it makes sense to make AdminLessonNav more generic and use it in both places since both places have the same requirements and behavior and we want them to be consistent with each other.

We also want the API to require less boilerplate to use. Currently, we need to pass in [our own component](https://github.com/garageScript/c0d3-app/blob/7af5b279be26cc4993ab4b274f4e45329dc3fda7/pages/admin/lessons/%5BlessonSlug%5D/%5BpageName%5D/index.tsx#L158) which requires [some boilerplate](https://github.com/garageScript/c0d3-app/blob/7af5b279be26cc4993ab4b274f4e45329dc3fda7/pages/admin/lessons/%5BlessonSlug%5D/%5BpageName%5D/index.tsx#L117-L135).

Since the only parts of the nav tabs that change are the nav tab text and the url and whether a tab is selected, we can change the AdminLessonNav props so it just takes in the props `{ tabs: { text: string, url: string, isSelected: boolean }[] }` to represent each tab. We can expect the styling of a selected and unselected tab to be consistent for each use case, so these simplified props cover all our use cases and it requires less boilerplate than the current props which require you to pass in a component.

Problems with this approach:
* It's possible for multiple tabs to be selected, which seems like an invalid state, we could solve this by instead having the props be `{ selectedTabIndex: number | null, tabs: { text: string, url: string }[] }`.
* We could potentially remove the isSelected and selectedTabIndex and just have the AdminLessonNav determine if a tab is selected by checking the current URL with the URL of each tab. This would make it so we only need to pass in props `{ tabs: { text: string, url: string }[] }`, which is more simple, but it makes it so the tab is only selected if the current URL exactly matches the tab.
* We don't have a high degree of customization of the styling, we could solve this by adding an optional component that we can pass in which determines the look of each tab, similar to the current props but this time it's optional.

bryanjenningz:
Closes https://github.com/garageScript/c0d3-app/issues/2268

This pull request changes:
* AdminLessonNav is now more generic and called NavCard
* NavCard takes in more simplified props
* Makes it easier for the DOJO exercises page to use this NavCard component (https://github.com/garageScript/c0d3-app/pull/2265)

How to test:
* Go to /admin/lessons/js1/modules
* Verify that NavCard works and looks exactly how it looked and worked before

navcard


pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
```diff
@@ -148,14 +125,14 @@ const Lessons = ({ data }: GetAppProps) => {
...
         <section>
-          <AdminLessonNav
+          <NavCard
             tabs={[
               {
-                tabName: 'modules',
-                urlPageName: 'modules'
+                text: 'modules',
+                url: `${MAIN_PATH}/${lesson.slug}/modules`,
+                isSelected: pageName === 'modules'
```
Comment (bryanjenningz) [inline pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:133]: There's much less boilerplate now, so users of this generic component no longer need to provide their own nav item component with their own styling.

An alternative way to do this would be to make it so we determine whether a tab is selected by looking at the current URL, but this would have the problem of not being customizable enough for some potential use cases where we might want the tab to be selected even if the URL is not exactly the same as the tab's URL.

scss/navCard.module.scss
```diff
@@ -55,12 +55,8 @@ $light_primary: color.adjust(variables.$primary, $lightness: 28%);
...
 }
 
-.lessons_tabsNav__nav__item--inactive:hover {
+.navCard__tabsNav__nav__item--inactive:hover {
   background-color: color.change(variables.$primary, $lightness: 95%);
   color: $dark_primary;
   transition: background-color 0.3s ease-out;
 }
-
-.lessons_tabs {
```
bryanjenningz, line 64: This .lessons_tabs CSS class wasn't being used (https://github.com/garageScript/c0d3-app/search?q=lessons_tabs) so I deleted it.

pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx
Comment (flacial) [reply pages/admin/lessons/[lessonSlug]/[pageName]/index.tsx:133]: `tabSelected` is great. Make things way simpler.

flacial (approved):

HS-90 (approved):

PR #2273, opened 2022-09-08
fix: Submission review count shows with none submissions

### LINKED ISSUE #2208: Lesson submissions-to-review badge 0 submissions
By @Ulisseus 

# Problem

> It looks like there was a redesign at some point (new font looks great) but submission count labels are incorrect REVIEW 1 SUBMISSIONS instead of REVIEW 1 SUBMISSION. And REVIEW 0 SUBMISSIONS probably shouldn't be visible in the first place. 

image

# Solution

Hide the badge when there are 0 submissions. Pluralize the word "SUBMISSION" if the submitted challenges are more than 1.

flacial:
Closes #2208 

Merging this PR will stop the submissions review count from being displayed if there's no submissions.

Before:

image

After:

image

To test it:
1. Go to `/curriculum`
2. Notice how it doesn't show for some of the lessons


components/LessonCard.tsx
```diff
@@ -27,50 +27,42 @@ type ReviewButtonProps = {
...
-    return <span>0</span>
+  if (!props.isCompleted) {
+    return null
   }
-  const pendingSubmissionsCount = data.submissions.reduce(
-    (acc: number, val: any) => {
+
+  const pendingSubmissionsCount =
+    data &&
+    data.submissions.reduce((acc: number, val: any) => {
```
bryanjenningz, line 47: Why is val's type any? Is there way to give val a more safe type?

components/LessonCard.tsx
```diff
@@ -27,50 +27,42 @@ type ReviewButtonProps = {
...
     variables: { lessonId: props.lessonId },
     fetchPolicy: 'network-only',
     nextFetchPolicy: 'cache-first'
   })
 
-  if (loading) {
-    return (
-      <div className="spinner-border spinner-border-sm mx-1" role="status" />
-    )
-  }
```
bryanjenningz, line 45: Why is the loading spinner being removed?

components/LessonCard.test.js
```diff
@@ -47,10 +52,25 @@ describe('Lesson Card Complete State', () => {
...
+      data: {
+        submissions: []
+      }
+    })
+
+    render(<LessonCard {...props} currentState="completed" />)
+
+    expect(
+      screen.queryByTestId('review-submissions-count')
+    ).not.toBeInTheDocument()
```
bryanjenningz, line 73: What happens if there are no elements with data-testid="review-submissions-count"? Does screen.queryByTestId('review-submissions-count') return undefined and then it checks if undefined is not in the document like this: `expect(undefined).not.toBeInTheDocument()`?

__tests__/pages/__snapshots__/curriculum.test.js.snap
```diff
@@ -2065,13 +2065,12 @@ exports[`Curriculum Page Should render with lessonStatus data 1`] = `
...
+              class="btn btn-sm btn-primary text-white float-end mb-2 me-2 me-0"
+              data-testid="review-submissions-count"
               href="/review/js0"
             >
               Review 
-              <span>
-                0
-              </span>
+              <span />
                Submissions
```
bryanjenningz, line 2074: Is this supposed to not show "Review Submissions" if there are 0 submissions that need to be reviewed?

components/LessonCard.tsx
flacial: I'll try that!

components/LessonCard.test.js
flacial: Yes exactly! If there no element found, `queryByTestId ` will return `undefined`. I could've done `expect(getByTestId(...))` but I think `queryByTestId` + `.not.toBeInDocument` is more descriptive.

__tests__/pages/__snapshots__/curriculum.test.js.snap
flacial: No, something is wrong. I wonder if it's the conditions.

bryanjenningz (approved):

components/LessonCard.tsx
```diff
@@ -27,50 +27,42 @@ type ReviewButtonProps = {
...
-  if (!data) {
-    return <span>0</span>
+  if (!props.isCompleted) {
+    return null
   }
-  const pendingSubmissionsCount = data.submissions.reduce(
-    (acc: number, val: any) => {
+
+  const pendingSubmissionsCount =
+    data &&
```
anthonykhoa, line 49: we use lodash `get` to safety get properties from objects, we should the same here.

HS-90 (approved):

components/LessonCard.tsx
```diff
@@ -45,14 +46,14 @@ const ReviewButton: React.FC<ReviewButtonProps> = props => {
...
 
-  const pendingSubmissionsCount =
-    data &&
-    data.submissions.reduce((acc, val) => {
-      if (val.status === SubmissionStatus.Open) {
-        return acc + 1
-      }
-      return acc
-    }, 0)
+  const submissions = get(data, 'submissions') || []
```
anthonykhoa, line 49: We dont have to do `|| []` ourselves, lodash provides third parameter for that use case

const submissions = get(data, 'submissions', [])

PR #2278, opened 2022-09-09
Edit Comments Frontend Functionality

HS-90:
**Title:** enhancement : adds ability to edit comments on a submission
**Description:** closes https://github.com/garageScript/c0d3-app/issues/893

This PR will allow users to edit their own comments on submissions from the frontend. Clicking the edit button will allow users to make changes to existing comments, as well as discarding changes, which will revert the comment to it's original state pr previously saved edit.

**To test:** login as a Dummy 'Admin' user, try to create then edit existing comments. You can then log out and login as another user(such as 'Leetcoder'), You should now see that you can no longer edit or delete comments that were created by "Admin".

Note: `querySelector` classes were used to select button elements in the test file, as our custom Button components do not accept `test-id` as a prorp


components/SubmissionComments.tsx
```diff
@@ -1,19 +1,24 @@
...
+import { MdInput } from './MdInput'
+import _ from 'lodash'
 
 export const SubmissionComments: React.FC<{
   comments: Comment[]
   submission: Submission
 }> = ({ comments, submission }) => {
   const [deleteComment] = useDeleteCommentMutation()
+  const [editComment] = useEditCommentMutation()
+  const [editing, setEditing] = useState<undefined | number>()
```
flacial, line 20: I think it's better to use `null | number` because undefined means this value is not assigned and we don't know it.

`undefined`: unintentional missing value
`null`: intentional missing value

components/SubmissionComments.tsx
```diff
@@ -1,19 +1,24 @@
...
+  const [curInput, setCurInput] = useState('')
```
flacial, line 21: NIT

`current` is usually referred to as `curr` I think.

components/SubmissionComments.tsx
```diff
@@ -34,6 +39,46 @@ export const SubmissionComments: React.FC<{
     })
   }
 
+  const handleStartEditing = (index: number, initialInput: string) => {
+    if (editing !== undefined)
+      alert(
```
flacial, line 44: This should be a modal to make the experience more pleasant. Look into react-bootstrap modal

components/SubmissionComments.tsx
```diff
@@ -34,6 +39,46 @@ export const SubmissionComments: React.FC<{
...
+    setCurInput('')
+  }
+
+  const handleDiscardChange = () => {
+    setEditing(undefined)
+    setCurInput('')
+  }
+
+  //to make sure comments load in correct order
+  comments = _.orderBy(comments, ['id'], ['asc'])
```
flacial, line 80: I'm not a fan of modifying the parameters. Perhaps I'm biased. But isn't there away to not modify the parameters? Like storing a local copy of the comments and modify it.

flacial (commented):
Really cool feature!

components/SubmissionComments.test.js
```diff
@@ -51,10 +68,42 @@ const comments = [
...
+    line: 2,
+    fileName: 'js7/1.js',
+    author: {
+      username: 'admin'
+    }
   }
 ]
 
 describe('Test SubmissionComments Component', () => {
+  const fillOutLoginForm = async (getByTestId, str) => {
```
flacial, line 87: This should be renamed to `fillOutCommentInput`

components/SubmissionComments.test.js
```diff
@@ -105,4 +139,82 @@ describe('Test SubmissionComments Component', () => {
...
+        <Wrapper>
+          <SubmissionComments
+            comments={comments.slice(0, 1)}
+            submission={submissionData}
+          />
+        </Wrapper>
+      </MockedProvider>
+    )
+
+    const editButton = container.querySelector('.btn-info.btn-sm')
```
flacial, line 202: In the future, we should update the `Button` component to take a `data-testid` prop so we can use `getByTestId('edit-button')`.

components/SubmissionComments.test.js
```diff
@@ -105,4 +139,82 @@ describe('Test SubmissionComments Component', () => {
...
+    await fillOutLoginForm(getByTestId, 'edited test comment') // previous text: test comment
+
+    const saveButton = container.querySelector('.btn-info')
+    await userEvent.click(saveButton)
+
+    const editCommentMutation = mocks[2].newData
+
+    expect(editCommentMutation).toHaveBeenCalled()
+
+    expect(deleteButton).toBeTruthy()
```
flacial, line 217: `expect(...).toBeInDocument()` is more descriptive.

flacial: To use it, you must import `import '@testing-library/jest-dom'`

components/__snapshots__/DiffView.test.js.snap
```diff
@@ -3437,6 +3443,12 @@ exports[`DiffView component Should render diff with comments 1`] = `
                         </a>
                       </div>
                       <div>
+                        <button
+                          class="btn borderless btn-info btn-sm"
+                          style="color: rgb(255, 255, 255);"
```
bryanjenningz, line 3448: I think you could use className="text-white" instead
https://getbootstrap.com/docs/5.0/utilities/colors/

components/SubmissionComments.tsx
```diff
@@ -53,16 +106,54 @@ export const SubmissionComments: React.FC<{
...
+                    data-testid="edit-button"
+                  >
+                    Edit
+                  </Button>
+                  <Button
+                    outline
+                    type="danger"
+                    color="danger"
+                    onClick={() => handleDeleteComment(c)}
+                    data-testid="delete-button"
```
bryanjenningz, line 151: Maybe add a title="Delete comment" for better accessibility

components/SubmissionComments.tsx
HS-90: perhaps a deepcopy of the comments?

flacial: I don't think the `comments` is modified. Just a `const submissionsComments = comments` will do the job.

components/__snapshots__/DiffView.test.js.snap
HS-90: I think this is assigned by the  custom Button component when you assign values to certain properties.
I dont think the component allows you to pass in a `className` prop.

components/SubmissionComments.tsx
HS-90: same as above, I dont think our Button component accepts a title prop

components/SubmissionComments.test.js
HS-90: tests fail when i changed them to `.toBeIntheDocument()` here for some reason. They passed in other tests though.

components/SubmissionComments.tsx
```diff
@@ -53,16 +128,54 @@ export const SubmissionComments: React.FC<{
...
-                  <TrashIcon />
-                </Button>
-              </div>
+              editing === i ? (
+                <div>
+                  <Button
+                    type="light"
+                    color="black"
+                    onClick={() => {
+                      handleDiscardChange()
```
anthonykhoa, line 137: we can simplify to
```
onClick={handleDiscardChange}
```

components/__snapshots__/DiffView.test.js.snap
anthonykhoa: The Button component should be updated so it can accept html attributes. Here is an example of materialui library doing so with their button component:

https://github.com/mui/material-ui/blob/902b4b070b88bf6ce713553a7a30d7a4c3e20c0a/packages/mui-material/src/Button/Button.js#L309

Making this change would definitely make things more convenient. For example you could pass `data-testid` which would allow for easier testing.

Also this is actually necessary so that we can be more semantically correct when using button elements too. For example, on the admin lessons page, the `update Lesson` button doesn't have `type=submit`, and currently that isn't possible.

HS-90: Ok that sounds good. Do you think it makes more sense to include in this PR, or create a separate issue for it? Im fine with either approach.

components/SubmissionComments.tsx
bryanjenningz: Sounds good, we can maybe add this as a prop for the Button component in the future so we can get better accessibility.

components/__snapshots__/DiffView.test.js.snap
flacial: Separate PR is better because this PR has its own issue to solve.

flacial (approved):

components/SubmissionComments.tsx
```diff
@@ -53,16 +128,54 @@ export const SubmissionComments: React.FC<{
...
+                  <Button
+                    type="info"
+                    size="sm"
+                    color="white"
+                    onClick={() => {
+                      handleStartEditing(i, c.content)
+                    }}
+                    data-testid="edit-button"
+                  >
+                    Edit
```
flacial, line 166: To keep the UI consistent, it's better to use an icon instead of a text like the delete comment trash icon

image

components/SubmissionComments.tsx
```diff
@@ -53,16 +128,54 @@ export const SubmissionComments: React.FC<{
...
+                <div>
+                  <Button
+                    type="light"
+                    color="black"
+                    onClick={() => {
+                      handleDiscardChange()
+                    }}
+                    data-testid="discard-button"
+                  >
+                    Discard Changes
```
flacial, line 141: I think this button needs to emphasize that what the user is about to do will remove what they've written in the input box. What do you think?

It could do so by making the button style `danger` like the `UNLINK DISCORD` button.

image

image

components/SubmissionComments.tsx
```diff
@@ -34,15 +41,83 @@ export const SubmissionComments: React.FC<{
...
+      <Modal show={show} onHide={handleModalClose}>
+        <Modal.Header>
+          <Modal.Title>Error</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>{modalText}</Modal.Body>
+        <Modal.Footer>
+          <Button
+            onClick={handleModalClose}
+            type="info"
+            size="lg"
```
flacial, line 99: The button is a bit large. I think medium or small will look better.

image

components/SubmissionComments.tsx
flacial: Extra margin between the two.

components/SubmissionComments.tsx
```diff
@@ -34,15 +41,83 @@ export const SubmissionComments: React.FC<{
...
+      </Modal>
+      {submissionsComments.map((c, i) => (
         <div
           key={`${c.content}${i}`}
           className="border border-lightgray rounded my-1"
         >
           <div className="ms-3">
-            <Markdown wrapper="code_wrapper">{c.content}</Markdown>
+            {editing === i ? (
+              <MdInput
```
flacial, line 113: Some `padding-top` + `margin-left` will make the input box centered like:

image

Currently, it's so close to the top border of its container.

flacial (approved):

songz (approved):

Issue #2080, opened 2022-07-13
Folders structure


flacial:
Each component will have its folder. This folder will contain the component and its: test, story, and styles

The component entry point will be called `index.tsx` so it can be easily imported (or it could be used to export to `[componentName].tsx`, so we don't end up with all the file names being `index.tsx`). The folder will also contain the CSS, test file, and story.

```
components/
  CopyButton/
    index.tsx
    CopyButton.tsx
    CopyButton.module.scss
    CopyButton.stories.ts
    CopyButton.test.js
```

`index.tsx` will contain the following, so we can do `import Component from '../components/ComponentFolder'`

`export { default } from './CopyButton.tsx'`

Our current folder structure for components files:
- Component test under `components/`
- Component code under `components/`
- Component styles under `styles/`
- Component story under `stories/`

evo777:
Folders are good.

scotthallock:
@flacial Is this something I can help out with? It would be nice to have everything in folders.

Also, I'm not sure I understand the problem with moving the `*.stories.tsx` files into the same folder as the component. Is it about the `stories` glob in `./storybook/main.js`?

I played around with the glob on the [Learn Storybook repo](https://github.com/chromaui/learnstorybook-code):

### Example 1
`stories: ["../src/components/*/*.stories.js"]` matches the stories in the `src/components/[ComponentName]/` folder but not elsewhere:

<img width="1224" alt="Screenshot 2023-04-02 at 5 23 41 AM" src="https://user-images.githubusercontent.com/121207468/229345237-9e87801b-4dfe-4247-a5f6-7e67e7fd6a60.png">

### Example 2
`stories: ["../src/**/*.stories.js"]` matches stories anywhere in the `src/` folder, at any level:
<img width="1224" alt="Screenshot 2023-04-02 at 5 36 21 AM" src="https://user-images.githubusercontent.com/121207468/229345249-bae9842b-88a8-4296-87ad-f329e8561b27.png">

### For c0d3 app
If we changed the glob to `"../**/*.stories.tsx"` - wouldn't that match **all** `[ComponentName].stories.tsx` files anywhere in our repo? 

This will also allow us to keep the pages and MDX stories wherever we want. I can try this it out later to check.

PR #2281, opened 2022-09-11
Add exercise preview cards to DOJO exercises page

bryanjenningz:
This pull request adds a new ExercisePreviewCard component which gets used on the DOJO exercises page.
The exercise preview cards are populated with mock data on the frontend, but this will be removed once the backend returns DOJO exercise preview data.

screenshot

This change doesn't affect any user-facing pages since there are no links to the /exercises/:lessonSlug page right now.

How to test:
* Go to /exercises/js0
* Make sure everything looks good

This pull request is a part of https://github.com/garageScript/c0d3-app/issues/2253.


SlyBouhafs:
The `answered` state badge looks a bit big, can we make it the same size as the `admin` badge?

bryanjenningz:
@SlyBouhafs Thanks for suggestion. I updated the badge size.

components/ExercisePreviewCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export type ExercisePreviewCardProps = {
+  topic: string
+  state: 'NOT ANSWERED' | 'ANSWERED'
+  problem: string
+  className?: string
+}
+
+const ExercisePreviewCard = ({
+  topic,
```
flacial, line 12: This should be named `moduleName`

components/ExercisePreviewCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <section
+      className={`card p-3 d-inline-block border-0 shadow position-relative overflow-hidden ${className}`}
+    >
+      <div className={topBorderStyle} />
+      <div className="d-flex align-items-center mb-3">
+        <h2 className="fw-bold fs-6 my-2 me-4">{topic.toUpperCase()}</h2>
+        <div className={`badge ${topMessageStyle}`}>{state}</div>
+      </div>
+      <div className="mb-2">Problem</div>
+      <pre className="bg-light py-2 px-3">{problem}</pre>
```
flacial, line 38: In the future, we need to look into highlighting the code. The code doesn't get highlighted in many places such as the comments and lesson/challenge description.

scss/exercisePreviewCard.module.scss
```diff
@@ -0,0 +1,35 @@
+@use './_variables.module';
+
+%exercisePreviewCard__topBorder {
```
flacial, line 3: What's this `%`?

flacial:
We decided to create a folder for each component. This folder will contain the component itself, its styles, and test file.

When creating new components, refer to https://github.com/garageScript/c0d3-app/issues/2080

flacial (approved):

scss/exercisePreviewCard.module.scss
bryanjenningz: It marks a block of code that can be extended with `@extend`
It apparently uses up less space in the compiled CSS compared to mixins because mixins will inline the CSS properties, but these will make a reused CSS class.
Look at the "Extend/Inheritance" section on this page: https://sass-lang.com/guide

components/ExercisePreviewCard.tsx
bryanjenningz: I agree, we can use something like react-syntax-highlighter: https://react-syntax-highlighter.github.io/react-syntax-highlighter/demo/

components/ExercisePreviewCard.tsx
bryanjenningz: Done

bryanjenningz:
@flacial Thanks for letting me know. I just changed the code so that the component file and styles file are in the same folder. I tried putting the storybook file in the same folder too but it didn't show up when I ran storybook, so I moved the storybook file back to the storybook folder with all the other storybook files.

PR #2304, opened 2022-09-14
Bootstrap 5.2.1 Upgrade

HS-90:
**Description**: related to #2266 

This PR upgrades bootstrap to the latest version. While fixing some of the styling issues that made certain buttons opaque instead of transparent(the correct way):
image

image

To test:

1. Login as the dummy Admin and go to `review/js0`
2. Hover over the edit(pencil icon) and delete(trash icon) comment buttons to see that they are now transparent on hover.

### DIFF
--- a/scss/button.module.scss
+++ b/scss/button.module.scss
@@ -16,4 +16,10 @@
   .btn-outline-bg-#{$name}:hover {
     background-color: rgba($color, 0.2);
   }
+
+  .btnOutlineInfoBgFix-#{$name} {
+    &:hover {
+      background-color: rgba($color, 0.2) !important;
+    }
+  }
 }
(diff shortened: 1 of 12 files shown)


scss/button.module.scss
```diff
@@ -16,4 +16,10 @@
   .btn-outline-bg-#{$name}:hover {
     background-color: rgba($color, 0.2);
   }
+
+  .btnOutlineInfoBgFix-#{$name} {
+    &:hover {
+      background-color: rgba($color, 0.2) !important;
```
flacial, line 22: `!important` is the only way we could fix it?

HS-90: without the `!important`, i couldn't get it to override the default `--bs-btn-hover-bg` setting

flacial (approved):

PR #2306, opened 2022-09-14
Update Admin Lesson Page Links

HS-90:
**Description:** related to issue #1959, and PR #2199

In the new Admin Lessons page, this PR updates the `Edit Lesson` links to navigate to the new introduction page created for each lesson in PR #2199. 

Test:
1. Login as dummy admin and go to `/admin/lessons/new`
2.  Click on each `Edit Lesson` link which should take you to the corresponding admin lesson introduction page.


components/admin/lessons/AdminLessonCard.tsx
```diff
@@ -31,7 +31,7 @@ export const AdminLessonCard: React.FC<Props> = ({
         {pendingFlaggedQuestions} Pending Questions
       </div>
 
-      <Link href={docUrl || '_blank'}>
+      <Link href={`/admin/lessons/${slug}/introduction`}>
```
anthonykhoa, line 34: We should have a constant made for `/admin` path already. If not, please make one since its used elsewhere in the codebase as well

anthonykhoa: Also, is the slug variable guaranteed to always exist?

flacial (approved):

components/admin/lessons/AdminLessonCard.tsx
HS-90: > Also, is the slug variable guaranteed to always exist?

Yeah. Every Lesson has a corresponding slug(js0-js9)

components/admin/lessons/AdminLessonCard.tsx
```diff
@@ -31,7 +32,7 @@ export const AdminLessonCard: React.FC<Props> = ({
         {pendingFlaggedQuestions} Pending Questions
       </div>
 
-      <Link href={docUrl || '_blank'}>
+      <Link href={ADMIN_PATH + `/lessons/${slug}/introduction`}>
```
anthonykhoa, line 35: admin path should be in the template literal as well

PR #2315, opened 2022-09-15
Add all button attributes to Button Component

### LINKED ISSUE #2292: Button component should take any html button attributes as a parameter.
Right now our custom Button component does not allow us to pass in certain html button attributes as a parameter. Being able to assign these will facilitate testing in the future. 

For example, being able to pass in a `data-testid' prop will allow for easier selection and simulation of clicking a Button in our test files. As of now we use a querySelector by class, which isn't as precise, nor intuitive.

HS-90:
**Description:** closes #2292

This PR allows engineers to pass in all standard html button attributes into the Button component, along with `data-testid`.

### DIFF
--- a/components/theme/Button.tsx
+++ b/components/theme/Button.tsx
@@ -13,6 +13,21 @@ type ButtonProps = {
   size?: 'lg' | 'sm'
   onClick?: Function
   outline?: boolean
+  testId?: string
+  autoFocus?: boolean
+  disabled?: boolean
+  name?: string
+  clickType?: 'submit' | 'reset' | 'button'
+  value?: string
+  form?: string
+  formAction?: string
+  formEncType?:
+    | 'application/x-www-form-urlencoded'
+    | 'multipart/form-data'
+    | 'text/plain'
+  formMethod?: 'post' | 'get'
+  formNoValidate?: boolean
+  formTarget?: string
 }
 
 export const Button: React.FC<ButtonProps> = ({
@@ -24,7 +39,19 @@ export const Button: React.FC<ButtonProps> = ({
   onClick = noop,
   children,
   size,
-  outline
+  outline,
+  testId,
+  autoFocus,
+  disabled,
+  name,
+  clickType,
+  value,
+  form,
+  formAction,
+  formEncType,
+  formMethod,
+  formNoValidate,
+  formTarget
 }) => {
   const classes = ['btn']
 
@@ -48,6 +75,18 @@ export const Button: React.FC<ButtonProps> = ({
       className={classes.join(' ')}
       onClick={() => onClick()}
       style={{ color: colors[color] }}
+      data-testid={testId}
+      autoFocus={autoFocus}
+      disabled={disabled}
+      name={name}
+      type={clickType}
+      value={value}
+      form={form}
+      formAction={formAction}
+      formEncType={formEncType}
+      formMethod={formMethod}
+      formNoValidate={formNoValidate}
+      formTarget={formTarget}
     >
       {children}
     </button>


components/theme/Button.tsx
```diff
@@ -48,6 +75,18 @@ export const Button: React.FC<ButtonProps> = ({
       className={classes.join(' ')}
       onClick={() => onClick()}
       style={{ color: colors[color] }}
+      data-testid={testId}
```
anthonykhoa, line 78: It would better to use the spread operator to implement this feature.

You can try out the following code for a POC I made on a react typescript playground:
https://codesandbox.io/embed/react-typescript-playground-j2dt6

```
import * as React from "react";
import { render } from "react-dom";

function Button({title, id, children, ...other}) {
	return <button title={title} id={id} {...other}>{children}</button>
}

function App() {
	return <Button title="fakeTitle" id="fakeId" type="fakeType">Potato</Button>;
}

render(<App />, document.getElementById("root"));

```

components/theme/Button.tsx
```diff
@@ -24,7 +39,9 @@ export const Button: React.FC<ButtonProps> = ({
   onClick = noop,
   children,
   size,
-  outline
+  outline,
+  testId,
```
anthonykhoa, line 43: I don't think we need to create a prop for testId. It should be treated like the other html attributes(...other)

components/theme/Button.tsx
```diff
@@ -24,7 +39,9 @@ export const Button: React.FC<ButtonProps> = ({
...
+  ...other
 }) => {
   const classes = ['btn']
```
anthonykhoa, line 46: What happens if someone passes in className as a prop? Will that override the current css of the button component, or will it just add on additional css?

components/theme/Button.tsx
```diff
@@ -13,6 +13,21 @@ type ButtonProps = {
   size?: 'lg' | 'sm'
   onClick?: Function
   outline?: boolean
+  testId?: string
```
anthonykhoa, line 16: You dont have to add all the html attributes yourself to the type definition
Try this:
```
interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
  border?: boolean
  type?: 'success' | 'danger' | 'primary' | 'light' | 'info' | 'mute'
  color?: ColorTypes
  m?: '1'
  ml?: '2'
  size?: 'lg' | 'sm'
  outline?: boolean
}
```

One other concern I have is with the `type` prop, since thats also an html attribute.

components/theme/Button.tsx
HS-90: When I try to pass in `className` manually: this is the type error:
`"message": "Type '{ children: string; className: string; type: \"light\"; color: \"black\"; onClick: () => void; \"data-testid\": string; }' is not assignable to type 'IntrinsicAttributes & ButtonProps & { children?: ReactNode; }'.\n  Property 'className' does not exist on type 'IntrinsicAttributes & ButtonProps & { children?: ReactNode; }'.",
`
I've encountered this in the past when adding the ability to edit comments. This made me think i needed to add all the html attributes manually to the type definition(related to your other comment).

components/theme/Button.tsx
HS-90: > One other concern I have is with the `type` prop, since thats also an html attribute, but is defined in the props as ` type?: 'success' | 'danger' | 'primary' | 'light' | 'info' | 'mute'`

I think this a bootstrap thing. It might be a good idea to change the variables to something like `btnType`?
This variable just affects what strings get added to the className, it's not actually passed into the `button` element.

anthonykhoa: Yeah thats a good idea.

components/theme/Button.tsx
anthonykhoa: I see, did my suggestion to use `interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement>` work to solve this issue?

I tried this out and got no errors:
```
interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
  potato?: string
}

function Button({title, id, children, potato, className, ...other} : ButtonProps) {
  const classes = "potatoClass" + className || ''
	return <button title={title} id={id} className={classes} {...other}>{children}{potato}</button>
}

export default function App() {
	return <Button title="fakeTitle" id="fakeId" type="submit" potato=" amazing" data-testid="testId">Potato</Button>;
}
```

anthonykhoa: One thing that will throw an error is `type`, because for buttons it can only have 3 values. But in our Button component in code its 
```
type?: 'success' | 'danger' | 'primary' | 'light' | 'info' | 'mute'
```

HS-90: > Also for `onClick`, we can remove explicitly defining that in button props(line 14) since it is an html attribute as well.

Extending the `ButtonProps` interface worked, but I'm going to keep the onClick explicitly defined in the button props, because leaving it out breaks the tests and it would make the PR too big. removing explicitly defining it would result in multiple snapshots being updated as well, so if it's something that's necessary I think it would be better to address it in a different PR.

HS-90: > This means we gotta get rename the `type` prop name to something else, as you suggested in the other thread. Also for `onClick`, we can remove explicitly defining that in button props(line 14) since it is an html attribute as well.

I was able address all your comments except for the onClick. I chose to explicitly state it because changing it was breaking many tests and would result in many snapshots having to be updated. This PR is already  big due to changing the `type` prop to `btnProp`(thus having to change it every file that used the `Button` component and passed in a `type` prop. I think if we want to focus on not explicitly defining `onClick` it should be in a future PR so that it will be easier to review.

flacial (approved):

components/theme/Button.tsx
```diff
@@ -46,8 +45,9 @@ export const Button: React.FC<ButtonProps> = ({
   return (
     <button
       className={classes.join(' ')}
```
anthonykhoa, line 47: What happens if someone passes in className as a prop? Will the passed-in className prop override line 47 ?`className={classes.join(' ')}`

components/Thanks.tsx
```diff
@@ -7,7 +7,14 @@ export const Thanks: React.FC<{ close: Function }> = ({ close }) => (
     <h2 className="mt-4 mb-4 pt-3 pb-3 fw-bold text-center">
       Thanks for letting us know!
     </h2>
-    <Button type="primary" color="white" size="lg" onClick={close}>
+    <Button
+      btnType="primary"
+      color="white"
+      size="lg"
+      onClick={() => {
```
anthonykhoa, line 14: Why is there an update from `onClick={close}` to 

```
onClick={() => {
  close()
 }}
```
?

Is there an error otherwise?

components/theme/Button.tsx
anthonykhoa: On line 28 can you try `const classes = ['btn', className]`
And if className doesnt exist(`aka not passed as a prop`) you can set default value of className as empty string

HS-90: > After line 28 can you try
> 
> ```
> className && classes.push(className)
> ```

Yeah this is a good thing to point out. It works, but I have to explicitly state the `className` prop

components/Thanks.tsx
HS-90: > Why is there an update from `onClick={close}` to
> 
> ```
> onClick={() => {
>   close()
>  }}
> ```
> 
> ?

With `onclick={close}` I was getting this error:
`Type 'Function' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.
  Type 'Function' provides no match for the signature '(event: MouseEvent<HTMLButtonElement, MouseEvent>): void'.`

Changing it seemed to fix the issue.

components/theme/Button.tsx
anthonykhoa: https://github.com/mui/material-ui/blob/master/packages/mui-joy/src/Button/Button.tsx#L60

materialsui button uses styled components, but that currently isn't our setup for setting css in c0d3

anthonykhoa (approved):

components/theme/Button.tsx
HS-90: > > It works, but I have to explicitly pass in the `className` prop
> 
> I don't understand, what is the issue with doing that? Is there a better way to handle this use case for this repo?

Oh I have no issue with it, just more of an FYI since we were discussing trying to explicitly state as few props as possible.

PR #2316, opened 2022-09-16
Move new button to Button component file

bryanjenningz:
Moving the button to a separate component will be helpful so the button can be reused in the DOJO exercise card (https://github.com/garageScript/c0d3-app/issues/2253).

Here's what it looks like in storybook:
new-buttons

### DIFF
--- a/components/FormCard.tsx
+++ b/components/FormCard.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from 'react'
-import { Button } from './theme/Button'
+import { Button, NewButton } from './theme/Button'
 import { MdInput } from './MdInput'
 import { DropdownMenu, Item } from './DropdownMenu'
 import _ from 'lodash'
@@ -169,12 +169,7 @@ export const FormCard: React.FC<FormCardProps> = ({
         {submitError && <ErrorMessage error={submitError} />}
         <div className={`text-center align-self-${textAlignment}`}>
           {newBtn ? (
-            <button
-              onClick={btnOnClick}
-              className={styles['formCard--submitButton']}
-            >
-              {onSubmit.title}
-            </button>
+            <NewButton onClick={btnOnClick}>{onSubmit.title}</NewButton>
           ) : (
             <Button onClick={btnOnClick} type="primary" color="white">
               {onSubmit.title}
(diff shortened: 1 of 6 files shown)


components/FormCard.tsx
```diff
@@ -169,12 +169,7 @@ export const FormCard: React.FC<FormCardProps> = ({
         {submitError && <ErrorMessage error={submitError} />}
         <div className={`text-center align-self-${textAlignment}`}>
           {newBtn ? (
-            <button
-              onClick={btnOnClick}
-              className={styles['formCard--submitButton']}
-            >
-              {onSubmit.title}
-            </button>
+            <NewButton onClick={btnOnClick}>{onSubmit.title}</NewButton>
```
bryanjenningz, line 172: I wasn't sure what to call the new button, so I just named it NewButton since the `newBtn` variable was used to refer to it.
Other possible names we could have used: MaterialButton, ResponsiveButton
Let me know if you think of a good name.
We can always change the name in the future if we think of a better name.

HS-90 (approved):

anthonykhoa:
What is the purpose of this NewButton Component?

anthonykhoa:
Oh i see, it replaced a button in the FormCard component

anthonykhoa:
Why do we need to create a new button? Is the button component we currently have not enough? Is it unable to be updated to solve your use case?

bryanjenningz:
> Why do we need to create a new button? Is the button component we currently have not enough? Is it unable to be updated to solve your use case?

We need to create a new button because the Button component doesn't have the same features (background color, nice transitions, hover/active states) as the new button and this new button is used in the form and it will be used on the DOJO exercises page in multiple spots.

We could have also added extra props to the Button component which would add the features we need, but I couldn't think of a good way to do that. Maybe we could have done this `<Button newBtn>` which would add the new background color, nice transitions, and hover/active states, but it seems like it would be more simple to just have a separate button since this new button doesn't need any of the features of the Button component.

components/FormCard.tsx
flacial: `NewButton` is good.

flacial (approved):

PR #2323, opened 2022-09-17
Add DOJO exercise card

bryanjenningz:
Add ExerciseCard from the DOJO exercise card [designs](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1405%3A7764).

This pull request has no visible changes to the website.

The ExerciseCard component will be used on the DOJO exercises page (e.g. /exercises/js0) in a subsequent pull request.

not-answered-yet
incorrect-answer
correct-answer

### DIFF
--- a/components/ExerciseCard/ExerciseCard.tsx
+++ b/components/ExerciseCard/ExerciseCard.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { NewButton } from '../theme/Button'
 import { Text } from '../theme/Text'
+import styles from './exerciseCard.module.scss'
 
 export type ExerciseCardProps = {
   challengeName: string
@@ -32,8 +33,8 @@ const ExerciseCard = ({
             value={studentAnswer}
             onChange={e => setStudentAnswer(e.target.value)}
           />
-          <div className={`my-2 ${message === '' ? 'invisible' : ''}`}>
-            <Text size="sm">{message === '' ? '.' : message}</Text>
+          <div className={`${styles.exerciseCard__message} my-2`}>
+            <Text size="sm">{message}</Text>
           </div>
           <div className="d-flex">
             <NewButton
--- a/components/ExerciseCard/ExerciseCard.tsx
+++ b/components/ExerciseCard/ExerciseCard.tsx
@@ -7,22 +7,20 @@ export type ExerciseCardProps = {
   problem: string
   answer: string
   explanation: string
-  className?: string
 }
 
 const ExerciseCard = ({
   challengeName,
   problem,
   answer,
-  explanation,
-  className = ''
+  explanation
 }: ExerciseCardProps) => {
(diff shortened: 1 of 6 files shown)


components/ExerciseCard/ExerciseCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+  explanation: string
+  className?: string
+}
+
+const ExerciseCard = ({
+  challengeName,
+  problem,
+  answer,
+  explanation,
+  className = ''
```
anthonykhoa, line 18: What is the need for this `className` prop?

components/ExerciseCard/ExerciseCard.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <div className="w-50 ms-3">
+          <div className="mb-2">{challengeName}</div>
+          <input
+            aria-label="User answer"
+            className="form-control mb-2"
+            value={studentAnswer}
+            onChange={e => setStudentAnswer(e.target.value)}
+          />
+          <div className={`my-2 ${message === '' ? 'invisible' : ''}`}>
+            <Text size="sm">{message === '' ? '.' : message}</Text>
```
bryanjenningz, line 38: The reason why I put a "." here is because I wanted the text to retain its height so that there wouldn't be a content shift. There might be a better way to do this. Maybe I can add a min-height CSS property to this element.

components/ExerciseCard/ExerciseCard.tsx
bryanjenningz: It can be useful for adding CSS classes. For example if you want to add margin-bottom you can pass in "mb-2" `<ExerciseCard className="mb-2" ... />`.

anthonykhoa: Are there any pages in the designs that require this use case?

bryanjenningz: I'm not sure yet. Maybe I can remove the className prop until it's actually needed.

anthonykhoa: Yeah lets remove it for now

components/ExerciseCard/ExerciseCard.tsx
anthonykhoa: Yeah I agree that min-height would be a better solution than using `.`, its less hacky since we're explicitly setting the css rules

components/ExerciseCard/ExerciseCard.tsx
bryanjenningz: Done

components/ExerciseCard/ExerciseCard.tsx
bryanjenningz: Done.

anthonykhoa (approved):
LGTM :+1:

PR #2325, opened 2022-09-18
Add exercise card to DOJO exercises page

bryanjenningz:
This pull request doesn't change any user-facing pages, it only changes the DOJO exercises pages (e.g. /exercises/js0) which currently doesn't have any links to it.

This pull request adds the exercise card to the DOJO exercises page. It currently uses mock data on the frontend and doesn't store the user's answers. It will be connected to the backend and real data in a subsequent pull request.

How to test:
* Go to /exercises/js0
* Click "SOLVE EXERCISES" button
* Answer questions, show and hide solutions, click "SKIP" button, click "PREVIOUS" button, click "Exit" button

The newly added "SOLVED EXERCISES" button on the right side of the /exercises/js0 page:
solve-exercises-button

The first exercise:
solve-exercise

The solved second exercise:
solve-exercise-2

### DIFF
--- a/pages/exercises/[lessonSlug].tsx
+++ b/pages/exercises/[lessonSlug].tsx
@@ -61,7 +61,7 @@ const mockExercises: ExerciseCardProps[] = [
 const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
   const { lessons, alerts } = queryData
   const router = useRouter()
-  const [exerciseIndex, setExerciseIndex] = useState<number>(-1)
+  const [exerciseIndex, setExerciseIndex] = useState(-1)
   if (!router.isReady) return <LoadingSpinner />
 
   const slug = router.query.lessonSlug as string
@@ -125,10 +125,7 @@ const Exercise = ({
         className="btn ps-0 d-flex align-items-center"
         onClick={() => setExerciseIndex(-1)}
       >
-        <ArrowLeftIcon size="medium" />
-        <div className="p-3">
-          <Text size="xs">Exit</Text>
-        </div>
+        <ArrowLeftIcon size="medium" aria-label="Exit" />
       </button>
 
       <h1 className="mb-4 fs-2">{lessonTitle}</h1>
--- a/pages/exercises/[lessonSlug].tsx
+++ b/pages/exercises/[lessonSlug].tsx
@@ -138,7 +138,8 @@ const Exercise = ({
         {showPreviousButton ? (
           <button
             onClick={() => setExerciseIndex(i => i - 1)}
-            className="btn btn-outline-primary"
+            className="btn btn-outline-primary fw-bold px-4 py-2"
+            style={{ fontFamily: 'PT Mono', fontSize: 14 }}
           >
             PREVIOUS
           </button>
@@ -148,7 +149,8 @@ const Exercise = ({
         {showSkipButton ? (
           <button
             onClick={() => setExerciseIndex(i => i + 1)}
-            className="btn btn-outline-primary"
+            className="btn btn-outline-primary fw-bold px-4 py-2"
+            style={{ fontFamily: 'PT Mono', fontSize: 14 }}
           >
             SKIP
           </button>
--- a/pages/exercises/[lessonSlug].tsx
+++ b/pages/exercises/[lessonSlug].tsx
@@ -1,5 +1,5 @@
 import { useRouter } from 'next/router'
-import React from 'react'
+import React, { useState } from 'react'
 import Layout from '../../components/Layout'
 import withQueryLoader, {
   QueryDataProps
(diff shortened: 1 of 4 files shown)


flacial:
More margin (32px) between the main content and the nav would be nice 

image

pages/exercises/[lessonSlug].tsx
```diff
@@ -13,20 +13,55 @@ import NavCard from '../../components/NavCard'
...
+    answer: `'510'`,
+    explanation:
+      'If you add a string and number together, the number gets converted to a string and then they get concatenated.'
+  }
+]
+
 const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
   const { lessons, alerts } = queryData
   const router = useRouter()
+  const [exerciseIndex, setExerciseIndex] = useState<number>(-1)
```
Comment (flacial) [inline pages/exercises/[lessonSlug].tsx:64]: Will something throw an error if we didn't specify the type? If no, I think it can be removed because TS is capable of inferring the type in this case.

pages/exercises/[lessonSlug].tsx
```diff
@@ -45,14 +80,113 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
     { text: 'exercises', url: `/exercises/${currentLesson.slug}` }
   ]
 
+  const exercise = mockExercises[exerciseIndex]
+
   return (
     <Layout title={currentLesson.title}>
```
Comment (flacial) [inline pages/exercises/[lessonSlug].tsx:86]: Do you think the admins should be the only ones with access to this page, or does it matter since there aren't links that users can use to access this page?

pages/exercises/[lessonSlug].tsx
```diff
@@ -45,14 +80,113 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
...
+}: ExerciseProps) => {
+  return (
+    <div className="w-75 mx-auto">
+      <button
+        className="btn ps-0 d-flex align-items-center"
+        onClick={() => setExerciseIndex(-1)}
+      >
+        <ArrowLeftIcon size="medium" />
+        <div className="p-3">
+          <Text size="xs">Exit</Text>
```
Comment (flacial) [inline pages/exercises/[lessonSlug].tsx:130]: I'm wondering. This exists in the design but do we need the text `Exit`? Isn't the icon enough to indicate that it's an exit button?

flacial:
The error is different than the design:

image
 
Should be:

flacial:
Other features such as flagging and the filtering dropdown will be addressed in a subsequent PR?

pages/exercises/[lessonSlug].tsx
```diff
@@ -45,14 +80,113 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
...
-      <h1 className="my-4">{currentLesson.title}</h1>
-      {alerts && <AlertsDisplay alerts={alerts} />}
+      <div className="d-flex justify-content-between align-items-center">
+        <h1 className="my-4 fs-2">{lessonTitle}</h1>
+        <NewButton onClick={() => setExerciseIndex(0)}>
+          SOLVE EXERCISES
+        </NewButton>
+      </div>
       <div className="container">
         <div className="row">
```
Comment (flacial) [inline pages/exercises/[lessonSlug].tsx:191]: After connecting it with the backend, it'll then be updated to show many exercises (20-30) in a good layout?

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:86]: It shouldn't matter since there aren't any links that point to this page.

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:130]: Yeah, we can remove it and label the exit button with `aria-label="Exit"`.

bryanjenningz:
> Other features such as flagging and the filtering dropdown will be addressed in a subsequent PR?

Yeah, I'm just trying to get the features done in small pull requests so they're easy to review. We can include more stuff in later pull requests.

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:191]: Yes.

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:64]: Yeah, good catch, I can remove the type parameter so it's just `useState(-1)`.

bryanjenningz:
> The error is different than the design:
> 
> image
> 
> Should be:
> 
> image
> 
> With red border and red text.
> 
> Actually, this is related to the previous PR and shouldn't be addressed in this PR.

Thanks! Fixed in https://github.com/garageScript/c0d3-app/pull/2326

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx]: Done.

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx]: Done.

flacial (approved):

PR #2326, opened 2022-09-18
Make exercise card error message red

bryanjenningz:
red-error-message

This is a part of https://github.com/garageScript/c0d3-app/issues/2253

### DIFF
--- a/components/ExerciseCard/ExerciseCard.tsx
+++ b/components/ExerciseCard/ExerciseCard.tsx
@@ -10,6 +10,14 @@ export type ExerciseCardProps = {
   explanation: string
 }
 
+const messages = {
+  EMPTY: '',
+  ERROR: 'Your answer is incorrect - please try again.',
+  SUCCESS: '🎉 Your answer is correct!'
+} as const
+
+type MessageKey = keyof typeof messages
+
 const ExerciseCard = ({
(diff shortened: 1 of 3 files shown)


components/ExerciseCard/ExerciseCard.tsx
```diff
@@ -10,15 +10,24 @@ export type ExerciseCardProps = {
   explanation: string
 }
 
+const messages = {
```
flacial, line 13: Why an object instead of an `enum`?

flacial (approved):

components/ExerciseCard/ExerciseCard.tsx
bryanjenningz: There's no specific reason, so I'll switch this to an enum. 
```typescript
const messages = {
  EMPTY: '',
  ERROR: 'Your answer is incorrect - please try again.',
  SUCCESS: '🎉 Your answer is correct!'
} as const
type MessageKey = keyof typeof messages
```
```typescript
enum Message {
  EMPTY = '',
  ERROR = 'Your answer is incorrect - please try again.',
  SUCCESS = '🎉 Your answer is correct!'
}
type MessageKey = keyof typeof Message
```

PR #2329, opened 2022-09-20
Fix buttons after bootstrap update

### LINKED ISSUE #2313: Fix styles broken by bootstrap
AppNav:

image

image

Review submissions page:

Lacks focus state styles
image

These are some of the issues I found. If you'll take this, please go through the website and ensure everything looks good (hover, focus, and active states; sizes, colors, fonts..etc)

HS-90:
**Description:** closes #2313 

In this PR the hover functionalities for these buttons were fixed following the latest bootstrap update:
- Login and Signup buttons in AppNav
- Login with Discord' button on login page.

I looked through the site, but in the case I did miss something please feel free to point it out.

### DIFF
--- a/scss/appNav.module.scss
+++ b/scss/appNav.module.scss
@@ -79,6 +79,18 @@
   }
 }
 
+.btn-hoverFix {
+  &:hover,
+  &:focus {
+    background-color: color.change(
+      variables.$primary,
+      $lightness: 90%
+    ) !important;
+    color: variables.$primary !important;
+    border: 1px solid variables.$primary !important;
+  }
+}
+
 .isAdminBadge {
   font-size: 11px;
   letter-spacing: 0.07em;
--- a/scss/appNav.module.scss
+++ b/scss/appNav.module.scss
@@ -79,15 +79,14 @@
   }
 }
 
-.btn-hoverFix {
+.login-signup-size {
+  margin: 8px;
+  padding: 10.2px 13.6px;
+  border-radius: 0.375rem;
   &:hover,
   &:focus {
-    background-color: color.change(
-      variables.$primary,
-      $lightness: 90%
-    ) !important;
-    color: variables.$primary !important;
-    border: 1px solid variables.$primary !important;
+    border: 1px solid variables.$primary;
+    padding: 9.2px 12.6px;
   }
 }
 
--- a/__tests__/__snapshots__/storyshots.test.js.snap
+++ b/__tests__/__snapshots__/storyshots.test.js.snap
@@ -27649,7 +27649,7 @@ exports[`Storyshots Pages/Login Login 1`] = `
           />
         </div>
         <button
-          className="btn text-white btn-lg btn discord__button"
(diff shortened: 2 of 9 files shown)


scss/appNav.module.scss
```diff
@@ -79,6 +79,18 @@
   }
 }
 
+.btn-hoverFix {
+  &:hover,
+  &:focus {
+    background-color: color.change(
+      variables.$primary,
+      $lightness: 90%
+    ) !important;
```
anthonykhoa, line 88: is it possible to not use `!important`? Its best to be avoided because it breaks natural cascading

flacial (commented):
I'm wondering why they're being overridden by bootstrap. NavLink doesn't use any bootstrap component or styles.

scss/appNav.module.scss
HS-90: I've tried using an `id` , using multiple classes as a selector to add more specificity. I wanted to try inline style, but saw that was not recommended either, also not sure if it's possible to add pseudoclasses with inline. 

example of what i tried with id and adding specificity:

appNav.tsx:
`
<div className={`${styles['nav-buttons']}`} id="btnOverride">
    <NavLink
      path={LOGIN_PATH}
      className={`btn m-2 me-lg-3 ${styles['login-btn']} ${styles['btn-hoverFix']}`}
    >
      Login
    </NavLink>
    <NavLink
      path={SIGNUP_PATH}
      className={`btn m-2 me-lg-3 ${styles['signup-btn']} ${styles['btn-hoverFix']}`}
    >
      Signup
    </NavLink>
  </div>
`
appNav.module.scss file:
`#btnOverride .nav-buttons .btn-hoverFix{
  &:hover,
  &:focus {
    background-color: color.change(
      variables.$primary,
      $lightness: 90%
    );
    color: variables.$primary;
    --bs-btn-hover-color: variables.$primary;
    border: 1px solid variables.$primary;
  }
}
`

flacial:
> > I'm wondering why they're being overridden by bootstrap. NavLink doesn't use any bootstrap component or styles.
> 
> I think it has to do with this: image
> 
> I'm not sure how to access the file that would change this css in our codebase though.

Which has to do with this:

https://github.com/garageScript/c0d3-app/blob/4588408a6a1c435591cdeff6e767ad051b20019d/scss/index.scss#L41

HS-90:
> @HS-90 In order to not use `!important`, we should remove this class and give the button our own styles.

Does this mean we will have to do a mass refactoring of our Buttons and AppNavs? Majority of our buttons have 'btn' in the className.

flacial:
> > @HS-90 In order to not use `!important`, we should remove this class and give the button our own styles.
> 
> Does this mean we will have to do a mass refactoring of our Buttons and AppNavs? Majority of our buttons have 'btn' in the className.

For now, we can refactor these only and later refactor the reset if the issue spread.

bryanjenningz (approved):

PR #2335, opened 2022-09-20
Get exercises from backend on DOJO exercises page

bryanjenningz:
This pull request doesn't affect any user-facing pages since there are no direct links to the DOJO exercises page.

This pull request makes it so the DOJO exercises page gets the exercises from the backend instead of relying on mock data on the frontend.

How to test:
* Create exercises at /curriculum/js0/mentor
* Go to /exercises/js0 and click on the "SOLVE EXERCISES" button on the right and verify the exercises still work

The exercise page currently doesn't send any mutations to the backend. We will add mutations to update the user's answers to the exercises on the backend in a subsequent pull request.

### DIFF
--- a/graphql/queries/getExercises.ts
+++ b/graphql/queries/getExercises.ts
@@ -1,0 +1,31 @@
+import { gql } from '@apollo/client'
+
+const GET_EXERCISES = gql`
+  query GetExercises {
+    lessons {
+      title
+      docUrl
+      slug
+    }
+    alerts {
+      id
+      text
+      type
+      url
+      urlCaption
+    }
+    exercises {
+      module {
+        name
+        lesson {
+          slug
+        }
+      }
+      description
+      answer
+      explanation
+    }
+  }
+`
+
+export default GET_EXERCISES
--- a/pages/exercises/[lessonSlug].tsx
+++ b/pages/exercises/[lessonSlug].tsx
@@ -4,10 +4,9 @@ import Layout from '../../components/Layout'
 import withQueryLoader, {
   QueryDataProps
 } from '../../containers/withQueryLoader'
-import { GetAppQuery } from '../../graphql'
+import { GetExercisesQuery } from '../../graphql'
 import Error, { StatusCode } from '../../components/Error'
 import LoadingSpinner from '../../components/LoadingSpinner'
-import GET_APP from '../../graphql/queries/getApp'
 import AlertsDisplay from '../../components/AlertsDisplay'
 import NavCard from '../../components/NavCard'
 import ExercisePreviewCard, {
@@ -16,17 +15,10 @@ import ExercisePreviewCard, {
 import { NewButton } from '../../components/theme/Button'
 import ExerciseCard, { ExerciseCardProps } from '../../components/ExerciseCard'
 import { ArrowLeftIcon } from '@primer/octicons-react'
+import GET_EXERCISES from '../../graphql/queries/getExercises'
 
 const exampleProblem = `const a = 5
 a = a + 10
-// what is a?`
-
-const exampleProblem2 = `let a = 5
-a = a + 10
-// what is a?`
-
-const exampleProblem3 = `let a = 5
-a = a + '10'
 // what is a?`
 
 const mockExercisePreviews: ExercisePreviewCardProps[] = [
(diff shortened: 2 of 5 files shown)


pages/exercises/[lessonSlug].tsx
```diff
@@ -79,7 +51,16 @@ const Exercises: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
...
   ]
 
-  const exercise = mockExercises[exerciseIndex]
+  const currentExercises = exercises
+    .filter(exercise => exercise?.module.lesson.slug === slug)
+    .map(exercise => ({
+      challengeName: exercise?.module.name!,
+      problem: exercise?.description!,
+      answer: exercise?.answer!,
+      explanation: exercise?.explanation!
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:60]: Using the `?.` and `!` operators here isn't ideal.

The reason why we need to do this is because the type definition for the exercises is `exercises: [Exercise]!` instead of `exercises: [Exercise!]!`: https://github.com/garageScript/c0d3-app/blob/master/graphql/typeDefs.ts#L9

Would it be better to change the schema to `exercises: [Exercise!]!` so that we don't have to worry about null values?

SlyBouhafs:
This file [graphql/index.tsx](https://github.com/garageScript/c0d3-app/pull/2335/files#diff-59cbb50dff68be6b357ab7fa62dba199ddd707e060f498abd5a588679a91e5ad) has the same diff as the file in this PR: https://github.com/garageScript/c0d3-app/pull/2330/files

bryanjenningz:
@SlyBouhafs 
> This file [graphql/index.tsx](https://github.com/garageScript/c0d3-app/pull/2335/files#diff-59cbb50dff68be6b357ab7fa62dba199ddd707e060f498abd5a588679a91e5ad) has the same diff as the file in this PR: https://github.com/garageScript/c0d3-app/pull/2330/files

Do you want me to close that pull request since it's overlapping this one or do you want me to merge that pull request so that this pull request has a smaller diff?

graphql/queries/getExercises.ts
```diff
@@ -0,0 +1,31 @@
+import { gql } from '@apollo/client'
+
+const GET_EXERCISES = gql`
```
bryanjenningz, line 3: I wasn't sure what to call this query. It does more than just get the exercises - it gets the lessons and alerts too which are used on the DOJO exercises page.

SlyBouhafs:
> Do you want me to close that pull request since it's overlapping this one or do you want me to merge that pull request so that this pull request has a smaller diff?

I think it's better if we keep this file out of this PR.

bryanjenningz:
> > Do you want me to close that pull request since it's overlapping this one or do you want me to merge that pull request so that this pull request has a smaller diff?
> 
> I think it's better to keep this file out of this PR.

I also added a new query in this pull request which changed this file as well. Should I move the new query I added to a separate pull request so that this file diff is not in this pull request?

SlyBouhafs:
> I also added a new query in this pull request which changed this file as well. Should I move the new query I added to a separate pull request so that this file diff is not in this pull request?

It's better if we merge the other PR, then have a new diff with that change only.

bryanjenningz:
> It's better if we merge the other PR, then have a new diff with that change only.

Sounds good.

graphql/queries/getExercises.ts
flacial: `GET_DOJO_EXERCISES` maybe, but it's longer.

pages/exercises/[lessonSlug].tsx
Comment (flacial) [reply pages/exercises/[lessonSlug].tsx:60]: I thought we should do that once there are some exercises. Wouldn't it throw an error when getting the exercises if there are none?

flacial (approved):

flacial:
This is unrelated to this PR; I just noticed it.

Doesn't it make sense for the input title to be `Answer` or `Your answer` rather than the module name?

image

flacial (approved):
Reapproving.

Super nice, I like it

image

songz (approved):

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:60]: If we changed the type definition to `exercises: [Exercise!]!` and if there are no exercises (i.e. it returns an empty array `[]`), then it shouldn't throw an error because `[]` is the type `[Exercise!]!`.

It would only throw an error if there were null values in the array (e.g. `[null, null]`).

graphql/queries/getExercises.ts
bryanjenningz: Yeah, or maybe even something like `GET_DOJO_EXERCISES_PAGE_DATA`  since we're getting all the data we need for the DOJO exercises page.

bryanjenningz:
> This is unrelated to this PR; I just noticed it.
> 
> Doesn't it make sense for the input title to be `Answer` or `Your answer` rather than the module name?
> 
> image

Yeah, I actually agree with you on this. I was following the design, but maybe we should change the design so it says "Your answer" or something similar instead.

bryanjenningz:
I'll merge this pull request since it's mostly good and everything works. We can fix the [exercises type definition](https://github.com/garageScript/c0d3-app/blob/master/graphql/typeDefs.ts#L9) to be `[Exercise!]!` in a separate pull request which will make things more simple and we can also possibly handle changing the naming convention of the DOJO exercises queries in a separate pull request too.

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:60]: I added a pull request for if we want to make this change: https://github.com/garageScript/c0d3-app/pull/2336

PR #2336, opened 2022-09-21
Make DOJO exercises not nullable and fix React key bug

bryanjenningz:
This pull request doesn't affect any user-facing pages.

This pull request makes the exercises type definition on the backend non-nullable (i.e. `[Exercise!]!`) which makes the data easier to deal with. This won't cause any problems because we never return any null values in the exercises array.

This pull request also fixes a small React key bug which made it possible to previously press "Show Answer" then "Skip" which would make it go to the next exercise but the next exercise would already be showing the answer. Now if you press "Show Answer" then "Skip" the next answer will not be showing because we added a `key={exerciseIndex}` property to the Exercise component which causes a rerender when `exerciseIndex` changes.

We also safely handle the case where there is no exercise explanation (notice the bottom right is empty because this exercise happens to not have an explanation):
empty-explanation

How to test:
* Create exercises at /curriculum/js0/mentor
* Go to /exercises/js0 and click on the "SOLVE EXERCISES" button on the right and verify the exercises still work

### DIFF
--- a/__dummy__/getExercisesData.ts
+++ b/__dummy__/getExercisesData.ts
@@ -87,7 +87,7 @@ const getExercisesData: GetExercisesQuery = {
       },
       description: '```\nlet a = 1\na -= 2\n// what is a?\n```',
       answer: '-1',
-      explanation: '`a -= 2` is a shorter way to write `a = a - 2`'
+      explanation: null
     }
   ]
 }
--- a/__tests__/pages/exercises/[lessonSlug].test.js
+++ b/__tests__/pages/exercises/[lessonSlug].test.js
@@ -64,10 +64,11 @@ describe('Exercises page', () => {
     // Previous button is not in the document on the first exercise.
     expect(queryByRole('button', { name: 'PREVIOUS' })).not.toBeInTheDocument()
 
-    const skipButton = getByRole('button', { name: 'SKIP' })
+    let skipButton = getByRole('button', { name: 'SKIP' })
     fireEvent.click(skipButton)
     expect(queryByRole('button', { name: 'PREVIOUS' })).toBeInTheDocument()
 
+    skipButton = getByRole('button', { name: 'SKIP' })
     fireEvent.click(skipButton)
     // Skip button should not be in the document because we're on the last exercise now.
     expect(queryByRole('button', { name: 'SKIP' })).not.toBeInTheDocument()
--- a/pages/exercises/[lessonSlug].tsx
+++ b/pages/exercises/[lessonSlug].tsx
@@ -54,10 +54,10 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
   const currentExercises = exercises
     .filter(exercise => exercise?.module.lesson.slug === slug)
     .map(exercise => ({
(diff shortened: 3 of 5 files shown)


pages/exercises/[lessonSlug].tsx
```diff
@@ -66,6 +66,7 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
     <Layout title={currentLesson.title}>
       {exercise ? (
         <Exercise
+          key={exerciseIndex}
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:69]: This line fixes a small React key bug which previously would make it possible to press "Show Answer" (to show the current exercise answer) then "Skip" (to skip to the next exercise) which caused the next exercise's answer to be shown.

pages/exercises/[lessonSlug].tsx
```diff
@@ -54,10 +54,10 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
...
     .filter(exercise => exercise?.module.lesson.slug === slug)
     .map(exercise => ({
-      challengeName: exercise?.module.name!,
-      problem: exercise?.description!,
-      answer: exercise?.answer!,
-      explanation: exercise?.explanation!
+      challengeName: exercise.module.name,
+      problem: exercise.description,
+      answer: exercise.answer,
+      explanation: exercise.explanation || ''
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:60]: [Exercise explanations are nullable](https://github.com/garageScript/c0d3-app/blob/master/prisma/schema.prisma#L187) so we have to handle it here by defauting to an empty string, which works fine and looks good (see pull request description for a picture of what it looks like).

__tests__/pages/exercises/[lessonSlug].test.js
```diff
@@ -64,10 +64,11 @@ describe('Exercises page', () => {
     // Previous button is not in the document on the first exercise.
     expect(queryByRole('button', { name: 'PREVIOUS' })).not.toBeInTheDocument()
 
-    const skipButton = getByRole('button', { name: 'SKIP' })
+    let skipButton = getByRole('button', { name: 'SKIP' })
     fireEvent.click(skipButton)
     expect(queryByRole('button', { name: 'PREVIOUS' })).toBeInTheDocument()
 
+    skipButton = getByRole('button', { name: 'SKIP' })
```
Comment (bryanjenningz) [inline __tests__/pages/exercises/[lessonSlug].test.js:71]: The reason why we need to reassign `skipButton` to the current skip button in the DOM is because we are using a React key prop which makes it so the Exercise component rerenders.

__dummy__/getExercisesData.ts
```diff
@@ -87,7 +87,7 @@ const getExercisesData: GetExercisesQuery = {
       },
       description: '```\nlet a = 1\na -= 2\n// what is a?\n```',
       answer: '-1',
-      explanation: '`a -= 2` is a shorter way to write `a = a - 2`'
+      explanation: null
```
bryanjenningz, line 90: Adding a null exercise explanation to the mock exercise data so we can get full test coverage.

HS-90 (approved):

Issue #2252, opened 2022-09-02
DOJO(discussions): Create the Discussions component


flacial:
# Discussions

[The discussions component](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=404%3A13721) will be used for students to ask and answer questions about the problem.

## Research

[Look into Reddit implementation of Threads.](https://www.reddit.com/r/AskReddit/comments/t0ynr/throwaway_time_whats_your_secret_that_could/)

## Issues/Questions

- Its design isn't complete. Feel free to apply any appropriate modifications, and make sure to discuss them with the engineers if they're major.
- It needs pagination
- Each message can only have 1 layer of nested replies?
   - If not, when a single message has 10 nested replies, how will they be displayed?

PR #2339, opened 2022-09-23
Discussions Card Component Design Only

HS-90:
**Description:** relates to #2252 

This PR creates the initial design for the two different types of discussion cards as shown in this design:
https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3

The submitted PR is not designed 1 to 1 compared to figma, but we can discuss and come up with a finalized design if necessary. 
You can view this design by running `yarn storybook`  and view `DiscussionsCard`.  Additionally, here are screenshots of the design below:

 **Main Card:**
maincard

**Sub Card:**
sub card

### DIFF
--- a/components/DiscussionsCard/DiscussionsCard.tsx
+++ b/components/DiscussionsCard/DiscussionsCard.tsx
@@ -1,0 +1,82 @@
+import React from 'react'
+import styles from './discussionsCard.module.scss'
+import {
+  ReplyIcon,
+  ThumbsupIcon,
+  ThumbsdownIcon,
+  TrashIcon,
+  PencilIcon
+} from '@primer/octicons-react'
+
+export type DiscussionsCardProps = {
+  discussionType: 'main' | string
+  username: string
+  userPic: string
+  timeStamp: string
+  content: string
+  likes: number
+  dislikes: number
+  expandClick: Function
+  likeClick: Function
+  dislikeClick: Function
+}
+
+const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
+  discussionType,
+  username,
+  userPic,
+  timeStamp,
+  content,
+  likes,
+  dislikes,
+  expandClick,
+  likeClick,
+  dislikeClick
+}) => {
+  const mainType = discussionType === 'main'
+
+  return (
+    <div className={mainType ? styles.card_body_main : styles.card_body_sub}>
+      <div
+        className={
+          mainType ? styles.card_user_info_main : styles.card_user_info_sub
+        }
+      >
+        <div>
+          <img src={userPic} className={styles.user_pic} />
+          <span className={styles.username}>{username}</span>
+          <span> </span>
+          <span> - </span>
+          <span>{timeStamp}</span>
+        </div>
+        <div className={styles.expand_button} onClick={expandClick()}>
+          ...
+        </div>
+      </div>
+      <div className={styles.card_content}>{content}</div>
+      <div className={styles.card_buttons}>
+        <div className={styles.buttons_left}>
+          <div className={styles.reply_button}>
+            <ReplyIcon /> Reply
+          </div>
+          <div className={styles.edit_delete}>
+            <PencilIcon className={styles.pencil} size={28} />
+            <TrashIcon className={styles.trash} size={28} />
+          </div>
+        </div>
+        <div className={styles.likes_container}>
+          <div className={styles.likes_button} onClick={likeClick()}>
+            <ThumbsupIcon className={styles.thumbs_up_down} />{' '}
+            <span className={styles.number_shown}>{likes}</span>
+          </div>
+          <div className={styles.likes_button} onClick={dislikeClick()}>
+            <ThumbsdownIcon className={styles.thumbs_up_down} />{' '}
+            <span className={styles.number_shown}>{dislikes}</span>
+          </div>
+        </div>
+      </div>
+    </div>
+  )
+}
+
+export default DiscussionsCard
--- a/components/DiscussionsCard/DiscussionsCard.tsx
+++ b/components/DiscussionsCard/DiscussionsCard.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import styles from './discussionsCard.module.scss'
-import { ReplyIcon, TrashIcon, PencilIcon } from '@primer/octicons-react'
+import { CommentIcon, TrashIcon, PencilIcon } from '@primer/octicons-react'
 
 export type DiscussionsCardProps = {
   isMain: boolean
@@ -44,7 +44,7 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
       <div className={styles.card_buttons}>
         <div className={styles.buttons_left}>
           <div className={styles.reply_button} onClick={replyClick}>
-            <ReplyIcon /> Reply
+            <CommentIcon /> Reply
           </div>
           <div className={styles.edit_delete}>
             <div onClick={editClick}>
--- a/components/DiscussionsCard/DiscussionsCard.tsx
+++ b/components/DiscussionsCard/DiscussionsCard.tsx
@@ -1,45 +1,35 @@
 import React from 'react'
 import styles from './discussionsCard.module.scss'
-import {
-  ReplyIcon,
-  ThumbsupIcon,
-  ThumbsdownIcon,
-  TrashIcon,
-  PencilIcon
-} from '@primer/octicons-react'
+import { ReplyIcon, TrashIcon, PencilIcon } from '@primer/octicons-react'
 
 export type DiscussionsCardProps = {
-  discussionType: 'main' | string
+  isMain: boolean
   username: string
   userPic: string
   timeStamp: string
   content: string
   likes: number
   dislikes: number
-  expandClick: Function
-  likeClick: Function
-  dislikeClick: Function
+  replyClick: () => void
+  editClick: () => void
+  deleteClick: () => void
 }
 
 const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
-  discussionType,
+  isMain,
   username,
   userPic,
   timeStamp,
   content,
-  likes,
-  dislikes,
-  expandClick,
-  likeClick,
-  dislikeClick
+  replyClick,
+  editClick,
+  deleteClick
 }) => {
-  const mainType = discussionType === 'main'
-
   return (
-    <div className={mainType ? styles.card_body_main : styles.card_body_sub}>
+    <div className={isMain ? styles.card_body_main : styles.card_body_sub}>
       <div
         className={
-          mainType ? styles.card_user_info_main : styles.card_user_info_sub
+          isMain ? styles.card_user_info_main : styles.card_user_info_sub
         }
       >
         <div>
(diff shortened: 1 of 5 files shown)


components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+          <div className={styles.reply_button}>
+            <ReplyIcon /> Reply
+          </div>
+          <div className={styles.edit_delete}>
+            <PencilIcon className={styles.pencil} size={28} />
+            <TrashIcon className={styles.trash} size={28} />
+          </div>
+        </div>
+        <div className={styles.likes_container}>
+          <div className={styles.likes_button} onClick={likeClick()}>
```
bryanjenningz, line 68: Change to `onClick={likeClick}`.

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <ThumbsupIcon className={styles.thumbs_up_down} />{' '}
+            <span className={styles.number_shown}>{likes}</span>
+          </div>
+          <div className={styles.likes_button} onClick={dislikeClick()}>
```
bryanjenningz, line 72: Change to `onClick={dislikeClick}`.

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+import {
+  ReplyIcon,
+  ThumbsupIcon,
+  ThumbsdownIcon,
+  TrashIcon,
+  PencilIcon
+} from '@primer/octicons-react'
+
+export type DiscussionsCardProps = {
+  discussionType: 'main' | string
```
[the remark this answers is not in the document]
anthonykhoa: @HS-90 Is there any potential that there will be more than 2 possible values for `discussionType`? Also, what is the reason for different styling options? Do we have different styling on different pages for DiscussionCard component?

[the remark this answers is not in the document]
anthonykhoa: I'm thinking that if we know there will never be more than 2 possible values for `discussionType`, then we might as well make the prop a boolean value, ex. `subMode?: boolean`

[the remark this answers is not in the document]
anthonykhoa: If theres high potential that discussionType will have more than 2 possible values, id suggest we use a mapping approach instead of using ternary operators(see lines 39 and 42 for example)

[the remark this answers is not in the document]
anthonykhoa: Example of mapping approach: 

```
className={styles[`card_body_${discussionType}]}
```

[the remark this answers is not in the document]
anthonykhoa: The above might be a bad example.. im not sure its a good idea for name of the`class` to be part of the logic in this kind of way.

A more simple approach would probably be to have `main` and `sub` as keys in a hashmap, and then have as its values the corresponding styles for them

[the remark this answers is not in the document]
HS-90: > If theres high potential that discussionType will have more than 2 possible values, id suggest we use a mapping approach instead of using ternary operators(see lines 39 and 42 for example)

There should be only 2 types, one is a main card, which will be the original comment or thread, then replies will be the sub design, branching off of the original comment. 

A subcard can also have subcards branching off of them, kind of like how reddit works. but the design should be the same for every subcard, if that makes sense. I think I agree with your boolean approach

SlyBouhafs:
Preview: https://deploy-preview-2339--youthful-thompson-448332.netlify.app/?path=/story/components-discussionscard--discussions-card-main

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+      >
+        <div>
+          <img src={userPic} className={styles.user_pic} />
+          <span className={styles.username}>{username}</span>
+          <span> </span>
+          <span> - </span>
+          <span>{timeStamp}</span>
+        </div>
+        <div className={styles.expand_button} onClick={expandClick()}>
+          ...
```
SlyBouhafs, line 53: What does the expandClick function do in this case? I remember from the design sessions the idea behind was to use it to edit/delete the comment, but it looks like those two have their own button in this design, should we keep it?

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <ReplyIcon /> Reply
+          </div>
+          <div className={styles.edit_delete}>
+            <PencilIcon className={styles.pencil} size={28} />
+            <TrashIcon className={styles.trash} size={28} />
+          </div>
+        </div>
+        <div className={styles.likes_container}>
+          <div className={styles.likes_button} onClick={likeClick()}>
+            <ThumbsupIcon className={styles.thumbs_up_down} />{' '}
```
SlyBouhafs, line 69: If i remember correctly, during the design meetings we agreed on not having a like/dislike button and instead use arrows and only show a sum of the votes.

In the designs we have two versions and i think we agreed on this one:
Screen Shot 2022-09-24 at 20 17 03

@flacia Can you confirm this?

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: It depends. Do we want to show every response to the thread(including responses to responses) upon load, or do we want the user to click first before it expands?

components/DiscussionsCard/DiscussionsCard.tsx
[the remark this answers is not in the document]
SlyBouhafs: If the only purpose of having a boolean is to check if a reply is the main one, wouldn't be better to instead check if the subreply doesn't have any parent comment and style accordingly?
Also, should we consider that any subreply that has n replies would be considered a main? if so, wouldn't that make things a bit confusing?
We didn't get to this point in the design meetings.

components/DiscussionsCard/DiscussionsCard.tsx
SlyBouhafs: Interesting, i didn't think it would be used for collapsing comments.
I don't think hiding the content of the comment is a good idea, comments should all be visible and if the thread is too long, we can add a button to show more responses instead of making them collapsible.
Maybe we can make the comments who have a negative score collapsed by default.
Also do we have a design on the collapsed state?

components/DiscussionsCard/DiscussionsCard.tsx
SlyBouhafs: We also discussed another idea where we should only have the like button:
Screen Shot 2022-09-24 at 20 43 27

@flacia Can you confirm which one did we choose?

flacial:
Some comments:

image

components/DiscussionsCard/DiscussionsCard.tsx
SlyBouhafs: @flacial Can you confirm which one did we agree on?

flacial: I think without likes. On v2 of the discussions component, we can implement likes then.

components/DiscussionsCard/DiscussionsCard.tsx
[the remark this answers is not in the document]
anthonykhoa: > If the only purpose of having a boolean is to check if a reply is the main one, wouldn't be better to instead check if the subreply doesn't have any parent comment and style accordingly?

Why is it better?

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: > I think without likes. On v2 of the discussions component, we can implement likes then.

Should we not include likes in the current PR? And add the up down arrows in a future PR?

HS-90: Or are you talking about V2 as in, after this initial discussions component has been fully implemented without likes.

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: > Do we have a design on the collapsed state?

Only the discussion cards so far. Haven't implemented the full UI yet. I agree that we can load them all to a certain point. 
Should we have all responses feed to main parent post(the purple design), even if they were responding to a sub post? Or should we have it like reddit, where it continues branching out if you reply to a sub post?

components/DiscussionsCard/DiscussionsCard.tsx
flacial: > Or are you talking about V2 as in, after this initial discussions component has been fully implemented without likes.

Yes, after this initial design.

components/DiscussionsCard/DiscussionsCard.tsx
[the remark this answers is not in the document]
SlyBouhafs: After a bit of though, it seems like it won't make any difference in our case, the only case it would be useful is if we want to style subthread replies with different levels/shades of colors.

components/DiscussionsCard/DiscussionsCard.tsx
SlyBouhafs: > Should we have all responses feed to main parent post(the purple design), even if they were responding to a sub post? Or should we have it like reddit, where it continues branching out if you reply to a sub post?

I think having something similar to reddit is the way to go, we just need to make sure that the design will not break once we have too many branches.

HS-90:
Updated screenshots of designs:
image

image


If everyone is ok with the initial designs of these cards, we can merge and then update functionality in future PRs.

SlyBouhafs (approved):

anthonykhoa:
> Updated screenshots of designs: image
> 
> image
> 
> If everyone is ok with the initial designs of these cards, we can merge and then update functionality in future PRs.

Can we change the image used for `reply`? That image looks like something id see on a `go back` button

HS-90:
> The image used for `reply` looks like something id see on a `go back` button. However, i see you are using the reply icon from react. Any thoughts @bryanjenningz @HS-90 @SlyBouhafs @flacial ? IMO id rather see something like what reddit uses, they use something that looks like the outline of a txt message. I think the image they have next to the reply button is more closely associated with the idea of `replying` than compared to an image of an arrow pointing to the left Screen Shot 2022-09-27 at 3 24 23 PM

I good with switching the icon. Is that also a react icon? Or a different library?

anthonykhoa:
> > The image used for `reply` looks like something id see on a `go back` button. However, i see you are using the reply icon from react. Any thoughts @bryanjenningz @HS-90 @SlyBouhafs @flacial ? IMO id rather see something like what reddit uses, they use something that looks like the outline of a txt message. I think the image they have next to the reply button is more closely associated with the idea of `replying` than compared to an image of an arrow pointing to the left Screen Shot 2022-09-27 at 3 24 23 PM
> 
> I good with switching the icon. Is that also a react icon? Or a different library?

Im not sure about where to find an icon like reddit's to use.
This isn't a blocker though, we can always update this later since this flow isn't public-facing yet. I just wanted to point this out

anthonykhoa (approved):

SlyBouhafs:
> I good with switching the icon. Is that also a react icon? Or a different library?
@HS-90 Yeah, a similar icon is available in Octicons: https://primer.style/octicons/comment-16

HS-90:
https://deploy-preview-2339--youthful-thompson-448332.netlify.app/?path=/story/components-discussionscard--discussions-card-sub

What it looks like with comment icon.

SlyBouhafs:
> https://deploy-preview-2339--youthful-thompson-448332.netlify.app/?path=/story/components-discussionscard--discussions-card-sub
> 
> What it looks like with comment icon.

Looks good! One small detail, i think 10px is too much for the border radius on the hover state, 5px would look a lot better!

HS-90:
https://deploy-preview-2339--youthful-thompson-448332.netlify.app/?path=/story/components-discussionscard--discussions-card-main

PR #2348, opened 2022-09-25
Fix overflowing text between 768px and 991px on DOJO exercise card

bryanjenningz:
Changes:
* The "Show Answer" button text on the DOJO exercise card no longer overflows to the next line on screen widths from 768px to 991px.

How to test:
* Create an exercise at /curriculum/js0/mentor
* Go to /exercises/js0
* Click on "Solve Exercises" button
* Check that the exercise card "Show Answer" button no longer has text overflow to the next line on screen widths from 768px to 991px.

After this change (screen width 770px):
after

Before this change (screen width 770px):
before

### DIFF
--- a/scss/exercises.module.scss
+++ b/scss/exercises.module.scss
@@ -12,7 +12,7 @@
 
 .exercise__container {
   width: 100%;
-  @media (min-width: 768px) {
+  @media (min-width: 992px) {
     width: 75%;
   }
 }


scss/exercises.module.scss
```diff
@@ -12,7 +12,7 @@
 
 .exercise__container {
   width: 100%;
-  @media (min-width: 768px) {
+  @media (min-width: 992px) {
```
bryanjenningz, line 15: This applies width: 75% at the lg breakpoint and above instead of the md breakpoint (see [Bootstrap breakpoints here](https://getbootstrap.com/docs/5.0/layout/breakpoints/))

bryanjenningz: We have to do this in the SCSS file because Bootstrap doesn't have a CSS class for widths at different breakpoints.
This is equivalent to changing `md:w-3/4` to `lg:w-3/4` in Tailwind.

songz (approved):

Issue #2251, opened 2022-09-02
DOJO(student/page): Create student exercises list


flacial:
# Create student exercises list page

This page will list the student's exercises that they answered correctly & incorrectly. It'll have 2 actions:

1. [Start answering questions session](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1405%3A7764)
2. Review wrong answers sessions (similar to the first actions but instead show a customized list of questions)

The two buttons to `ANSWER QUESTIONS` & `REVIEW WRONG ANSWERS` will be disabled until their pages are implemented.
`View discussions` shouldn't work too.

## Exercises

**The exercise card component must be created first.**

Each exercise card has two actions:

1. Add a note to the exercise FOR THE STUDENT ONLY
2. ~~Skip answer~~ it doesn't make sense. Discuss it with engineers

## Exercises list

Pagination should be implemented because the student will solve many exercises and having to render all of these will affect the performance.

Check out [Apollo Pagination](https://www.apollographql.com/docs/react/pagination/overview/)

PR #2370, opened 2022-09-29
Add DOJO exercise submissions to the backend

bryanjenningz:
Changes:
* Adds ability to get and mutate DOJO exercise submissions using GraphQL
* Frontend now gets DOJO exercise submissions from the backend (but it doesn't mutate them on the backend yet, this will be handled in a subsequent pull request)

How to test:
* Create a DOJO module at /admin/lessons/js0/modules or by using GraphQL mutations
* Create DOJO exercises at /curriculum/js0/mentor or by using GraphQL mutations
* Create a exercise submissions by using GraphQL mutations
* Go to the DOJO exercises page at /exercises/js0 and check that the correct exercises are marked as "NOT ANSWERED" or "ANSWERED" depending on how you mutated the exercise submission data
* Notice that if you answer a DOJO exercise at /exercises/js0, it will be marked as "ANSWERED" when you go to the exercise list view, but if you refresh the page, the state will go back to how it was because this pull request doesn't mutate the exercise submissions.

Here are some useful GraphQL queries and mutations for testing (you can run these at /api/graphql):

Add a new module
```graphql
mutation AddModule(
  $lessonId: Int!
  $name: String!
  $content: String!
  $order: Int!
) {
  addModule(
    lessonId: $lessonId
    name: $name
    content: $content
    order: $order
  ) {
    id
  }
}
```
```json
{
  "lessonId": 1,
  "name": "Numbers",
  "content": "We will learn how to use numbers in this lesson.",
  "order": 1
}
```

Add a new exercise to a module
```graphql
mutation AddExercise(
  $moduleId: Int!
  $description: String!
  $answer: String!
  $explanation: String
) {
  addExercise(
    moduleId: $moduleId
    description: $description
    answer: $answer
    explanation: $explanation
  ) {
    id
  }
}
```
```json
{
  "moduleId": 1,
  "description": "```\nlet a = 1\na += 2\n// What is a?\n```",
  "answer": "3",
  "explanation": "`a += 2` is shorthand for `a = a + 2`"
}
```

Add an exercise submission to an exercise
```graphql
mutation AddExerciseSubmission($exerciseId: Int!, $userAnswer: String!) {
  addExerciseSubmission(exerciseId: $exerciseId, userAnswer: $userAnswer) {
    id
    exerciseId
    userId
    userAnswer
  }
}
```
```json
{
  "exerciseId": 1,
  "userAnswer": "3"
}
```

Here is the query that gets run when you go to /exercises/js0
```graphql
query GetExercises {
  lessons {
    title
    docUrl
    slug
  }
  alerts {
    id
    text
    type
    url
    urlCaption
  }
  exercises {
    id
    module {
      name
      lesson {
        slug
      }
    }
    description
    answer
    explanation
  }
  exerciseSubmissions {
    exerciseId
    userAnswer
  }
}
```

This pull request is a part of https://github.com/garageScript/c0d3-app/issues/2251 and https://github.com/garageScript/c0d3-app/issues/2253


graphql/resolvers/exerciseSubmissionCrud.test.js
```diff
@@ -0,0 +1,79 @@
...
+      userAnswer: '123'
+    }
+    prismaMock.exerciseSubmission.findFirst.mockResolvedValue(null)
+    prismaMock.exerciseSubmission.create.mockResolvedValue(
+      mockExerciseSubmission
+    )
+
+    expect(
+      addExerciseSubmission(undefined, mockArgs, mockContext)
+    ).resolves.toEqual(mockExerciseSubmission)
```
bryanjenningz, line 77: I should probably add `expect(prismaMock.exerciseSubmission.findFirst).toBeCalledWith(...)` and `expect(prismaMock.exerciseSubmission.create).toBeCalledWith(...)` but I was running into issues where `prismaMock.exerciseSubmission.create` wasn't being registered as being called for some reason, but `prismaMock.exerciseSubmission.findFirst` gets registered as called. I'm still trying to figure that out. Let me know if you have any ideas so that I can make these tests have checks that the mocked functions got called correctly.

graphql/resolvers/exerciseSubmissionCrud.ts
```diff
@@ -0,0 +1,46 @@
...
+import { Context } from '../../@types/helpers'
+import prisma from '../../prisma'
+
+export const exerciseSubmissions = (
+  _parent: void,
+  _args: void,
+  context: Context
+) => {
+  const userId = context.req.user?.id
+  if (!userId) return []
```
bryanjenningz, line 11: I decided to return an empty array here, but maybe I should throw instead since you need to be logged in to get your own exercise submissions.

graphql/resolvers/exerciseSubmissionCrud.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  return prisma.exerciseSubmission.findMany({
+    include: {
+      exercise: { include: { module: { include: { lesson: true } } } },
```
bryanjenningz, line 15: I am currently not using the module or lesson data for the exercise, all I'm currently using is the exercise's ID, so maybe I can remove this to get a possible performance improvement. Or we can leave it like this if it doesn't make much of a difference and if this keeps things more simple.

graphql/resolvers/exerciseSubmissionCrud.ts
```diff
@@ -0,0 +1,46 @@
...
+      where: { id: exerciseSubmission.id },
+      include: { exercise: true, user: true }
+    })
+  }
+
+  return prisma.exerciseSubmission.create({
+    data: { exerciseId, userId, userAnswer },
+    include: { exercise: true, user: true }
+  })
+}
```
bryanjenningz, line 46: I was thinking that DOJO exercise submissions to exercises should be one-to-one. So if you add an exercise submission and then you update it later, it will just be updated rather than creating a new exercise submission.

To enforce this constraint, I was thinking of making `addExerciseSubmission` update an exercise submission if it already exists rather than having a separate `addExerciseSubmission` and `updateExerciseSubmission`.

Another possible way of enforcing this constraint would be to have separate `addExerciseSubmission` and `updateExerciseSubmission` mutations and throw an error inside `addExerciseSubmission` if an exercise containing the userId and exerciseId already exists.

graphql/typeDefs.ts
```diff
@@ -13,6 +13,7 @@ export default gql`
     submissions(lessonId: Int!): [Submission!]
     alerts: [Alert!]!
     getPreviousSubmissions(challengeId: Int!, userId: Int!): [Submission!]
+    exerciseSubmissions: [ExerciseSubmission!]!
```
bryanjenningz, line 16: This `exerciseSubmissions` query gets all of the exercise submissions for the currently-logged-in user. Maybe a better place to put this would be on the `User` type. This might also be better if we need to grab other users' exercise submissions if we have some feature that requires that.

We might also want to add an `exerciseSubmissions` field to the `Exercise` type in the future as well if we want to add a feature that requires getting all the submissions for an exercise.

graphql/resolvers/exerciseSubmissionCrud.ts
flacial: If it's not used, I think it should be removed. Then It'll be added once needed.

graphql/resolvers/exerciseSubmissionCrud.test.js
flacial: I'll look into it.

graphql/resolvers/exerciseSubmissionCrud.ts
flacial: Returning an empty array make it more pleasant because we won't have to wrap it with try/catch block.

I'd have to look into the benefits of throwing errors in graphql resolvers in order to decide.

graphql/resolvers/exerciseSubmissionCrud.ts
anthonykhoa: Are there any plans to use module or lesson data in this resolver? If we know we need it soon for some other feature, I'd say we can keep the logic since we would need to add it soon anyway. If there's no set plans that require module or lesson data in this resolver, then i'd say remove it

bryanjenningz: Here are all the possible use cases I can think of for exercise submissions:
1. Find all exercise submissions of the currently-logged-in user
2. Find all exercise submissions of another user
3. Find all exercise submissions of an exercise (by many different users)

The only use case I think we have plans for at the moment is #1, which this pull request covers. I can't think of a situation when we would need module or lesson data from the exercise submission.

The problem is that if I were to remove the module and lesson data, then it would create a situation where the GraphQL queries are not type-safe. To make it type-safe, I would have to change the GraphQL typedefs to match the type that I return from the resolver.

The current way we define ExerciseSubmission in the typedef file is:
```graphql
  type ExerciseSubmission {
    id: Int!
    user: User!
    exercise: Exercise!
    userAnswer: String!
  }
```
As you can see, we have to include an entire User and Exercise type from the resolver for this to be type-safe because someone might use those fields in the future and expect them to all be there and they would get a GraphQL validation error if they try to select a field that doesn't get returned.

To fix this, we would have to remove the unused data from the typedef so its type would look more like this:
```graphql
  type ExerciseSubmission {
    id: Int!
    userId: Int!
    exerciseId: Int!
    userAnswer: String!
  }
```
So I think I'll make that change since it seems like we don't need any of the other data from Exercise and User.

prisma/schema.prisma
```diff
@@ -176,9 +177,9 @@ model Module {
 }
 
 model Exercise {
-  id          Int       @id @default(autoincrement())
-  createdAt   DateTime  @default(now()) @db.Timestamptz(6)
-  updatedAt   DateTime  @updatedAt @db.Timestamptz(6)
+  id          Int                  @id @default(autoincrement())
```
flacial, line 180: What caused the change in formatting? The IDE?

prisma/schema.prisma
```diff
@@ -188,9 +189,21 @@ model Exercise {
...
-  module      Module    @relation(fields: [moduleId], references: [id], onDelete: Cascade)
+  author      User                 @relation(fields: [authorId], references: [id], onDelete: Cascade)
+  flaggedBy   User?                @relation("flaggedExercises", fields: [flaggedById], references: [id])
+  module      Module               @relation(fields: [moduleId], references: [id], onDelete: Cascade)
+  submissions ExerciseSubmission[]
 
   @@map("exercises")
 }
+
+model ExerciseSubmission {
```
flacial, line 200: I see some benefits for `submittedAt`. But it's one of the things that is more suitable to be added once needed.

flacial (approved):
Looks good.

prisma/schema.prisma
bryanjenningz: I ran `yarn prisma format` and `yarn prisma migrate dev`

prisma/schema.prisma
bryanjenningz: Yeah, I agree, we can add fields when we need them

graphql/resolvers/exerciseSubmissionCrud.ts
bryanjenningz: Done

anthonykhoa (approved):

graphql/resolvers/exerciseSubmissionCrud.test.js
bryanjenningz: I figured out the problem, I wrote this code:
```js
    expect(
      addExerciseSubmission(undefined, mockArgs, mockContext)
    ).resolves.toEqual(mockExerciseSubmission)
```
But I should have written this
```js
    await expect(
      addExerciseSubmission(undefined, mockArgs, mockContext)
    ).resolves.toEqual(mockExerciseSubmission)
```

bryanjenningz: Done

PR #2373, opened 2022-09-30
Remove any type in DiffView and ReviewCard

bryanjenningz:


components/ReviewCard.tsx
```diff
@@ -60,7 +60,7 @@ const ReviewButtons: React.FC<{
   const [addComment] = useAddCommentMutation()
   const [accept] = useMutation(ACCEPT_SUBMISSION)
   const [reject] = useMutation(REJECT_SUBMISSION)
-  const reviewSubmission = (review: any) => async () => {
+  const reviewSubmission = (review: typeof accept) => async () => {
```
bryanjenningz, line 63: When `reviewSubmission` is called, the `review` argument will be either `accept` or `reject` which both have the same type, so we can simply put `typeof accept` or `typeof reject` here instead of `any`.

flacial (approved):

PR #2375, opened 2022-09-30
Add Edit and Reply UI to DiscussionsCard

HS-90:
**Description:** relates to #2252

This PR adds the Edit and Reply UI for the DiscussionsCard. Please see screenshots below: 

User clicks edit:
edit options

User clicks reply:
reply options

User clicks both reply and edit:
reply and edit options

Test component here: https://deploy-preview-2375--youthful-thompson-448332.netlify.app/?path=/story/components-discussionscard--discussions-card-main


components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -21,41 +41,121 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
...
+            <div className={styles.buttons_left}>
+              <div
+                className={styles.reply_button}
+                onClick={replyClick}
+                data-testid="reply"
+              >
+                <CommentIcon /> Reply
+              </div>
+              <div className={styles.edit_delete}>
+                <div onClick={editClick} data-testid="pencil">
```
bryanjenningz, line 147: You might want to use a button instead of a div for clickable elements.

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -21,41 +41,121 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
...
+      <div className={isMain ? styles.card_body_main : styles.card_body_sub}>
+        <div
+          className={
+            isMain ? styles.card_user_info_main : styles.card_user_info_sub
+          }
+        >
+          <div>
+            <img src={userPic} className={styles.user_pic} />
+            <span className={styles.username}>{username}</span>
+            <span> </span>
```
bryanjenningz, line 122: What's this span for?

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -21,41 +41,121 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
...
+  return (
+    <div className={styles.main_container}>
+      <div className={isMain ? styles.card_body_main : styles.card_body_sub}>
+        <div
+          className={
+            isMain ? styles.card_user_info_main : styles.card_user_info_sub
+          }
+        >
+          <div>
+            <img src={userPic} className={styles.user_pic} />
```
bryanjenningz, line 120: You might want to add `alt=""` here. (https://davidwalsh.name/accessibility-tip-empty-alt-attributes)

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -1,17 +1,37 @@
...
+
+export type DiscussionReply = {
+  //placeholder for DiscussionReply type, subject to change once we finalize structure of backend data
+  username: string
+  userPic: string
+  timeStamp: string
+  content: string
+  likes?: number
+  dislikes?: number
+  replies?: Array<DiscussionReply>
```
bryanjenningz, line 21: Why are the likes, dislikes, and replies optional?

components/DiscussionsCard/DiscussionsCard.tsx
flacial: We should use the `<Image>` component by next.js

components/DiscussionsCard/discussionsCard.module.scss
```diff
@@ -119,3 +119,72 @@
...
+
+.edit_buttons {
+  margin-left: 5px;
+}
+.discard_edit {
+  border-radius: 5px;
+  margin-right: 5px;
+  font-size: 14px;
+  border: 0px;
+  padding: 5px 6px 5px 6px;
```
flacial, line 131: A shorthand for this is `padding: 5px 6px`

components/DiscussionsCard/discussionsCard.module.scss
```diff
@@ -119,3 +119,72 @@
...
+  background-color: transparent;
+  &:hover,
```
flacial, line 133: It's nice to have a transition. 

`transition: background-color 0.3s ease`

components/DiscussionsCard/discussionsCard.test.js
```diff
@@ -0,0 +1,60 @@
...
+  it('Should open an input text editor when user hits pencil/edit', async () => {
+    const { getByTestId, getByText } = render(
+      <MockedProvider>
+        <DiscussionsCard {...mockDiscussionCard} />
+      </MockedProvider>
+    )
+
+    await userEvent.click(getByTestId('pencil'))
+
+    waitFor(() => {
```
flacial, line 48: I don't think you would need `waitFor` if you used `findBy*` instead of `getBy*`.

flacial:
Margin between `mdInput` and `DiscussionCard` would be nice.

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: We decided we aren't implementing likes in this version, but i just kept it in there as optional for now in case we change our minds or decide to put them in our backend data.

For replies, no every post will have replies, so I thought optional was the better way.

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: I used it to add a little space between elements.

anthonykhoa: If you are adding span for the effect setting a specific amount of space between elements, then you need to do that with CSS instead instead of using span elements with whitespace values. It's more appropriate to do it with CSS since CSS is meant to be used for styling. Using CSS will make the styling choice more explicit as well, thus making the code more simple and understandable. From my side, I  was confused about what this span was being used for because it wasn't very clear.

components/DiscussionsCard/DiscussionsCard.tsx
anthonykhoa: It's not a good idea to add those props now. 

1) This PR is about adding edit/reply, not about adding likes/dislikes -- adding likes/dislikes seems complex enough of a task that it should be its own PR. Being more organized with PRs will also make it easier for issue triaging when needed.

2) I also have concerns about not fully completing the `likes/dislikes` feature in this component because I also see no POC for the feature. Since its not completed in this PR(and shouldn't) and there's no POC, it makes me concerned that these props will become tech debt -- when implementing or making a POC for the `likes/dislikes` feature, it could be that you decide `oh, we don't need these like/dislike props anymore, instead I should use prop X`

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -21,41 +41,121 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
...
+                <div onClick={deleteClick}>
+                  <TrashIcon size={28} className={styles.trash} />
+                </div>
+              </div>
             </div>
-          </div>
+          )}
         </div>
       </div>
+      {replyMode ? replyBox : <></>}
```
anthonykhoa, line 158: We can simplify this to {replyMode && replyBox}

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -21,41 +41,121 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
...
+        </div>
+
+        {editMode ? (
+          editBox
+        ) : (
+          <div className={styles.card_content}>{content}</div>
+        )}
+
+        <div className={styles.card_buttons}>
+          {editMode ? (
```
anthonykhoa, line 135: {editMode && (<div>...</div>)}

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -21,41 +41,121 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
...
-              <TrashIcon size={28} className={styles.trash} />
+        </div>
+
+        {editMode ? (
+          editBox
+        ) : (
+          <div className={styles.card_content}>{content}</div>
+        )}
+
+        <div className={styles.card_buttons}>
```
anthonykhoa, line 134: Does this div do anything if editMode is false?

anthonykhoa: If it doesnt do anything, we can simplify logic by removing an instance of checking for `editMode`

{editMode ? editBox : (
  <div className={styles.card_content}>{content}</div>
  <div className={styles.card_buttons}>...</div>
)}

components/DiscussionsCard/DiscussionsCard.tsx
```diff
@@ -21,41 +41,121 @@ const DiscussionsCard: React.FC<DiscussionsCardProps> = ({
...
-          isMain ? styles.card_user_info_main : styles.card_user_info_sub
-        }
-      >
-        <div>
-          <img src={userPic} className={styles.user_pic} />
-          <span className={styles.username}>{username}</span>
+  const [replyMode, setReplyMode] = useState(false)
+  const [editMode, setEditMode] = useState(false)
+  const [replyText, setReplyText] = useState('')
+  const [editText, setEditText] = useState('')
```
anthonykhoa, line 49: Whats the difference between replyText and editText? Why are both kinds of states needed?

components/DiscussionsCard/DiscussionsCard.tsx
anthonykhoa: Similarly, why is the `replies` prop added? It doesn't seem to be used yet either

anthonykhoa: Lets first finalize on the architecture of how to display `replies to replies`(discussion replies) before we add all this code related to it. If this architecture isn't finalized, then it doesn't make sense to add all this logic relating to it because it'll just be tech debt

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: i ended up getting rid of it in my local, and didn't notice any difference visually actually. So it's removed now.

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: > We should use the `<Image>` component by next.js

When I use the Image component from next, for some reason it did not incorporate any of the styling, so I kept it as  for now.

flacial: > > We should use the `<Image>` component by next.js
> 
> When I use the Image component from next, for some reason it did not incorporate any of the styling, so I kept it as `` for now.

Does these help?

- https://nextjs.org/docs/api-reference/next/image#style
- https://www.peterlunch.com/snippets/next-image-styling

components/DiscussionsCard/DiscussionsCard.tsx
HS-90: replyText keeps track of the user inputting the reply, and editText keep track of inputs while a post is being edited by user. Here's a screenshot of both of them being edited at the same time:
image

anthonykhoa:
Can we focus this PR on just implementing the `editing` feature? I see that the `save` ability isn't implemented in this PR, but instead, some logic relating to the `replying` feature is. You finished half of the editing feature(edit/save), and half of the `replying` feature(replying and displays replies) in the same PR. The `edit` is useless without the `save` working, so it makes more sense to implement `save` before `replying`, since `replying` doesn't have anything to do with `edit`

It would be more organized if this PR contains only the editing feature fully finished. And then the next PR can be about `replying` and `showing replies`

HS-90:
> Can we focus this PR on just implementing either the `editing` feature or `replying` feature? You finished half of the editing feature(edit/save), and half of the `replying` feature(replying and displays replies) in the same PR. The `edit` is useless without the `save` working, likewise the `replying` is useless without the ability to see the replies. Its disorganized that you're implementing a bit of each feature in this PR, just focus this PR on one feature

So should I prioritize some of the backend things first? That would be required for the save and reply feature to work. Or should I create dummy objects for now to just save things on the frontend so we at least know what they look like once a user clicks 'save'(edit) or 'send'(replies).

anthonykhoa:
> So should I prioritize some of the backend things first? That would be required for the save and reply feature to work as intended in production. Or should I create dummy objects for now to just save things on the frontend so we at least know what they look like once a user clicks 'save'(edit) or 'send'(replies).

Working on back-end or front-end first is fine, at work it happens in parallel sometimes between teammates -- just be sure there is coordination between front-end changes and back-end changes. Frontend and backend should be following the same e2e architecture. If there's work done in parallel, there has to be design doc in place for the frontend and backend engineers to follow.

Also, is there some design doc in place? If there is none, then I have concerns that you are breaking up this feature of editing&saving into multiple PRs without a finalized architecture in place. If the architecture isn't finalized, then all the code in your PR is just going to be potential tech debt since everything will be subject to change

HS-90:
> > > Also, is there some design doc or POC in place? If there is none, then I have concerns that you are breaking up this feature of editing&saving into multiple PRs without a finalized architecture in place. If the architecture isn't finalized, then all the code in your PR is just going to be potential tech debt since everything will be subject to change.
> > 
> > 
> > There is none right now. I can get one started though so we can get a better idea of the architecture.
> 
> If the feature isn't complex enough you don't have to make a design doc, but just make sure you're organized in your PRs. Instead of implementing a bit of both `edit` and `reply` feature, it would be better to focus this PR on just one thing -- for example if you wanted to focus on just `editing`, then you should have made code changes in this PR to make `save` work instead of some code for `replying`. And then you could have another PR that implements backend changes needed. If the changes needed aren't complex enough, you can have the backend PR and front-end PR up at the same time with no need for design doc. If the feature requires complex changes though, then yeah you need to make a design doc

I think it would be a good idea to create one for this project. I think having one will provide a better sense of direction for myself and the team while working on this.

HS-90:
https://github.com/garageScript/c0d3-app/issues/2400

Here is an initial draft of the design doc. Welcome to any comments or suggestions for improvement!

flacial:
Closing due to inactivity.

PR #2383, opened 2022-10-01
Add DOJO exercise filtering checkbox

bryanjenningz:
Changes:
* Add DOJO exercise filter checkbox which shows which exercise previews get displayed and which exercises you will solve when you press the SOLVE EXERCISES button
* Add INCORRECT state to ExercisePreview component (so now there are 3 states: NOT ANSWERED, INCORRECT, ANSWERED)
* Make it so if you submit an answer with an empty string in the input box, it will make it so the exercise is NOT ANSWERED (this makes it easier to test exercises because you can set ANSWERED or INCORRECT exercises to NOT ANSWERED now)
* Move React state down to the Exercise component
* Only show the SOLVE EXERCISES button if there is at least 1 exercise

How to test:
* Create a DOJO module at /admin/lessons/js0/modules
* Create DOJO exercises at /curriculum/js0/mentor
* Go to /exercises/js0
* Answer exercises correctly, incorrectly, and leave them empty when submitting, then go back and see how the previews look
* Toggle the exercise filtering checkbox to make sure it works correctly

Desktop unchecked:
desktop-unchecked

Desktop checked:
desktop-checked

Mobile unchecked:
mobile-unchecked

Mobile checked:
mobile-checked

This pull request is related to https://github.com/garageScript/c0d3-app/issues/2253 and https://github.com/garageScript/c0d3-app/issues/2251

### DIFF
--- a/pages/exercises/[lessonSlug].tsx
+++ b/pages/exercises/[lessonSlug].tsx
@@ -24,8 +24,8 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
 }) => {
   const { lessons, alerts, exercises, exerciseSubmissions } = queryData
   const router = useRouter()
+  const [solvingExercise, setSolvingExercise] = useState(false)
   const [hideAnswered, setHideAnswered] = useState(false)
-  const [exerciseIndex, setExerciseIndex] = useState(-1)
   const [addExerciseSubmission] = useAddExerciseSubmissionMutation()
   const [userAnswers, setUserAnswers] = useState<Record<number, string>>({})
   useEffect(() => {
@@ -75,29 +75,31 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
       exercise => !hideAnswered || exercise.userAnswer !== exercise.answer
     )
 
-  const exercise = currentExercises[exerciseIndex]
-
   return (
     <Layout title={currentLesson.title}>
-      {exercise ? (
+      {solvingExercise ? (
         <Exercise
-          key={exerciseIndex}
-          exercise={exercise}
-          setExerciseIndex={setExerciseIndex}
           lessonTitle={currentLesson.title}
-          hasPrevious={exerciseIndex > 0}
-          hasNext={exerciseIndex < currentExercises.length - 1}
-          submitUserAnswer={(userAnswer: string) => {
-            setUserAnswers({ ...userAnswers, [exercise.id]: userAnswer })
+          exercises={currentExercises}
+          userAnswers={userAnswers}
+          onExit={localUserAnswers => {
+            setUserAnswers({ ...userAnswers, ...localUserAnswers })
+            setSolvingExercise(false)
+          }}
+          submitUserAnswer={(exerciseId, userAnswer) => {
             addExerciseSubmission({
-              variables: { exerciseId: exercise.id, userAnswer }
+              variables: { exerciseId, userAnswer }
             })
           }}
         />
       ) : (
         <ExerciseList
           tabs={tabs}
-          setExerciseIndex={setExerciseIndex}
+          onClickSolveExercises={() => {
+            if (currentExercises.length > 0) {
+              setSolvingExercise(true)
+            }
+          }}
           lessonTitle={currentLesson.title}
           hideAnswered={hideAnswered}
           setHideAnswered={setHideAnswered}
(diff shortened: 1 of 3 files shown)


pages/exercises/[lessonSlug].tsx
```diff
@@ -70,31 +71,34 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
...
+    .filter(
+      exercise => !hideAnswered || exercise.userAnswer !== exercise.answer
+    )
 
   return (
     <Layout title={currentLesson.title}>
-      {exercise ? (
+      {solvingExercise ? (
         <Exercise
-          key={exerciseIndex}
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:80]: I removed `key` here because we want the `Exercise` component to not rerender since the `Exercise` component now holds more state which we don't want to reset by rerendering.

Another way I could have implemented this would be to keep `userAnswers` and `localUserAnswers` at the same level here and then only pass down `localUserAnswers` to the `Exercise` component.

Both ways work, but I chose to keep `localUserAnswers` at the `Exercise` component level.

pages/exercises/[lessonSlug].tsx
```diff
@@ -70,31 +71,34 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
...
-          exercise={exercise}
-          setExerciseIndex={setExerciseIndex}
           lessonTitle={currentLesson.title}
-          hasPrevious={exerciseIndex > 0}
-          hasNext={exerciseIndex < currentExercises.length - 1}
-          submitUserAnswer={(userAnswer: string) => {
-            setUserAnswers({ ...userAnswers, [exercise.id]: userAnswer })
+          exercises={currentExercises}
+          userAnswers={userAnswers}
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:84]: The `userAnswers` prop here only gets used to initialize the `localUserAnswers` state in the `Exercise` component. Maybe a better name for this prop would be `initUserAnswers` or `initLocalUserAnswers`.

pages/exercises/[lessonSlug].tsx
```diff
@@ -70,31 +71,34 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
...
+          onExit={localUserAnswers => {
+            setUserAnswers({ ...userAnswers, ...localUserAnswers })
+            setSolvingExercise(false)
+          }}
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:88]: `onExit` gets called when we exit the `Exercise` component and go back to the `ExerciseList` view mode. The `userAnswers` gets updated only after we exit the `Exercise` component because if we don't want `userAnswers` changing which would change `currentExercises` which could cause the exercise that gets shown in the `Exercise` component to change in a way we don't want.

pages/exercises/[lessonSlug].tsx
```diff
@@ -70,31 +71,34 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
...
+          submitUserAnswer={(exerciseId, userAnswer) => {
             addExerciseSubmission({
-              variables: { exerciseId: exercise.id, userAnswer }
+              variables: { exerciseId, userAnswer }
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:91]: We no longer update the `userAnswers` locally after each submit. Instead, we update the server exercise submissions and we update the `localUserAnswers` because we don't want updating `userAnswers` to change the `currentExercises` which would cause a rerender in a buggy way when the exercise filtering checkbox is checked (e.g. if we only show incomplete exercises then we complete an exercise, then we would increment the exercise index and the exercises array would be shorter by 1 after that exercise got complete, so to avoid this we don't update `userAnswers` until after the user exits the `Exercise` component).

pages/exercises/[lessonSlug].tsx
```diff
@@ -104,55 +108,71 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
...
+
+  const hasPrevious = exerciseIndex > 0
+  const hasNext = exerciseIndex < exercises.length - 1
 
   return (
     <div className={`mx-auto ${styles.exercise__container}`}>
       <button
         className="btn ps-0 d-flex align-items-center"
-        onClick={() => setExerciseIndex(-1)}
+        onClick={() => onExit(localUserAnswers)}
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:145]: We pass all the `localUserAnswers` so that the parent component can update `userAnswers` to reflect the current state.

pages/exercises/[lessonSlug].tsx
```diff
@@ -104,55 +108,71 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
...
+        }}
       />
       <div className="d-flex justify-content-between mt-4">
         {hasPrevious ? (
           <button
-            onClick={() => setExerciseIndex(i => i - 1)}
+            onClick={() => {
+              setExerciseIndex(i => i - 1)
+              setAnswerShown(false)
+              setMessage(Message.EMPTY)
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:174]: The reason why we now have to set the message to EMPTY and set the answer to be not shown is because we are no longer completely rerendering the `Exercise` component (it never gets completely rerendered now because we removed `key` prop and because we don't update any parent state until after we exit the `Exercise` component).

pages/exercises/[lessonSlug].tsx
```diff
@@ -208,24 +248,39 @@ const ExerciseList = ({
...
         </div>
+        {exercises.length > 0 && (
+          <div
+            className={`mb-3 mb-md-0 d-flex d-md-block ${styles.exerciseList__solveExercisesButtonContainer}`}
+          >
+            <NewButton className="flex-grow-1" onClick={onClickSolveExercises}>
+              SOLVE EXERCISES
+            </NewButton>
+          </div>
+        )}
```
Comment (bryanjenningz) [inline pages/exercises/[lessonSlug].tsx:272]: The SOLVE EXERCISES button only shows if there are exercises left. So if the user completes all the exercises and then checks the exercise filtering checkbox, there will be no exercises left, so this button is not shown.

Maybe we can also add a message in the empty exercise list body like "Congratulations! You finished all the exercises!" and then maybe we can add a button that says "Go to the next lesson" or something similar.

pages/exercises/[lessonSlug].tsx
```diff
@@ -208,24 +248,39 @@ const ExerciseList = ({
...
       </div>
       <div className={styles.exerciseList__container}>
         {exercises.map((exercise, i) => (
           <ExercisePreviewCard
             key={i}
             moduleName={exercise.moduleName}
-            state={exercise.userAnswer === null ? 'NOT ANSWERED' : 'ANSWERED'}
+            state={(() => {
+              if (exercise.userAnswer === exercise.answer) return 'ANSWERED'
```
Comment (flacial) [inline pages/exercises/[lessonSlug].tsx:280]: Instead of setting the state in the client side, is it better to store it in the user's `exerciseSubmission`?

```js
{
  ....,
  state: 'ANSWERED'
}
```

flacial (approved):
Looks good!

pages/exercises/[lessonSlug].tsx
Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx:280]: Yeah, you're right. It's better to do all the data processing in one location. Also, if we need this property in multiple places in the future, it will be easier to grab it.

Comment (bryanjenningz) [reply pages/exercises/[lessonSlug].tsx]: Done.

Issue #2029, opened 2022-06-29
DOJO(mentor/page): List mentor submitted exercises


flacial:
This page will list all the mentor's submitted exercises, with the ability to edit and delete them.

[Here's the page mock](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1046%3A5592)

The page's states:
  - [When a new question is added](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1312%3A6493)
  - [When a question is deleted](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1312%3A6226) 

Here are the components that you'll have to make before building the page:
- [Mentor problem card](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1046%3A5725)
- [Filter by module dropdown](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1046%3A5594) (could be generic)
  - [Here's the filter by module dropdown `onClick` event design](https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=1166%3A5155)

There are a few things missing from the Mentor problem card:
- Top 3 wrong answers
- Correct out of total ratio
- The card design on Edit state

We already created a card that has these things (https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=311%3A2314) but for some reason a new one that doesn't have them was created in the last design meeting. Feel free to discuss it with the engineers or the designer.

Notes:
~~- @SlyBouhafs is working on #1975 which is a subpage for this page.~~

evo777:
This is a good idea.

flacial:
@JasirZaeem Hi, are you working on this?

flacial:
Things to add:

1. Create a query to get all the mentor's created exercises based on the user ID
1. Modify the exercise card to not show ANSWERED or NOT ANSWERED states
1. Add a back button to the `addExercise` page

flacial:
4. Make `NavBar` references DRY for the curriculum pages.

flacial:
5. Add filtering box
6. Add edit/delete functionalities

flacial:
- [ ] Add filtering box
- [ ] Add edit/delete functionalities

PR #2394, opened 2022-10-03
feat(components): Create LessonTabs

flacial:
Related to #2029 

### Description

For each lesson sub-page such as exercises, mentor exercises, challenges, and lesson, we create a the `NavTab` in all of them. To keep it DRY, this PR create a component that'll be used in all of the lesson pages.

Before:

```ts
const tabs = [
  ...(currentLesson.docUrl
    ? [{ text: 'lessons', url: currentLesson.docUrl }]
    : []),
  { text: 'challenges', url: `/curriculum/${currentLesson.slug}` },
  { text: 'exercises', url: `/exercises/${currentLesson.slug}` },
  {
    text: 'mentor exercises',
    url: `/curriculum/${currentLesson.slug}/mentor/`
  }
]

  <NavCard
    tabSelected={tabs.findIndex(tab => tab.text === 'exercises')}
    tabs={tabs}
  />
```

After:

```ts
<LessonTabs lesson={lesson} activeTab={'exercises'} />
```

### Testing

1. Go to `/exercise/js0`
2. Confirm switching between the tabs work

### DIFF
--- a/components/LessonTabs/LessonTabs.tsx
+++ b/components/LessonTabs/LessonTabs.tsx
@@ -1,0 +1,35 @@
+import React from 'react'
+import NavCard from '../NavCard'
+
+type Lesson = {
+  title: string
+  docUrl?: string | null
+  slug: string
+}
+
+const LessonTabs = ({
+  lesson,
+  activeTab
+}: {
+  lesson: Lesson
+  activeTab: string
+}) => {
+  const tabs = [
+    ...(lesson.docUrl ? [{ text: 'lesson', url: lesson.docUrl }] : []),
+    { text: 'challenges', url: `/curriculum/${lesson.slug}` },
+    { text: 'exercises', url: `/exercises/${lesson.slug}` },
+    {
+      text: 'mentor exercises',
+      url: `/curriculum/${lesson.slug}/mentor`
+    }
+  ]
+
+  return (
+    <NavCard
+      tabSelected={tabs.findIndex(tab => tab.text === activeTab)}
+      tabs={tabs}
+    />
+  )
+}
+
+export default LessonTabs
--- a/pages/exercises/[lessonSlug].tsx
+++ b/pages/exercises/[lessonSlug].tsx
@@ -11,7 +11,6 @@ import {
 import Error, { StatusCode } from '../../components/Error'
 import LoadingSpinner from '../../components/LoadingSpinner'
 import AlertsDisplay from '../../components/AlertsDisplay'
-import NavCard from '../../components/NavCard'
 import ExercisePreviewCard, {
   ExercisePreviewCardProps
 } from '../../components/ExercisePreviewCard'
@@ -20,6 +19,13 @@ import ExerciseCard, { Message } from '../../components/ExerciseCard'
 import { ArrowLeftIcon } from '@primer/octicons-react'
 import GET_EXERCISES from '../../graphql/queries/getExercises'
 import styles from '../../scss/exercises.module.scss'
+import LessonTabs from '../../components/LessonTabs'
+
+type ExerciseLesson = {
+  title: string
+  docUrl?: string | null
+  slug: string
+}
 
 const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
   queryData
@@ -50,18 +56,6 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
   const currentLesson = lessons.find(lesson => lesson.slug === slug)
   if (!currentLesson)
     return <Error code={StatusCode.NOT_FOUND} message="Lesson not found" />
-
-  const tabs = [
-    ...(currentLesson.docUrl
-      ? [{ text: 'lessons', url: currentLesson.docUrl }]
-      : []),
-    { text: 'challenges', url: `/curriculum/${currentLesson.slug}` },
-    { text: 'exercises', url: `/exercises/${currentLesson.slug}` },
-    {
-      text: 'mentor exercises',
-      url: `/curriculum/${currentLesson.slug}/mentor/`
-    }
-  ]
 
   const currentExercises = exercises
     .filter(exercise => exercise?.module.lesson.slug === slug)
@@ -103,12 +97,12 @@ const Exercises: React.FC<QueryDataProps<GetExercisesQuery>> = ({
         />
       ) : (
         <ExerciseList
-          tabs={tabs}
           onClickSolveExercises={() => setSolvingExercise(true)}
           lessonTitle={currentLesson.title}
           hideAnswered={hideAnswered}
           setHideAnswered={setHideAnswered}
           exercises={currentExercises}
+          lesson={currentLesson}
         />
       )}
       {alerts && <AlertsDisplay alerts={alerts} />}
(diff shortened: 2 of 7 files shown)


components/LessonTabs/LessonTabs.tsx
```diff
@@ -0,0 +1,35 @@
...
+  lesson: Lesson
+  activeTab: string
+}) => {
+  const tabs = [
+    ...(lesson.docUrl ? [{ text: 'lesson', url: lesson.docUrl }] : []),
+    { text: 'challenges', url: `/curriculum/${lesson.slug}` },
+    { text: 'exercises', url: `/exercises/${lesson.slug}` },
+    {
+      text: 'mentor exercises',
+      url: `/curriculum/${lesson.slug}/mentor`
```
SlyBouhafs, line 23: If the paths are static, can we use the constants [here]?(https://github.com/garageScript/c0d3-app/blob/master/constants/index.ts)

flacial: Yeah, great idea!

components/LessonTabs/LessonTabs.tsx
```diff
@@ -0,0 +1,36 @@
...
+  lesson,
+  activeTab
+}: {
+  lesson: Lesson
+  activeTab: string
+}) => {
+  const tabs = [
+    ...(lesson.docUrl ? [{ text: 'lesson', url: lesson.docUrl }] : []),
+    { text: 'challenges', url: `/curriculum/${lesson.slug}` },
+    { text: 'exercises', url: `/exercises/${lesson.slug}` },
```
SlyBouhafs, line 21: Might as well add this path to the constants.

components/LessonTabs/LessonTabs.tsx
```diff
@@ -0,0 +1,36 @@
...
+const LessonTabs = ({
+  lesson,
+  activeTab
+}: {
+  lesson: Lesson
+  activeTab: string
+}) => {
+  const tabs = [
+    ...(lesson.docUrl ? [{ text: 'lesson', url: lesson.docUrl }] : []),
+    { text: 'challenges', url: `/curriculum/${lesson.slug}` },
```
SlyBouhafs, line 20: This path should be a constant too.

components/LessonTabs/LessonTabs.tsx
flacial: The exercises shouldn't be under `/exercises` but something like `/curriculum/js0/exercises`.

components/LessonTabs/LessonTabs.tsx
```diff
@@ -0,0 +1,36 @@
...
+  docUrl?: string | null
+  slug: string
+}
+
+const LessonTabs = ({
+  lesson,
+  activeTab
+}: {
+  lesson: Lesson
+  activeTab: string
```
HS-90, line 16: should we restrict the activeTab type to only certain inputs like for example `'lesson' | 'exercises' | 'challenge' | 'mentor exercises' `?

components/LessonTabs/LessonTabs.tsx
SlyBouhafs: Yes, i also think the mentor path also needs a bit more thought!

SlyBouhafs (approved):

pages/exercises/[lessonSlug].tsx
```diff
@@ -233,29 +227,26 @@ type ExerciseListItem = {
...
+  lesson
 }: ExerciseListProps) => {
   return (
     <>
       <div className="mb-4">
-        <NavCard
-          tabSelected={tabs.findIndex(tab => tab.text === 'exercises')}
-          tabs={tabs}
-        />
+        <LessonTabs lesson={lesson} activeTab={'exercises'} />
```
Comment (anthonykhoa) [inline pages/exercises/[lessonSlug].tsx:249]: No need for brackets, change to `activeTab="exercises"`

pages/exercises/[lessonSlug].tsx
```diff
@@ -20,6 +19,13 @@ import ExerciseCard, { Message } from '../../components/ExerciseCard'
 import { ArrowLeftIcon } from '@primer/octicons-react'
 import GET_EXERCISES from '../../graphql/queries/getExercises'
 import styles from '../../scss/exercises.module.scss'
+import LessonTabs from '../../components/LessonTabs'
+
+type ExerciseLesson = {
+  title: string
+  docUrl?: string | null
```
Comment (anthonykhoa) [inline pages/exercises/[lessonSlug].tsx:26]: Why is there `null` here as a possible option?

components/LessonTabs/LessonTabs.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+type Lesson = {
+  title: string
+  docUrl?: string | null
+  slug: string
+}
+
+type Props = {
+  lesson: Lesson
+  activeTab: 'lesson' | 'challenges' | 'exercises' | 'mentor exercises'
```
anthonykhoa, line 13: Can we make an enum for these possible values for activeTab?

pages/exercises/[lessonSlug].tsx
Comment (flacial) [reply pages/exercises/[lessonSlug].tsx:26]: [The lesson type for `docUrl` is `string` or `null`](https://github.com/flacial/c0d3-app/blob/0dd1d4dcb09ce7978f98501af1e449bdb56a8792/graphql/typeDefs.ts#L215)

pages/exercises/[lessonSlug].tsx
Comment (anthonykhoa) [reply pages/exercises/[lessonSlug].tsx:249]: Actually it would be better to use an enum here instead of hardcoding "exercises", since this component is to be used as a child in multiple components

Comment (flacial) [reply pages/exercises/[lessonSlug].tsx]: Done.

components/LessonTabs/LessonTabs.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+type Props = {
+  lesson: Lesson
+  activeTab: LessonTab
+}
+
+const LessonTabs = ({ lesson, activeTab }: Props) => {
+  const tabs = [
+    ...(lesson.docUrl ? [{ text: 'lesson', url: lesson.docUrl }] : []),
```
anthonykhoa, line 25: We can simplify this to 

```
...(lesson.docUrl && [{ text: 'lesson', url: lesson.docUrl }])
```

components/LessonTabs/LessonTabs.tsx
anthonykhoa: > The exercises shouldn't be under /exercises but something like /curriculum/js0/exercises.

@flacial Can you explain the reason why the above should mean that we shouldn't make a `/exercises` constant?

components/LessonTabs/LessonTabs.tsx
flacial: When TypeScript sees this, it throws the following:

```
Type '"" | { text: string; url: string; }[] | null | undefined' must have a '[Symbol.iterator]()' method that returns an iterator.
```

It's saying it's possible for `[{ text: 'lesson', url: lesson.docUrl }]` to be falsey and `lesson.docUrl` which has the types `string | undefined | null` to be returned. Type `string`, `undefined`, and `null` aren't iterable; therefore, it throws an error when destructuring whatever the expression returns.

components/LessonTabs/LessonTabs.tsx
anthonykhoa: > I thought /exercises/js0 would make things inconsistent with the way we manage things that are related to a lesson (e.g, challenges) and thus it should be changed to something like /curriculum/js0/exercises. 

I'm not understanding the reasoning -- even if the path is `/curriculum/js0/exercises`, why do you think a `/exercises` constant should still not be made for that?

flacial: > > I thought /exercises/js0 would make things inconsistent with the way we manage things that are related to a lesson (e.g, challenges) and thus it should be changed to something like /curriculum/js0/exercises.
> 
> I'm not understanding the reasoning -- even if the path is `/curriculum/js0/exercises`, why do you think a `/exercises` constant should still not be made for that?

I decided not to add that constant because, based on the possibility of introducing inconsistency in how we manage things that are related to a lesson, we could change the exercises' paths before launching DOJO. I was wrong as we already have the `/review/[lessonSlug]` path.

anthonykhoa: I was thinking that since we are going to need `/exercises` in the path it would be a good idea to make the constant anyway(which I see you note you'll do in another PR). Whether the path is `/curriculum/js0/exercises` or `/exercises/js0` it needs `/exercises`. The reason I think `/exercises` should be a constant is because of its high potential for reusability -- if you search the path `/curriculum` for example, you can find a lot of results of its usage,
 https://github.com/garageScript/c0d3-app/search?p=4&q=%2Fcurriculum

flacial: I like this idea. By creating the exercises constant, we can join constants to create paths like the following `CURRICULUM_PATH/[lessonSlug]/REVIEW_PATh`.

PR #2399, opened 2022-10-04
Allow user to hit enter to submit answer

bryanjenningz:
Changes:
* User can now hit the enter key to submit an answer on the exercise card

Why:
* Reduce friction and allow users to more quickly answer questions

How to test:
* Go to /exercises/js0
* Click "SOLVE EXERCISES" button
* Click the input box then hit the enter key and verify that the answer submits
* Verify that clicking the SUBMIT button still submit the answer too

### DIFF
--- a/components/ExerciseCard/ExerciseCard.tsx
+++ b/components/ExerciseCard/ExerciseCard.tsx
@@ -33,6 +33,16 @@ const ExerciseCard = ({
 }: ExerciseCardProps) => {
   const [studentAnswer, setStudentAnswer] = useState('')
 
+  const submitAnswer = () => {
+    if (studentAnswer.trim() === answer.trim()) {
+      setMessage(Message.SUCCESS)
+      setAnswerShown(true)
+    } else {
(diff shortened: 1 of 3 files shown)


components/ExerciseCard/ExerciseCard.tsx
```diff
@@ -48,6 +58,11 @@ const ExerciseCard = ({
               message === Message.ERROR ? styles.exerciseCard__input__error : ''
             }`}
             value={studentAnswer}
+            onKeyDown={e => {
+              if (e.key === 'Enter') {
+                submitAnswer()
+              }
+            }}
```
bryanjenningz, line 65: Another possible way to do this would be to wrap the exercise card in a form and then selectively choose which buttons cause the form to submit. For example, the "Show/hide answer" button shouldn't cause the form to submit.

flacial:
I think `cmd/alt + enter` is nice to have for the exercise card and submission comments.

Issue #2400, opened 2022-10-04
DOJO(Discussions Component) Design Doc


HS-90:
Related to issue #2252
This relates to C0D3 DOJO Design doc #1192  and includes a more detailed documentation of the implementation of the discussion component.

## Intro
Current UI design for discussions in Figma(unfinalized): https://www.figma.com/file/f9xRyl1bJOjtBRd8zFORzg/c0d3?node-id=404%3A13721

Current Component Design in Codebase: 
https://deploy-preview-2339--youthful-thompson-448332.netlify.app/?path=/story/components-discussionscard--discussions-card-main

## Goals
Students can benefit immensely from communicating with mentors and fellow students while working on lessons, exercises, and challenges. 

## Problem
While students can currently discuss topics on Discord, conversations can drift towards different topics, older messages that have valuable information get buried and become hard to track down over time.

## Solution
Having a discussion forum within C0D3 can serve as a valuable tool to discuss topics, concepts, projects, exercises, etc. It allows for better organization of the lesson/exercise being discussed. It can also serve as a lasting reference for future students who run into similar questions or issues. 

## Sample Discussion Flow
![image](https://user-images.githubusercontent.com/77421872/193900874-0ad3a1d4-e93b-4669-beec-4b275d4a7f3d.png)

## Implementation
A model in Prisma and typeDef in GraphQL called 'Discussion' should be created with the following properties(assumes `id`, `createdAt`, and `updatedAt` are always present):

 ```
 Discussion{
	isMain: boolean
	username: string (name of user who posted)
	authorId: int (foreign key to refer to user who posted)
	exerciseId: int (foreign key to refer to exercise being discussed)
	userPic: string (url to user profile picture)
	content: string (body of post)
	replies: Discussion[]
}

```
In addition, `Discussions: Discussion[]` should be added as a property to the `Exercise` model and typeDefs. This way `Discussions` are linked to a specific exercise in the backend.

### For use with DOJO exercise
On the frontend for DOJO exercises, our `DiscussionsCard` component should accept the following props:
```
type DiscussionsCardProps = {
   isMain: boolean
   username: string
   userPic: string
   timeStamp: string
   content: string
   editAndDeletePermission: boolean
   deleteClick: Function
   saveClick: Function
   sendClick: Function
}
```
If `isMain` is true, the post is considered the original/main post and will be styled like the following:
[...]

PR #2404, opened 2022-10-06
Create Discussions table in Prisma

HS-90:
**Description:** relates to #2400

This PR creates the backend data structure in prisma/PostgreSQL. 

This sets up the relational database between Discussions, Exercises, and Users. This also includes a self relation for the `replies` property.


prisma/migrations/20221006161545_create_discussions_table2/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Discussion" ALTER COLUMN "userPic" DROP NOT NULL;
```
HS-90, line 2: Setting userPic to not null, since not every user may have uploaded a profile picture

prisma/schema.prisma
```diff
@@ -207,3 +209,19 @@ model ExerciseSubmission {
 
   @@map("exerciseSubmissions")
 }
+
+model Discussion {
```
flacial, line 213: I'm a little bit confused by naming this table `Discussion`. Wouldn't `Comment` make more sense?

flacial: I feel `Discussion` refer to the feature whilst a `Comment` refer to the comments in the `Discussion`.

HS-90: image

Open to suggestions about naming it something different but there's already a `Comment` table.

prisma/migrations/20221006163340_create_discussions_table3/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- DropForeignKey
+ALTER TABLE "Discussion" DROP CONSTRAINT "Discussion_authorId_fkey";
+
+-- AddForeignKey
+ALTER TABLE "Discussion" ADD CONSTRAINT "Discussion_authorId_fkey" FOREIGN KEY ("authorId") REFERENCES "users"("id") ON DELETE SET NULL ON UPDATE CASCADE;
```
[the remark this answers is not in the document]
flacial: `SetNull` sounds like a better option because once the user delete their account, we can display the user details as `Deleted account` but show the comment content at the same time.

prisma/schema.prisma
```diff
@@ -207,3 +209,19 @@ model ExerciseSubmission {
...
+  id         Int          @id @default(autoincrement())
+  createdAt  DateTime     @default(now()) @db.Timestamptz(6)
+  updatedAt  DateTime     @updatedAt @db.Timestamptz(6)
+  authorId   Int
+  exerciseId Int
+  userPic    String?
```
flacial, line 219: `userPic` should exist in the `author` object. Can't we get it from that?

anthonykhoa:
Are we certain that this `discussions table model` in the backend will not have to be changed anymore in order to complete the Discussions flow?

HS-90:
> Are we certain that this `discussions table model` in the backend will not have to be changed anymore in order to complete the Discussions flow?

I believe it has everything we would need. 
It has unique identifiers, self relation with the ability to nest responses, relational database with Exercise and User(author).

Although do we want the `replies` to automatically inherit the `exerciseId` from it's parent? Or is that something we would do with a resolver in GraphQL?

prisma/schema.prisma
anthonykhoa: Lets finalize all these details before implementation. We should not have to be touching this model in the backend anymore to finish implementing a MVP for the discussions flow

flacial: > Lets finalize all these details before implementation. After implementing this model, we should not have to be touching this model in the backend anymore to finish implementing a MVP for the discussions flow

What do you mean by "finalize"?

anthonykhoa: If we are following a design doc, all of the architecture should be decided upon before implementation. By finalized details I mean complete design doc. What a complete design doc looks like will differ depending on what feature the design doc was made for. Here are good tips on design docs, it's also in our engineering practices
https://www.industrialempathy.com/posts/design-docs-at-google/

prisma/schema.prisma
```diff
@@ -207,3 +209,19 @@ model ExerciseSubmission {
 
   @@map("exerciseSubmissions")
 }
+
+model ExerciseComments {
```
flacial, line 213: Should be `ExerciseComment`

prisma/schema.prisma
```diff
@@ -156,6 +156,7 @@ model User {
   submissionsReviewed       Submission[]         @relation("userReviewedSubmissions")
   submissions               Submission[]         @relation("userSubmissions")
   userLessons               UserLesson[]
+  exerciseComments          ExerciseComments[]
```
flacial, line 159: Should be `ExerciseComment[]`

flacial (approved):
Looks good to me. Apply the comments before merging.

prisma/schema.prisma
HS-90: should be updated now

prisma/schema.prisma
HS-90: updated

Issue #2345, opened 2022-09-24
Debug the disappearance of the Discord tokens


flacial:
# Discord tokens disappearance

### Problem

When the student creates an account and accesses the curriculum, a modal will show up telling them to connect to Discord. After they connect to discord and use the app until their Discord access token expires, often, the modal shows up again INSTEAD of using the stored Discord refresh token to get a new access token.

### Debugging

I tried to output to the console what the request that's used to get new tokens responds with, and it responds with:

```
{ error: 'invalid_grant' }
```

The error is vague, as you can see. 

Using Postman, I tried to manually send a request after getting a new refresh token to the Discord auth API in order to refresh the token. It responded with the data. Following that request with another, and it returned the same error.

The response made me reach the conclusion that a refresh token can't be used more than a single time. This indicated that in our code, we're sending two requests simultaneously to get a new access token. One of the requests succeeds at getting the new access token, but the other fails because the refresh access token was already consumed.

When the student access the curriculum page under `/curriculum`, three queries are sent: 

1. `session` that returns the user data
2. `getApp` that returns the session, alerts, and lessons (it gets executed before the page reaches the client)
3. Another `getApp` by the [AppNav component](https://github.com/garageScript/c0d3-app/blob/536c611ae9da8f363a56e66ba5531e9234cc8334/components/AppNav/AppNav.tsx#L84)

As we can see, two identical requests are sent: [`session` in the curriculum page](https://github.com/garageScript/c0d3-app/blob/adfca610438164dec24038aaa9d9592643e56317/pages/curriculum.tsx#L99-L103) and the `session` within the `getApp` query by [AppNav](https://github.com/garageScript/c0d3-app/blob/536c611ae9da8f363a56e66ba5531e9234cc8334/components/AppNav/AppNav.tsx#L84).

The main reason for this is that the [`getApp` in the AppNav has the fetch policy as `network-only`](https://github.com/garageScript/c0d3-app/blob/adfca610438164dec24038aaa9d9592643e56317/components/AppNav/AppNav.tsx#L85). When the fetch policy is set to that, it'll send a request to get the data without checking the cache if it has it. This result in sending two requests to get the same data. We changed it to `network-only` in this [PR](https://github.com/garageScript/c0d3-app/pull/1850/files#diff-1e1121fab77c7172754b353a7cba87ed074e4d2624505dc2fb2beb9e5e0dd538R85)

#### First request

[...]

PR #2432, opened 2022-10-17
fix: Discord tokens disappear after connecting to discord

flacial:
Final PR for #2345 

### Description

This PR change the query that the `AppNav` component uses to `getSession` query. To understand the reason of these changes, [please read the detailed issue description](#2345).

### Testing

1. Unlink discord if already connected — from `/profile/admin`
2. Log out from your c0d3 account  
3. Clear local storage
4. Login to your c0d3 account
5. Connect to discord
6. Refresh the page (/curriculum) multiple times 
7. Make sure you're still connected to discord


components/AppNav/AppNav.tsx
```diff
@@ -78,13 +78,10 @@ const NotLoggedInAuthNav = () => (
 )
 
 const AppNav: React.FC<{}> = () => {
-  const [session, setSession] = useState<GetAppQuery['session']>()
+  const [session, setSession] = useState<GetSessionQuery['session']>()
```
HS-90, line 81: Just want to confirm, by changing this line to GetSession instead of GetApp, it removes the issue of sending the same refreshToken twice to discord API?

HS-90 (approved):

components/AppNav/AppNav.tsx
flacial: Changing the fetch policy from `network-only`  to `cache-first` is what fixes the issue you mentioned. When it's `cache-first`, it doesn't send a request to the GraphQL API.

But changing the query from `getApp` to `getSession`, fixes the issue of the user Discord avatar not appearing after they connect to Discord. 

In order for a query (in our case, `getSession`) to get its data from the cache, an identical query must be stored there first.

[In the curriculum page](https://github.com/garageScript/c0d3-app/blob/8076c7a4f22307a36c96fb44d02a7934ac536273/pages/curriculum.tsx#L99-L103), a `getSession` query is issued and stored in the cache. When the `AppNav` component renders, it gets the `getSession` query data that was stored from the curriculum page's `getSession` query.

flacial: By default, the fetch policy is `cache-first` that's why the `fetchPolicy` property is removed.

SlyBouhafs (approved):

PR #2480, opened 2022-11-01
getExerciseComment and getChildComments GraphQL queries

HS-90:
**Description**: relates to https://github.com/garageScript/c0d3-app/issues/2400

This PR adds two GraphQL resolvers for getting exercise comments(filtered by exerciseId), and getting child comments(filtered by parentId). When implemented on the frontend, the `getExerciseComment` will get any main comments, and it's replies. Any nested replies beyond the first reply to main comments will be retrieved with `getChildComments`.

To test:
**First we need to add comments**
1. Go to 'api/graphql'
2. Write 3 mutations as follows:

`mutation AddExerciseComment{ addExerciseComment(exerciseId:1, content:"Cool"){ id exerciseId content } }`

`mutation AddExerciseComment{ addExerciseComment(exerciseId:1, parentId: 1, content:" Really Really COOL!"){ id exerciseId parentId content } }`

`mutation AddExerciseComment{ addExerciseComment(exerciseId:1, parentId: 1, content:"Cool bro"){ id exerciseId parentId content } }`

**Getting Exercise Comments**
3. Write a query to get all main comments with ExerciseId 1:
   `query GetExerciseComments{
  getExerciseComments(exerciseId:1){
    id
    createdAt
    exerciseId
    parentId
    content
    replies{
      id
      createdAt
      content
    }
  }
}
`

It should return something like this:
getExerciseComments query

**Getting Child Comments**
4. Write a query to get all `exerciseComment` with parentId of `1`:

`query GetChildComments{
  getChildComments(parentId:1){
    id
    createdAt
    exerciseId
    parentId
    content
  }
}`

It should return something like this:
image

### DIFF
--- a/graphql/resolvers/exerciseCommentCrud.ts
+++ b/graphql/resolvers/exerciseCommentCrud.ts
@@ -2,6 +2,32 @@ import { MutationAddExerciseCommentArgs } from '..'
 import { Context } from '../../@types/helpers'
 import prisma from '../../prisma'
 
+export const getExerciseComments = async (
+  _parent: void,
+  _args: { exerciseId: number }
+) => {
+  const exerciseId = _args.exerciseId
+  return prisma.exerciseComment.findMany({
+    where: { parentId: null, exerciseId },
+    include: {
+      replies: true
+    }
+  })
+}
+
+export const getChildComments = async (
+  _parent: void,
+  _args: { parentId: number }
+) => {
+  const parentId = _args.parentId
+  return prisma.exerciseComment.findMany({
+    where: { parentId },
+    include: {
+      replies: true
+    }
+  })
+}
+
 export const addExerciseComment = async (
   _parent: void,
   { content, exerciseId, parentId, userPic }: MutationAddExerciseCommentArgs,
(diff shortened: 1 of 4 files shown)


graphql/resolvers/exerciseCommentCrud.ts
```diff
@@ -2,6 +2,32 @@ import { MutationAddExerciseCommentArgs } from '..'
...
 
+export const getExerciseComments = async (
+  _parent: void,
+  _args: { exerciseId: number }
+) => {
+  const exerciseId = _args.exerciseId
+  return prisma.exerciseComment.findMany({
+    where: { parentId: null, exerciseId },
+    include: {
+      replies: true
```
flacial, line 13: This will only include the exercise comment replies and not the replies nested *replies* like

```
// Main exercise comment
{
  replies: [
    {
      // ...props -> It won't have this, right?
      replies: []
    }
    // reply2...etc
  ]
}
```

flacial: I think the answer is no

image

flacial (approved):
Looks good! In a following PR, we can create their query files which will help in creating their hooks.

graphql/resolvers/exerciseCommentCrud.ts
HS-90: Right it will retrieve all the 'main' comments and one nested layer of replies below that, then we use the `getChildComments` query to get the nested replies beyond that level.

PR #2543, opened 2022-11-24
Next-Auth: Replace our auth system with next-auth

flacial:
Closes https://github.com/garageScript/c0d3-app/issues/1935

### Changed files in detail

- `success.tsx`: We were using a set of middlewares to populate the user from. We now use the next-auth [`getSession`](https://next-auth.js.org/getting-started/client#getsession) method to retrieve the session, and handle when a user is already connected to the same Discord account.
- `login.tsx` and `signup.tsx`: We used the Apollo mutations `signup` and `login` to start the flow. This has been changed with the next-auth [`signIn`](https://next-auth.js.org/getting-started/client#signin) method.
- `logoutContainer`: We used the Apollo mutation `logout` to log out the user. We now use [`signOut`](https://next-auth.js.org/getting-started/client#signout) method by next-auth.
- `nextAuth` (session callback): Updated the `session` callback to set the `session.user` as the user from the database. This callback runs every time we call `getSession` or `useSession`. The reason is that the data in the session stays outdated if one of them got updated from the client, such as if the user connected to discord. Created a custom credential provider that'll be used to handle the flow after the user confirm their email on signup flow. 

Fixes the bugs in the initial PR https://github.com/garageScript/c0d3-app/pull/1985:

1. Lack of email verification
2. Multiple c0d3 accounts connected to the same Discord account
3. Exposing sensitive data to the client
4. Outdated session data when getting redirected to a page immediately after login

### Testing

1. Does login work?
2. Does logout work?
3. Does signup work?
4. Does resetting the password work?
5. Does setting the user password after email verification work?
6. Does login with Discord work?
7. Does connecting to Discord work?

### Engineering changes 🔧 

- Use `getSession` and `useSession` to get the user's session.

### Following

Handle the error thrown when a user is already connected to the same Discord account.


graphql/resolvers/session.ts
```diff
@@ -45,7 +45,7 @@ export const session = async (_parent: void, _args: void, context: Context) => {
     discordUserId,
     discordAvatarUrl,
     discordUsername,
-    isConnectedToDiscord: !!user.discordRefreshToken // using this to avoid a second fetch to get Discord username
+    isConnectedToDiscord: !!user.discordId // using this to avoid a second fetch to get Discord username
```
flacial, line 48: It has been changed to `discordId` because with how next-auth work, we can't put sensitive data such as the user Discord refresh token in the session.

songz (approved):

components/LogoutContainer.tsx
```diff
@@ -1,26 +1,18 @@
 import React from 'react'
 import { useRouter } from 'next/router'
-import { useLogoutMutation } from '../graphql'
+import { signOut } from 'next-auth/react'
 
 export const LogoutContainer: React.FC = ({ children }) => {
   const router = useRouter()
-  const [logoutUser] = useLogoutMutation({
-    update(cache) {
-      cache.modify({
```
flacial, line 9: Expose the cache so we can modify without having to execute a query. There's probably a reason for setting the `lessonStatus` to an empty array.

pages/confirm/[token].tsx
```diff
@@ -31,25 +31,37 @@ const ExpiredToken: React.FC = () => (
...
 
-export const ResetPassword: React.FC = () => {
+export const ResetPassword = ({
+  onServerError
+}: {
+  onServerError: () => void
+}) => {
   const router = useRouter()
-  const [changePw, { data, error }] = useMutation(UPDATE_PASSWORD)
+  const [confirmToken, setConfirmToken] = useState({} as SignInResponse)
```
Comment (flacial) [inline pages/confirm/[token].tsx:40]: This should be `useState<SignInResponse>({})`

PR #2587, opened 2022-12-08
feat(component): Create ExerciesReportCard

flacial:
Related to #2251 

### Description

This PR creates the component that'll be used by the student to report an exercise. It has an input and two buttons to either submit the issue or cancel the report. It'll be used in in this [page](c0d3.com/exercises/js0)

We didn't design what happens when the report button is clicked, so I designed that part. Let me know if you have any suggestions.

<details>
<summary>
Images
</summary>
image
image
image

</details>


components/ExerciseReportCard/exerciseReportCard.module.scss
```diff
@@ -0,0 +1,30 @@
...
+
+.container {
+  max-width: 400px;
+  background-color: hsl(0, 0%, 96%);
+  padding: 20px;
+  border: 1px solid hsl(0, 0%, 84%);
+  border-radius: 10px;
+}
+
+.container__reportBtn {
```
SlyBouhafs, line 11: We do have styles for the buttons which kind of overlap, would it be possible to use that and extend it instead?
https://github.com/garageScript/c0d3-app/blob/master/scss/button.module.scss

flacial: Sure. Good idea.

SlyBouhafs (approved):

PR #2600, opened 2022-12-10
student feature: Add the ability to report exercises

flacial:
Related to #2251 

### Description

This PR adds the ability for the students to report exercise while solving them. 

In the exercise solving main page, it'll look like this:

<details>
<summary>
Images
</summary>
image
image
</details>

Once an exercise is flagged, it'll remain in the list, and it'll appear in the page for managing exercises under `/admin/lessons/[lessonId]/exercises`. The admin has the option to either unflag it or remove it.

### Bugs

- It doesn't gracefully handle the thrown error when an exercise is already flagged. Will be handled in a subsequent PR.

### Extra functional requirements for flagging exercises

- Students should know that the exercise they're solving has been flagged.
- If there are multiple reports for the same exercise, it should group them together for easier review.

### DIFF
--- a/__tests__/__snapshots__/storyshots.test.js.snap
+++ b/__tests__/__snapshots__/storyshots.test.js.snap
@@ -6490,9 +6490,7 @@ a = a + 10
       </div>
     </div>
   </div>
-  <div
-    className=""
-  >
+  <div>
     <button
       className="btn newButton container__reportBtn borderless"
       onClick={[Function]}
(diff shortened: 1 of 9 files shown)


SlyBouhafs (approved):

SlyBouhafs:
Wouldn't be better to make it wide, so that it covers all the white space on the right?

flag_exercise

flacial:
@SlyBouhafs I think gray is better because it's more neutral for the thing it does: returning the user to the previous screen. Red/the current color empathize a CTO, which is misleading.

flacial:
> Wouldn't it be better to make it wide so that it covers all the white space on the right?
> 
> flag_exercise

Yeah, good idea.

Issue #2641, opened 2022-12-23
Create the resolvers for updating the user data


flacial:
## Problem

We're in the process of making it possible for [the students to update their info](https://github.com/garageScript/c0d3-app/issues/1629), such as their username or password. Currently, there's no way to do that because in the backend side, the resolvers for updating the data are missing.

## Steps to reproduce

1. Go to the [resolvers](https://github.com/garageScript/c0d3-app/blob/master/graphql/resolvers.ts)
2. Observe that there are no resolvers for updating the user info

## Expected behavior

The backed should have the resolvers for updating the user data info

## Actual behavior

There is currently no way for users to update their info by calling a resolver

## Solution

implement GraphQL resolvers for updating user data (username, firstName, lastName, password) in the database. This will allow us to specify the logic for updating user data on the server, which can then be called by the client to make updates.

## Additional context

- Related to https://github.com/garageScript/c0d3-app/issues/1629

flacial:
Closed in https://github.com/garageScript/c0d3-app/pull/2656

PR #2656, opened 2022-12-27
feat: Create the resolver for updating the user password

flacial:
## Motivation
As part of the work on the [settings page](https://chat.openai.com/chat#1629), we need to add the ability for users to change their password. Currently, we have a backend function for resetting or setting a new password when a user signs up, but we don't have a way to allow users to update their password once they are already signed up. Adding this functionality will allow us to complete the section on the settings page for updating the user password.

## Changes
- Created the GraphQL resolver `updateUserPassword` to handle updating a user's password.
- Created the corresponding query file for `updateUserPassword` so that it can be called from the client.
- Added tests to ensure that the resolver is working as intended.
- Created the type definition for the `updateUserPassword` resolver.

## Resolver logic
The `updateUserPassword` resolver takes three arguments: the current password, the new password, and a re-entered version of the new password.

The resolver first checks that the new password and the re-entered password are the same. If they are not, it throws an error with the message Passwords do not match. If the passwords match, the resolver retrieves the user's password hash from the database and compares it to the current password provided by the user. If the current password is incorrect, it throws an error with the message Current password is invalid.

If all the checks pass, the user's password is updated with the new password.

## Testing results
Here are the results of testing the `updateUserPassword` resolver with different input:

**Executing `updateUserPassword` with an incorrect new password**
```gql
mutation {
  updateUserPassword(newPassword:"cat", currentPassword:"drowssap", newPasswordAgain:"password") {
    success
  }
}
```
Result:
```gql
{
  "errors": [
    {
      "message": "Passwords don't match",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
  "data": {
    "updateUserPassword": null
  }
}
```

**Executing `updateUserPassword` with an incorrect current password (assuming current password is `frenchfries`)**
```gql
mutation {
  updateUserPassword(newPassword:"password", currentPassword:"drowssap", newPasswordAgain:"password") {
    success
  }
}
```
Result:
```gql
{
  "errors": [
    {
      "message": "Current password is invalid",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
  "data": {
    "updateUserPassword": null
  }
}
```

**Executing `updateUserPassword` with correct arguments**
```gql
mutation {
  updateUserPassword(newPassword:"password", currentPassword:"drowssap", newPasswordAgain:"password") {
    success
  }
}
```
Result:
```gql
{
  "data": {
    "updateUserPassword": {
      "success": true
    }
  }
}
```

## Testing instructions
1. Go to `/api/graphql`.
2. Execute the queries from "Testing result" section.
3. Ensure it returns the correct result.

## Related issues
- #2641

### DIFF
--- a/graphql/resolvers/userDataCrud.ts
+++ b/graphql/resolvers/userDataCrud.ts
@@ -48,10 +48,10 @@ export const updateUserPassword = withUserContainer<
     throw new Error("Passwords don't match")
   }
 
-  const userFromDB = await prisma.user.findFirst({ where: { id: user.id } })
+  const userFromDB = await prisma.user.findUnique({ where: { id: user.id } })
 
-  const validLogin = userFromDB!.password
-    ? await bcrypt.compare(currentPassword, userFromDB?.password!)
+  const validLogin = userFromDB
+    ? await bcrypt.compare(currentPassword, userFromDB.password as string)
     : false
 
   if (!validLogin) {
--- a/graphql/resolvers/userDataCrud.ts
+++ b/graphql/resolvers/userDataCrud.ts
@@ -1,7 +1,12 @@
 import { User } from '@prisma/client'
-import { MutationUpdateUserNamesArgs } from '..'
+import {
+  MutationUpdateUserNamesArgs,
+  MutationUpdateUserPasswordArgs,
+  SuccessResponse
+} from '..'
 import { withUserContainer } from '../../containers/withUserContainer'
 import prisma from '../../prisma'
+import bcrypt from 'bcrypt'
 
 export const updateUserNames = withUserContainer<
   Promise<User>,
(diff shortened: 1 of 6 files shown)


graphql/resolvers/userDataCrud.ts
```diff
@@ -31,3 +36,38 @@ export const updateUserNames = withUserContainer<
...
+  MutationUpdateUserPasswordArgs
+>(async (_, args, { req }) => {
+  const user = req.user!
+  const { currentPassword, newPassword, newPasswordAgain } = args
+
+  if (newPassword !== newPasswordAgain) {
+    throw new Error("Passwords don't match")
+  }
+
+  const userFromDB = await prisma.user.findFirst({ where: { id: user.id } })
```
JasirZaeem, line 51: `findUnique` makes more sense here since user id has the unique constraint.

graphql/resolvers/userDataCrud.ts
```diff
@@ -31,3 +36,38 @@ export const updateUserNames = withUserContainer<
...
+
+  const validLogin = userFromDB!.password
+    ? await bcrypt.compare(currentPassword, userFromDB?.password!)
+    : false
```
JasirZaeem, line 55:
```ts
const validLogin = userFromDB
    ? await bcrypt.compare(currentPassword, userFromDB.password!)
    : false
```
I'd suggest writing here like this, makes the flow clearer and reduces extra assertions
or even 
```ts
const validLogin = userFromDB
    ? await bcrypt.compare(currentPassword, userFromDB.password as string)
    : false
```

flacial: I'll update the code. Thanks for the comments!

JasirZaeem (approved):

Issue #2630, opened 2022-12-21
Remove flagged exercises in accordance with design doc


flacial:
### Description

The current implementation for removing flagged exercises differs from the design doc. According to the design doc (https://github.com/garageScript/c0d3-app/issues/1192), a removed exercise should not be visible to students, but it should not be completely removed from the database.

We should update the implementation to match the design doc. This will ensure that the app behaves as intended and meets the requirements outlined in the design doc.

## Steps to reproduce:
1. Remove an exercise with [`deleteExercise`](https://github.com/garageScript/c0d3-app/blob/master/graphql/resolvers/exerciseCrud.ts#L79-L103) as inappropriate
2. Observe that the exercise is removed from the database instead of just being hidden from students

## Expected behavior:
Removed exercises should be hidden from students, but not completely removed from the database

## Actual behavior:
Removed exercises are completely removed from the database

### Solution
Update the implementation for removing flagged exercises to match the design doc. This will involve modifying the code so that flagged exercises are hidden from students, but not completely removed from the database.

### Plan
Create a resolver that specifically applies the solution, create the fields bool named `removed`, user named `removedBy`, and date named `removedAt` in the Exercise schema.

flacial:
Closed in #2668

PR #2657, opened 2022-12-29
feat(apollo): Create removeExercise resolver

flacial:
## Motivation
Currently, there is no way to set the removed flag on an exercise when an admin wants to remove it. The removed flag is a field in the database that indicates whether an exercise has been removed or not.

The PR is adding a resolver that will allow an admin to set the removed flag to true when they want to remove an exercise. This will mark the exercise as removed in the database, and it will no longer be visible to users.

## Steps taken
1. Created the resolver `removeExercise`
2. Ensure everything is working as intended by adding tests for the resolver
3. Created the query file for `removeExercise` resolver so it can be easily used in the client-side
4. Update GraphQL `index.ts` with the new resolver and hooks.

### Unrelated steps
- Fixes some typos

## Resolver logic
1. It receives one argument, the user `id`.
2. It fetches the exercise to be removed from the database using the provided `id`.
3. It retrieves the `id` of the user making the request (`authorId`) from the request context (req).
4. If the user is not an admin and the `authorId` does not match the `authorId` of the exercise being removed, the resolver throws an error, indicating that the user is not authorized to remove the exercise.
6. If the exercise is already removed, the resolver throws an error, indicating that exercise is already flagged.
7. If the user is authorized to remove the exercise, the resolver updates the exercise in the database, setting the `removedAt` and `removedById` fields to new Date & the user ID. It also includes the author and module relations in the update.
8. The resolver returns the updated exercise.

## Testing results

Executes `removeExercise`:
```gql
mutation {
  removeExercise(id: 1) {
    id
  }
}
```

**Result:**
```gql
{
  "data": {
    "removeExercise": {
      "id": 1,
      "removedAt": "1672591865633",
      "removedById": 1
    }
  }
}
```

**Result if it's already removed:**
```gql
{
  "errors": [
    {
      "message": "Exercise is already removed",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
  "data": null
}
```

**Result if the exercise is not found:**
```gql
{
  "errors": [
    {
      "message": "Exercise is not found",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
  ],
  "data": null
}
```

## Testing instructions
1. Go to `/api/graphql`
2. Run the queries from `Testing result` section
3. Make sure they return the correct results

## Related issues
- #2630

### DIFF
--- a/graphql/resolvers/exerciseCrud.ts
+++ b/graphql/resolvers/exerciseCrud.ts
@@ -133,7 +133,7 @@ export const removeExercise = withUserContainer<
   return prisma.exercise.update({
     where: { id },
     data: {
-      removedAt: new Date().toISOString(),
+      removedAt: new Date(),
       removedById: authorId as number
     },
     include: {
--- a/graphql/resolvers/exerciseCrud.ts
+++ b/graphql/resolvers/exerciseCrud.ts
@@ -120,8 +120,9 @@ export const removeExercise = withUserContainer<
 
   const authorId = req.user?.id
 
-  if (!isAdmin(req) && exercise?.authorId !== authorId)
+  if (!isAdmin(req) && exercise?.authorId !== authorId) {
     throw new Error('Not authorized to remove')
+  }
 
   if (exercise?.removed) {
     throw new Error('Exercise is already removed')
--- a/graphql/resolvers/exerciseCrud.ts
+++ b/graphql/resolvers/exerciseCrud.ts
@@ -77,7 +77,6 @@ export const updateExercise = withUserContainer<
   })
 })
 
-// will be removed in a following PR
 export const deleteExercise = withUserContainer<
   Promise<Exercise>,
   MutationDeleteExerciseArgs
@@ -104,7 +103,6 @@ export const deleteExercise = withUserContainer<
   })
 })
 
-// replaces deleteExercise
 export const removeExercise = withUserContainer<
   Promise<Exercise>,
   MutationRemoveExerciseArgs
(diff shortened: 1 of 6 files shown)


graphql/resolvers/exerciseCrud.ts
```diff
@@ -102,6 +104,41 @@ export const deleteExercise = withUserContainer<
...
+
+  const exercise = await prisma.exercise.findUnique({
+    where: {
+      id
+    }
+  })
+
+  const authorId = req.user?.id
+
+  if (!isAdmin(req) && exercise?.authorId !== authorId)
```
JasirZaeem, line 123: This if has no curly braces, the next one does, both have single line statements. This is just a style suggestion but can you make them uniform (preferably add braces to both)

JasirZaeem:
This is out of the scope of this particular PR but related, maybe removed should be removedAt/removed_at as well like some other booleans we have changed to date | null before.
Like flagged at and the user who flagged it, it would make sense for removed to be like this too with removed at and removed by data.
And even if removed remains a boolean it should not be an optional field, it should either be required with false as default so only two values are possible, or it should be a date field with default value being date. Currently removed can hold three values null, false, true.

flacial:
> This is out of the scope of this particular PR but related, maybe removed should be removedAt/removed_at as well like some other booleans we have changed to date | null before. Like flagged at and the user who flagged it, it would make sense for removed to be like this too with removed at and removed by data. And even if removed remains a boolean it should not be an optional field, it should either be required with false as default so only two values are possible, or it should be a date field with default value being date. Currently removed can hold three values null, false, true. My suggestion would be to refactor the schema for exercise and improve it with regards to remove before this PR.

I agree. It makes sense.

graphql/resolvers/exerciseCrud.ts
```diff
@@ -102,6 +103,46 @@ export const deleteExercise = withUserContainer<
...
+  }
+
+  if (exercise.removedAt) {
+    throw new Error('Exercise is already removed')
+  }
+
+  return prisma.exercise.update({
+    where: { id },
+    data: {
+      removedAt: new Date().toISOString(),
```
JasirZaeem, line 136: You don't need to convert to isostring, just passing a javascript date object is enough prisma will serialise it internally
e.g.
https://github.com/garageScript/c0d3-app/blob/cd8d8a595098ff376f734328fd08623d3c7b12a3/helpers/updateSubmission.ts#L108-L119

JasirZaeem (approved):

PR #2660, opened 2023-01-01
Remove `removed`, adds `removedAt`, `removedBy`, and `removedAt` fields to Exercise model

flacial:
## Motivation
In a [recent comment](https://github.com/garageScript/c0d3-app/pull/2657#issuecomment-1368216145), it was addressed that the fields for handling removing an exercise could be improved, and this PR reflects the suggested changes.

## Changes
This PR updates the Exercise model to improve how the removed field is handled.

**The changes include:**
1. Renaming `removed` to `removedAt` to match the naming conventions of other date fields in the model.
2. Making `removedAt` a required field with a default value of null. This ensures that there are only two possible values for the field: null or a date.
3. Adding a new field called `removedBy` to track the user who removed the exercise.
4. Removing the `removed` field.

These changes will make it easier to track when an exercise is removed and who removed it, and will improve the overall consistency of the model.

**Related issues**
- #2630

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -149,7 +149,8 @@ model User {
   comments                  Comment[]
   exercises                 Exercise[]
   exerciseSubmissions       ExerciseSubmission[]
-  Exercise                  Exercise[]           @relation("flaggedExercises")
+  FlaggedExercise           Exercise[]           @relation("flaggedExercises")
+  RemovedExercise           Exercise[]           @relation("removedExercises")
   modules                   Module[]
   starsMentor               Star[]               @relation("starMentor")
   starsGiven                Star[]               @relation("starStudent")
@@ -187,7 +188,9 @@ model Exercise {
   answer           String
   testStr          String?
   explanation      String?
-  removed          Boolean?             @default(false)
+  removedAt        DateTime?
+  removedById      Int?
+  removedBy        User?                @relation("removedExercises", fields: [removedById], references: [id])
   flagReason       String?
   flaggedAt        DateTime?
   flaggedById      Int?
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -149,8 +149,7 @@ model User {
   comments                  Comment[]
   exercises                 Exercise[]
   exerciseSubmissions       ExerciseSubmission[]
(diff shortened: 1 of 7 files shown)


prisma/schema.prisma
```diff
@@ -187,7 +188,9 @@ model Exercise {
   answer           String
   testStr          String?
   explanation      String?
-  removed          Boolean?             @default(false)
+  removedAt        DateTime?
```
flacial, line 191: It's only possible to set `removedAt` default value to a `DateTime`, and in the client, we will use it like the following:

```
if (exercise.removedAt) // do something
```

If it has a default, value, the above condition will be true, and we'd get an opposite behavior for what we want.

prisma/schema.prisma
```diff
@@ -187,7 +188,9 @@ model Exercise {
...
+  removedById      Int?
+  removedBy        User?                @relation("removedExercises", fields: [removedById], references: [id])
```
flacial, line 193: Something I discovered during testing the changes with Apollo playground: `removedBy` and `flaggedBy` returns null because they refer to a different model. In order to include them in the result of querying the db for exercises, we'd have to explicitly tell Prisma to include them:

```diff
export const exercises = () => {
  return prisma.exercise.findMany({
    include: {
      author: true,
+     flaggedBy: true,
+     removedBy: true,
      module: {
        include: { lesson: true }
      }
    }
  })
}
```

prisma/schema.prisma
JasirZaeem: If you don't set a default value postgresql will set it to `null` so we don't need to explicitly set the default to null in prisma,  postgresql will do it and we'll get the desired behaviour. Being limited to datetime as a default value is due to prisma, prisma doesn't let you set null as the default for an optional value but that is the default behaviour.
Just adding `removedAt DateTime?` like you have done is enough.

prisma/schema.prisma
JasirZaeem: Yeah, getting the related fields requires a sql JOIN so prisma will not do the extra work unless needed and requested by us.

JasirZaeem (approved):

PR #2663, opened 2023-01-03
feat(discord): Refactor rendering conditions and handle already used account error

flacial:
## Problem
When a user tries to connect their C0D3 account with a Discord account and a different Discord account is already connected, the `error` parameter with the value `connected` is added to the path `/discord/success`. While this indicates that there is an issue, it does not provide a clear explanation for the user.

To improve the user experience and make it easier for them to understand and communicate this issue to the engineers if needed, it would be helpful to provide a clear error message explaining why the user encountered this issue and what actions they can take to resolve it.

## Steps taken
1. Display a message that explains what went wrong
2. Add a test case to ensure code coverage
3. Refactored the conditions to only use `errorCode` to maintaining the code easier.

## Screenshot
Some screenshots of the before/after changes

### Before
image

### After
image

## Testing
1. Connect the `admin` account with Discord
2. Log out
4. Login with `newbie` account
5. Try to connect to Discord
6. Notice how it shows the message in the `Screenshot` section


JasirZaeem:
Do we not have a way to send an error code and and short message from the backend for this page ? Then on the frontend we could do something like a switch that can check if code is 'NOT_LOGG..' it can show the relevant detailed message and component for it. If it is 'ACCOUNT_ALRE..' then it says that and so on.
Right now it seems there are special cases for different errors checking a different combination of parameters to decide what the error is, it will be hard to understand, maintain, and add more error codes in the future with this approach. We can see what other team members think of it.

flacial:
> Right now it seems there are special cases for different errors checking a different combination of parameters to decide what the error is, it will be hard to understand, maintain, and add more error codes in the future with this approach. We can see what other team members think of it.

@JasirZaeem I didn't totally get this part. Are you talking about the conditions inside the `getServerSideProps` or this PR's code?

JasirZaeem:
I was talking about the `DiscordErrorPage` component mainly, it checks for username, or error.message, or alreadyConnected. It would be better if it followed a uniform way of using an error code, e.g. 
```
if(ErrorCode === USER_NOT_LOGGED_IN) {...}
if (...){...}
```
 and so on

flacial:
> I was talking about the `DiscordErrorPage` component mainly, it checks for username, or error.message, or alreadyConnected. It would be better if it followed a uniform way of using an error code, e.g.
> 
> ```
> if(ErrorCode === USER_NOT_LOGGED_IN) {...}
> if (...) {...}
> ```
> 
> and so on, or a switch statement if errorcode is an enum typescript will help in accounting for all values of the enum.

I now understand what you mean. I'll try refactoring the code.

JasirZaeem (approved):

Issue #1629, opened 2022-03-21
Add the ability to modify the user data in profile page


flacial:
The users should be able to:
- Change their name
- Change their username
- Change their password
- Change their profile picture (link to issue)
- Change their email?

evo777:
There are many people who wish to change their C0d3 username.

flacial:
We can't protect the page yet because [there's an issue](https://github.com/nextauthjs/next-auth/issues/4234) that makes any redirects to the login page go into an infinite redirects loop.

PR #2756, opened 2023-02-13
feat: Add next-auth middleware to protect pages

flacial:
## Motivation

At present, we do not have a streamlined approach to prevent unauthorized access to pages that require the user to be logged in. The current implementation involves manual checks on the session to determine if the user needs to be redirected to the login page or not.

## Solution

Introduce `next-auth` middleware to handle the protection of pages that require the user to be logged in. The middleware will maintain a list of protected paths and in case the user is not logged in, they will be immediately redirected to the `/login` page. This not only simplifies the codebase by removing the redundant code for page protection but also saves bandwidth for the user.

## Changes

- Implement `next-auth` middleware to protect the `/settings/account` page
- Create a constant for the path `/settings/account`
- Set the redirection to `/login` for unauthenticated users
- [Remove the `secret` option in favor of using `NEXTAUTH_SECRET` in Vercel's project environment variables](https://next-auth.js.org/configuration/options#nextauth_secret)
- Add a test file to cover the `middleware.ts` file

## Testing

1. Log out of the application
2. Access the `/settings/account` page and verify if it redirects to `/login`
3. Log in to the application by filling in the login form
4. Verify if it redirects to `/settings/account` after successful login

## Additional Info

In future pull requests, more pages such as `/review/[lessonSlug]` will be added to the list of protected pages.

## Related issues

- #1629


SlyBouhafs:
I noticed there are some hardcoded paths in the new changes, maybe consider using the constants instead?

__tests__/pages/api/auth/[...nextauth].test.js
```diff
@@ -26,7 +26,9 @@ describe('next-auth middleware', () => {
         jwt,
         session
       },
-      secret: process.env.SESSION_SECRET
+      pages: {
+        signIn: '/login'
```
Comment (SlyBouhafs) [inline __tests__/pages/api/auth/[...nextauth].test.js:30]: `LOGIN_PATH` here?

middleware.ts
```diff
@@ -0,0 +1,3 @@
+export { default } from 'next-auth/middleware'
+
+export const config = { matcher: ['/settings/account'] }
```
SlyBouhafs, line 3: Use the new `SETTINGS_ACCOUNT_PATH` constant here?

pages/api/auth/[...nextauth].ts
```diff
@@ -15,5 +15,7 @@ export default (
       jwt,
       session
     },
-    secret: process.env.SESSION_SECRET
+    pages: {
+      signIn: '/login'
```
Comment (SlyBouhafs) [inline pages/api/auth/[...nextauth].ts:19]: Same thing here, `LOGIN_PATH`?

pages/login.tsx
```diff
@@ -153,8 +153,11 @@ const LoginPage: React.FC<GetAppProps> & WithLayout = ({
 
     if (ok && !error) {
       window.localStorage.setItem('loggedIn', 'true')
-      const { next } = router.query
-      router.push(next ? (next as string) : '/curriculum')
+      const { next, callbackUrl } = router.query
+      const path = next || callbackUrl
+      const redirectTo = path || '/curriculum'
```
SlyBouhafs, line 158: We can use `CURRICULUM` here?

flacial:
@SlyBouhafs Thanks for reviewing the PR :)

I fixed most of the issues.

SlyBouhafs (approved):

Issue #2624, opened 2022-12-20
Add reusable loading skeleton components


flacial:
### Description

Currently, when a user open https://c0d3.com, some parts of the app are empty because the content that's supposed to show is still not available.

#### Why do we need Loading Skeletons?

1. They improve the user experience by providing a visual indication that something is happening, rather than just showing empty areas. This can help reduce user frustration and improve perceived performance.
2. They improve the overall aesthetic of the UI by filling in empty space and creating a more cohesive design.
3. They improve the user's perception of performance by reducing the time it takes for content to load.

### Resolving the issue

Come up with a design for the loading skeletons and share a POC with the engineers.

PR #2818, opened 2023-03-10
feat: Redirect to /login if user is unauthenticated

flacial:
## Overview

We failed [in the last attempt to add a streamlined solution for protecting pages](https://github.com/garageScript/c0d3-app/pull/2756) because of a [bug with the next-auth package](https://github.com/nextauthjs/next-auth/issues/4234). This PR creates a hook that redirects the user to the `login` page if they're not logged in, and updates the `/settings/account` and `/review/[lessonSlug]` pages. In the future, when next-auth fixes its bug, we'll replace this temporary solution to protect the page with a [next.js middleware](https://next-auth.js.org/tutorials/securing-pages-and-api-routes#nextjs-middleware). ~~After this PR is merged, we'll release the page and share it with the students.~~

## Changes

- Create `useRedirectionUnauthenticated` hook that takes a value and redirects the user to the `/login` page if the value is true (by @JasirZaeem)
- Add a test file for `useRedirectionUnauthenticated`
- Update the `/settings/account` page test file
- Reference the new hook in both the `/settings/account` and `/review/[lessonSlug]` pages
- Create a constant for the session status and update the session status references in the declaration types file `auth` with it

## Testing

0. Logout if you're not already
1. Go to `/settings/account`
2. Notice that it shows a loading spinner and then immediately redirects you to the `/login` page
3. Login
4. Go to `/settings/account`
5. Notice it displays the page without redirecting you to `/login`

### Related issues

- #1629


SlyBouhafs (approved):

pages/settings/account/index.tsx
```diff
@@ -271,7 +272,15 @@ const AccountSettings = () => {
   const loading = userInfoLoading || status === 'loading'
   const { username, name, discordUsername } = data?.userInfo?.user || {}
 
-  if (loading || !data) {
+  // prevents the page UI from showing if unauthenticated
+  if (loading || status === 'unauthenticated') {
+    if (status === 'unauthenticated') {
```
anthonykhoa, line 277: Is the `unauthenticated` status something that other pages will use as well? If so, then it should be made a constant

anthonykhoa: You repeat this logic twice, put it into a variable:

```
status === 'unauthenticated'
```

anthonykhoa: Actually, I notice you're only saving one line of code(`return <LoadingSpinner />`), but the result is that it makes the code more complex and harder to maintain since you are coupling logic together, and also using nested if statements as well. Nested ifs increase the number of linearly independent paths we have, which makes the codebase harder to understand and maintain

anthonykhoa: Lets decouple the logic and have something like this:

```

if (loading) {
    return <LoadingSpinner />
}
if (status === 'unauthenticated') {
  router.push({
    pathname: '/login',
    query: { next: router.asPath }
  })
  return <LoadingSpinner /> 
}
```

anthonykhoa (changes requested):

pages/settings/account/index.tsx
anthonykhoa: Also, if other places in the codebase are using loadingSpinner when redirecting a user, then a function should be made to reproduce this functionality of redirecting + showing spinner

anthonykhoa: What does showing a spinner do when redirecting a user? How long does a user see the spinner for when they get redirected?

flacial: > What does showing a spinner do when redirecting a user? Does the user see the spinner for when they are getting redirected?

It prevents the UI (forms, buttons,..etc) from showing if the user isn't yet determined to be unauthenticated or not. If you tried to go to your profile and refresh the page multiple times, you'll notice the loading spinner disappears and the UI shows up with dummy data because it didn't yet fetch your data from the GraphQL API. This is what that line is trying to achieve, display a loading spinner until all the data that's necessary for the page is fetched.

This apparently isn't the best approach, so we're looking into [this feature](#2624).

JasirZaeem:
We can extract this into a hook since it is being used in at least two places right now, and maybe needed in other pages to be added later. Since this is supposed to be temporary anyway until next-auth fixes their bug it'll be easier to rip it out later.

For example:
```ts
import { useEffect } from "react";
import { useRouter } from "next/router";

/**
 * Redirects to login page if user is not authenticated.
 * With next query param to redirect back to this page after login.
 * @param {boolean} shouldRedirect Condition for redirection, will redirect if this is true.
 */
export const useRedirectUnauthenticated = (shouldRedirect: boolean) => {
  const router = useRouter();
  useEffect(() => {
    if (shouldRedirect) {
      router.push({
        pathname: "/login",
        query: { next: router.asPath },
      });
    }
  }, [shouldRedirect, router]);
}
```

It is used with different ways to decide is the user is logged in in different places, so I've used a boolean as the param. Otherwise this hook can be written to take no params and directly use the session hook from next auth as well.

Usage:
 - In `pages/settings/account/index.tsx`: `useRedirectUnauthenticated(!loading && status === 'unauthenticated')` or `useRedirectUnauthenticated(status === 'unauthenticated')` looking at this PR it seems to me these are equivalent.
 - In `pages/review/[lesson].tsx`: `useRedirectUnauthenticated(!session?.user)`

constants/auth-constants.ts
```diff
@@ -0,0 +1,5 @@
+export enum SessionStatus {
+  Authenticated = 'unauthenticated',
```
anthonykhoa, line 2: looks like a typo here

pages/settings/account/index.tsx
```diff
@@ -268,10 +270,14 @@ const AccountSettings = () => {
...
-  const loading = userInfoLoading || status === 'loading'
+  const loading = userInfoLoading || status === SessionStatus.Loading
+  const unauthenticated = status === SessionStatus.unauthenticated
   const { username, name, discordUsername } = data?.userInfo?.user || {}
 
-  if (loading || !data) {
+  useRedirectUnauthenticated(unauthenticated)
+
+  // prevents the page UI from showing if unauthenticated
+  if (loading || unauthenticated) {
```
anthonykhoa, line 280: There is a withqueryloader container , we dont have to do this loading logic ourselves
https://github.com/garageScript/c0d3-app/blob/master/containers/withQueryLoader.tsx

anthonykhoa: Looks like it also accounts for error scenarios already too

anthonykhoa: That could be cleaned up in another PR tho

pages/settings/account/index.tsx
```diff
@@ -268,10 +270,14 @@ const AccountSettings = () => {
     }
   })
 
-  const loading = userInfoLoading || status === 'loading'
+  const loading = userInfoLoading || status === SessionStatus.Loading
+  const unauthenticated = status === SessionStatus.unauthenticated
   const { username, name, discordUsername } = data?.userInfo?.user || {}
 
-  if (loading || !data) {
```
anthonykhoa, line 274: what happens if no data is returned from the userinfoquery? or if there is an error in making the request?

flacial: If there's no data it means the user is unauthenticated and they'll be redirected to `/login` as we discussed above. It doesn't handle errors. We can make it handle errors in a separate PR.

pages/settings/account/index.tsx
flacial: Great idea, I'll look into it.

constants/auth-constants.ts
anthonykhoa: for Authenticated it looks like its value shouldn't be `unauthenticated`

pages/settings/account/index.tsx
anthonykhoa: My concern is that in your PR description it says `After this PR is merged, we'll release the page and share it with the students`. If features are getting released into production, error scenarios need to be accounted for as a requirement. We cannot have an incomplete experience for any scenario, even during error scenarios. All edge cases should be taken care of.

If we can merge the changes in this PR without releasing the changes to production, then it is okay to make a followup PR to account for error scenarios. Otherwise, they need to be addressed in this PR.

helpers/useRedirectUnauthenticated.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Redirects to login page if user is not authenticated.
+ * With next query param to redirect back to this page after login.
+ * @param {boolean} shouldRedirect Condition for redirection
+ */
+const useRedirectUnauthenticated = (shouldRedirect: boolean) => {
+  const router = useRouter()
+
+  useEffect(() => {
```
anthonykhoa, line 13: I think this should work without useEffect. Can you try removing useEffect and try to have it work without it? Because in the files that you refactored, useEffect is not used to handle this redirection logic for users who are not logged in. So it seems like we may be adding unnecessary logic by using useEffect here

constants/auth-constants.ts
flacial: Oh, I see it now. Thanks!

pages/settings/account/index.tsx
flacial: You got a point. We'll just make this PR about creating the hook, and after resolving the issues mentioned in this PR, we can then release it.

helpers/useRedirectUnauthenticated.ts
JasirZaeem: It'll work without useEffect, I used useEffect in my suggestion to not do it on every render which is what happens now.

anthonykhoa: For this use case, useEffect is not a good choice.

Using useEffect for this use case adds an additional layer of complexity and makes it harder to maintain because components will be forced to do a re-render in order for the redirection logic to get run. There should be no dependency on needing to re-render components when redirecting users, it doesn’t make sense. Its an additional layer of complexity that is unnecessary.

pages/settings/account/index.tsx
```diff
@@ -268,10 +270,14 @@ const AccountSettings = () => {
     }
   })
 
-  const loading = userInfoLoading || status === 'loading'
+  const loading = userInfoLoading || status === SessionStatus.Loading
```
anthonykhoa, line 273: Hmmm we shouldnt be making a graphql request for useUserInfoQuery when the user is unauthenticated, that makes no sense

helpers/useRedirectUnauthenticated.ts
JasirZaeem: I'm not sure I understand you.

This useEffect is not changing state, it will not cause any extra re-renders. This code runs at the same time as not using useEffect when the redirection has to occur, and does not the rest of the time. The body of this useEffect will at most, in the worst case, run as many times as not using useEffect, and in the typical case fewer times, depending on the condition used to decide wether to redirect.

This redirection is happening client side, after the app state changes. On this change a re render will happen with or without this useEffect and the redirection is happening after this render starts, currently (without) as well with the useEffect that is not avoidable. This render does not complete though in either case since as long the page location is changed the execution of the current page will halt.

This code is performing a side effect when a state changes, exactly what useEffect is meant for.

Some examples:
https://nextjs.org/docs/api-reference/next/router#usage
https://github.com/nextauthjs/next-auth/discussions/4078#discussioncomment-2270608

This useEffect can be removed still if it is too complicated and redirection will work, but useEffect is not changing when the redirection is happening/order of execution of its body. Redirection is happening after a re-render without it too since it can only happen after the state changes as that is prerequisite to deciding wether to redirect on the client.

anthonykhoa: useEffect only gets called on re renders. So if you implement this redirection logic with useEffect, its going to be dependent on needing a re-render in order for the redirection logic to kick in. It does not make sense to have our redirection logic be dependent on re-renders. Its an additional layer of complexity that provides no benefit

JasirZaeem: I see. Yeah sure this will happen after the render is done. But the render will just be the loading spinner the same as before. The benefit is not running on every render, but if you feel that is not worth the complexity then sure.

anthonykhoa: `The benefit is not running on every render`

We actually are not saving any extra pieces of logic from being run. After rendering finishes, useEffect will check the list of dependency values against the values from the last render, and will call your callback function if any one of them has changed. So the logic of checking still happens. Except it happens after every render rather than during.

Instead of `if (unauthenticated) router.push(...)` on every render

it'll be 
```
if (unauthenticated dependency has changed) {
 Do callback for redirection logic
}
```

after every render.

helpers/useRedirectUnauthenticated.ts
```diff
@@ -0,0 +1,20 @@
+import { useRouter } from 'next/router'
+import { LOGIN_PATH } from '../constants'
+
+/**
+ * Redirects to login page if user is not authenticated.
+ * With next query param to redirect back to this page after login.
+ * @param {boolean} shouldRedirect Condition for redirection
+ */
+const useRedirectUnauthenticated = (shouldRedirect: boolean) => {
```
anthonykhoa, line 9: Lets change this name, since its not being used a hook anymore

helpers/useRedirectUnauthenticated.ts
```diff
@@ -0,0 +1,20 @@
...
+  const router = useRouter()
```
anthonykhoa, line 10: Only declare this variable if the variable `shouldRedirect` is true. Otherwise its pointless declaring it

helpers/useRedirectUnauthenticated.ts
flacial: Why is that? It doesn't make sense because the whole purpose for prefixing a hook name with `use*` is to indicate whatever this variable is, in this case, it's a react hook. 

If a function **hooks** into react lifecycle methods, then the name should be starting with the `use` prefix, by convention.

helpers/useRedirectUnauthenticated.ts
flacial: Makes sense.

pages/settings/account/index.tsx
flacial, line 1: I found a bug. It should let users update individual fields instead of all of them together. If the username for instance is not changed, and the user only changes the name, it'll throw an error saying the username is already taken (by the user!)

constants/auth-constants.ts
```diff
@@ -0,0 +1,5 @@
+export enum SessionStatus {
+  Authenticated = 'authenticated',
+  Loading = 'loading',
+  unauthenticated = 'unauthenticated'
```
anthonykhoa, line 4: I just noticed that this name is lowercase but the other ones are pascal case.

flacial: Nice catch. I'll fix it ASAP.

pages/review/[lesson].tsx
```diff
@@ -37,28 +38,25 @@ const Review: React.FC<QueryDataProps<GetAppQuery>> = ({ queryData }) => {
   const context = useContext(GlobalContext)
   const slug = router.query.lesson as string
   const currentLesson = lessons.find(lesson => lesson.slug === slug)
+
   useEffect(() => {
+    redirectUnauthenticated(!session?.user)
```
Comment (anthonykhoa) [inline pages/review/[lesson].tsx:43]: Why did you put this inside useEffect?

pages/settings/account/index.tsx
```diff
@@ -262,16 +268,28 @@ const LinkedAccountsSetting = ({
...
+  const authenticated = status === SessionStatus.Authenticated
+  const unauthenticated = status === SessionStatus.Unauthenticated
   const { username, name, discordUsername } = data?.userInfo?.user || {}
 
-  if (loading || !data) {
+  useEffect(() => {
+    redirectUnauthenticated(unauthenticated)
```
anthonykhoa, line 280: Same here,  why is this in useEffect?

pages/review/[lesson].tsx
Comment (flacial) [reply pages/review/[lesson].tsx:43]: My understanding is that it's a _side-effect_, so it should be inside `useEffect`. Correct me if I'm wrong.

anthonykhoa (approved):

PR #2887, opened 2023-04-05
Fix overlapping act() calls error on `SubmissionComments.test.js`

scotthallock:
### Overview
When running the test suites, `SubmissionComments.test.js` logs two identical console errors:

Screenshot 2023-04-05 at 10 56 45 AM

### Changes
- In the `it('should allow user to edit comment and discard changes when edit button is pressed, but not edit more than 1 comment in a chain', ... )` test, remove the unneccessary nested `waitFor()` calls.
- The `findByText()` gets an `await` keyword in front of it because it is async
- The test suite no longer logs errors on the console:

Screenshot 2023-04-05 at 6 22 11 PM

### Notes
- We can be certain all 4 tests are running because of `expects.assertions(4)`
- [Related conversation about overlapping act calls](https://github.com/testing-library/react-testing-library/issues/605), and needing to `await` the `findBy*` queries.

### DIFF
--- a/components/SubmissionComments.test.js
+++ b/components/SubmissionComments.test.js
@@ -177,6 +177,7 @@ describe('Test SubmissionComments Component', () => {
     fireEvent.click(discardButton)
     expect(editButton).toBeInTheDocument()
   })
+
   it('should update comment when save changes button is clicked', async () => {
     const cache = new InMemoryCache({ addTypename: false })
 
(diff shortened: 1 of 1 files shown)


components/SubmissionComments.test.js
```diff
@@ -160,28 +160,26 @@ describe('Test SubmissionComments Component', () => {
...
+    expect(
+      await findByText(
         'You can only edit one comment in a single comment chain at a time.'
+      )
+    ).toBeTruthy()
 
-      waitFor(() => {
-        expect(findByText(warningText)).toBeTruthy()
-      })
+    fireEvent.click(discardButton)
```
flacial, line 177: FYI, you'll notice that `fireEvent` is used in some files in the codebase, but `userEvent` is preferred because it's closer to how a user actually interacts with an element.

flacial (approved):
Well done fixing these concerning console screams!

PR #2889, opened 2023-04-06
Fix bad setState call causing error in `pages/reset.test.js`

scotthallock:
### Overview

When running the test suites, `pages/reset.test.js` logs a console error. This PR fixes the bad setState() call so that the error is fixed.

The error:
> Cannot update a component (`ResetPasswordContainer`) while rendering a different component (`ResetPassword`). To locate the bad setState() call inside `ResetPassword`, follow the stack trace ...

Screenshot 2023-04-05 at 10 11 29 PM

### Changes

- Wrap the setState() call (`setServerError()`) inside of a useEffect so that it runs *after* the `ResetPassword` component renders instead of *during* the render.
- The useEffect call is not given a 2nd argument. We want it to run every time `ResetPassword` renders.
- The "Cannot update a component while render a different component" error is fixed.

Screenshot 2023-04-05 at 10 19 13 PM

### Notes
- [Related conversation about this error](https://stackoverflow.com/questions/62336340/cannot-update-a-component-while-rendering-a-different-component-warning)

### DIFF
--- a/pages/confirm/[token].tsx
+++ b/pages/confirm/[token].tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import { useRouter } from 'next/router'
 import { Formik, Form, Field } from 'formik'
 import Link from 'next/link'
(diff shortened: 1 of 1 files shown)


flacial (approved):
Looks good!

One thing I thought about, I think the code would be robust if the `confirmToken` was added to the dependencies of `useEffect`. If any engineer worked on this component and added a new state, the `useEffect` won't be _affected_ by it.

PR #2894, opened 2023-04-10
fix inconsistent font styles

### LINKED ISSUE #2824: Improve the font weight in the landing page
## Problem

The h1 and the paragraph elements have the same weight (bold). It looks off because it's common for headings to be bold and big to make it easier to distinguish the main points from the rest. Paragraphs could be big, but their weight would be less than headings; generally, they should be less noticeable than headings. This applies to all the other paragraphs or sub-headings.

## Expected design
image

> could be inaccurate to what we've in the app because it uses a very high Rubik weight and paragraphs use Inter instead of Rubik

## Actual design

image

scotthallock:
## Overview

This PR fixes an inconsistent UI issue raised in #2824:
 - Large paragraphs have `bold Rubik` font
 - Small paragraphs have `normal Inter` font

## Changes
- On the landing page (`components/LandingPage.tsx`)
  - remove the "h4" class on the `<p>` elements
  - add `fontSize: '1.5rem'` to styles of `<p>` elements that are associated with `<h1>` headings
    - the last `<p>` element on this page ("Become a full stack software engineer - 100% Free") already had this style
  - After these changes:
    - the font of the headings is `bold Rubik` (unchanged)
    - the font of all paragraph elements is `normal Inter`
  
### Before:
Screenshot 2023-04-10 at 3 56 00 PM
Screenshot 2023-04-10 at 3 58 01 PM

### After:
Screenshot 2023-04-10 at 3 55 41 PM
Screenshot 2023-04-10 at 3 58 16 PM

## Notes
- Closes #2824

### DIFF
--- a/components/LandingPage.tsx
+++ b/components/LandingPage.tsx
@@ -25,7 +25,10 @@ const LandingPage: React.FC = () => {
         </div>
         <div className="row justify-center mt-4">
           <div className="col-md-10 offset-md-1">
-            <p style={{ lineHeight: 1.5 }} className="fw-light h4">
+            <p
+              style={{ lineHeight: 1.5, fontSize: '1.5rem' }}
+              className="fw-light"
+            >
               Learn the right foundations you need to become a full stack
               software engineer. Our curriculum takes no shortcuts and
               effectively trains students with no technical background to become
@@ -49,7 +52,10 @@ const LandingPage: React.FC = () => {
         </div>
         <div className="row mt-5">
           <div className="col-md-10 offset-md-1">
-            <p className="fw-light h4" style={{ lineHeight: 1.5 }}>
+            <p
+              style={{ lineHeight: 1.5, fontSize: '1.5rem' }}
+              className="fw-light"
+            >
               Our learning process is interactive and follows the same practice
               as a well functioning engineering team. From the first line of
               code you write, you will be code reviewed by engineers to ensure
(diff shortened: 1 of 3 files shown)


components/LandingPage.tsx
```diff
@@ -25,7 +25,10 @@ const LandingPage: React.FC = () => {
         </div>
         <div className="row justify-center mt-4">
           <div className="col-md-10 offset-md-1">
-            <p style={{ lineHeight: 1.5 }} className="fw-light h4">
+            <p
+              style={{ lineHeight: 1.5, fontSize: '1.5rem' }}
```
flacial, line 29: I think the file was already repeating styles before you work on it, but it's worth noting that it now seems to require a module CSS file so we don't have to repeat the styles for the `p` element or any other one.

If it already has one, we could create a class there with these styles and use it here.

flacial (approved):

components/LandingPage.tsx
scotthallock: Good idea. I don't see a module CSS file for the landing page, but I can create one and update this PR

scotthallock:
@flacial I moved the `LandingPage` component to its own folder and created a scss module for it. Could you review the changes?

flacial (approved):
This looks way better. Thanks for applying the suggestion!

PR #2947, opened 2023-04-28
deps: Upgrade discord.js from v12 to v14

flacial:
### Problem
To prevent any API mismatches and ensure the smooth functioning of c0d3r, this pull request updates the discord.js version of c0d3-app to v14, which aligns with the upcoming update of the bot.

### Changes

- Updated discord.js dependency in package.json from v12.5.3 to v14.9.0
- Modified the `constants/index.ts` file to change the `PRIMARY_COLOR_HEX` constant to a hexadecimal number.
- Modified the `graphql/resolvers/addComment.ts` file to change the `MessageEmbedOptions` import to `EmbedBuilder` and update the `notificationEmbed` and `reviewNotificationEmbed` objects to match the `EmbedBuilder` format.
- Modified the `helpers/updateSubmission.ts` file to change the `MessageEmbedOptions` import to `EmbedBuilder` and update the `reviewNotificationEmbed` object to match the `EmbedBuilder` format.

### Related PRs

- Closes https://github.com/garageScript/c0d3-app/issues/2948

### Notes

Depends on #2949


graphql/resolvers/addComment.ts
```diff
@@ -57,7 +57,7 @@ export const addComment = async (
   const { author, submission } = comment
 
   if (author.id !== submission.user.id && submission.user.discordId) {
-    const notificationEmbed: MessageEmbedOptions = {
+    const notificationEmbed: EmbedBuilder['data'] = {
```
scotthallock, line 60: Question my own understanding: is `EmbedBuilder['data']` is a type annotation, right?

Is there a reason for using bracket notation instead of dot notation (`EmbedBuilder.data`), or will it work either way?

JasirZaeem:
The embed builder from constructor expects the type `APIEmbed` https://old.discordjs.dev/#/docs/discord.js/14.9.0/class/EmbedBuilder?scrollTo=s-from so I think that might be the better type to use instead of `EmbedBuilder['data']` since this data will be used to create an Embed using this method when received by c0d3r.

JasirZaeem:
Types here also need to be updated https://github.com/garageScript/c0d3-app/blob/master/%40types/discordBot.d.ts can you check once that all places where MessageEmbedOptions was used gets updated.

graphql/resolvers/addComment.ts
JasirZaeem: Indexed access is how we can look up a specific property on another type, this is a typescript feature, `EmbedBuilder.data` will not work here. https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html

`EmbedBuilder` class has a property `data`, when using `EmbedBuilder` as a type (like here) if you want to access the type of `data` property, you will have to use the indexed access, dot access will not work.

I have created this [example](https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBYKodp0b6CRqgAC8oAA3mJmJB8v12IhpN8ANz9BjIDmCVCUTjMfkMaC6dnjeDwFpeAC+vMUlKIlTZDOpP0aAG0AEQMNkGgC6qvVKAAchUtYzaDTGgFjZBVcUcM0PGQvIh4H0GMD2DQvm4zMx2C7+LBpbBEgH4HoRF4DQBRVaChb6A1JbEG8AOVAAcmEVsgEgNQRUAHkvmR3oJIPw2UdUI4FK5pHhQE5fTphlh9CgEIhocJLiGJH7oiIDBxsQsmKwODwgiUgA) as you can see that `type Age` got correctly defined as `number` but `NAge` did not.

flacial:
> Types here also need to be updated https://github.com/garageScript/c0d3-app/blob/master/%40types/discordBot.d.ts can you check once that all places where MessageEmbedOptions was used gets updated.

Nice catch. Seems like our tsc configs doesn't check *.d.ts files.

JasirZaeem (approved):

scotthallock (commented):
LGTM! :)

PR #2985, opened 2023-05-15
feat(settings/account): Show a meaningful msg on query error

flacial:
## Changes

The `useUserInfoLazyQuery` hook has been modified to include an `error` property in its response object, which is then passed to `QueryInfo` to display any error messages. The `AccountSettings` component has also been updated to include the `QueryInfo` component.

These changes aim to improve the user experience by providing better feedback when there are issues with retrieving user information.

### DIFF
--- a/pages/settings/account/index.tsx
+++ b/pages/settings/account/index.tsx
@@ -268,7 +268,7 @@ const LinkedAccountsSetting = ({
 
 const AccountSettings = () => {
   const { data: session, status } = useSession() as SessionContext
-  const [userInfoQuery, { data, loading: userInfoLoading }] =
+  const [userInfoQuery, { data, loading: userInfoLoading, error }] =
     useUserInfoLazyQuery()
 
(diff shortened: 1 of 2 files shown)


pages/settings/account/index.tsx
```diff
@@ -298,6 +298,17 @@ const AccountSettings = () => {
...
           <h1>Settings</h1>
+          <QueryInfo
+            data={null}
+            loading={false}
+            error={error?.message || ''}
+            texts={{
+              loading: '',
+              data: '',
+              error:
+                'Oops, something went wrong when getting your info. Please try again'
```
SlyBouhafs, line 309: Should we also start organizing our error messages in a constant file at this point?
NIT: The error message would sound better like this 'We encountered an issue while retrieving your information. Please try again.'

flacial: I agree. Good idea!

SlyBouhafs (approved):

Issue #2849, opened 2023-03-23
Add Reviewer Notification Functionality to Discord Bot


flacial:
### Overview 

Currently, our Discord bot only sends notifications to students when new comments are posted on their submissions. However, we want to ensure that reviewers are also notified when their comments receive a response. To achieve this, we need to add functionality to the bot that sends a notification to the reviewer when their message is responded to.

To implement this functionality, we'll need to add the ability for the bot to mention the reviewer in the notification message. We'll also need to include a link to the message that was responded to so that the reviewer can easily navigate to it.

This feature will improve the review process and ensure that all parties involved are notified when there is new activity on a submission.

## Proposed implementation

- Add the ability for the bot to mention reviewers in notification messages
- Include a link to the message that was responded to in the notification message
- Test the new functionality thoroughly to ensure it works as expected

## Expected Outcome

Reviewers will receive notifications when their messages receive a response on Discord. This will improve communication and collaboration during the review process.

> generated with ChatGPT

scotthallock:
If nobody is currently working on this, I'd like to give it a shot

PR #2997, opened 2023-05-19
feat: Send a notification to reviewers on new comment

flacial:
## Overview

Currently, when a student receives a comment from a reviewer on their submission and attempts to reply to it, the reviewer doesn't get a notification. This behavior introduces difficulty in communication between the student and the reviewer, as the reviewer's only way to find out about the new comments by the student is to frequently visit the website. It can quickly degrade productivity for both the student who is trying to learn and the reviewers who are trying to teach.

### Solution

Send a notification to all the people who wrote a comment on a certain submission's line except the comment author.

## Missing requirements

> comment section: comment(s) under a certain line in a submission

There are some things this PR does not introduce that are crucial for having an ideal notification system.

1. Participants in a comment section can't opt out of receiving notifications about it.
2. Participants cannot reply to or mention a user in a specific comment.

These shall be addressed in the project of adding a notification system (I could not find the issue for this project, but it did exist). @SlyBouhafs, any references?

### Changes

- Export the logic for sending a notification to a function in the `addComment` resolver.
- Update the logic of sending a notification to send multiple ones by querying all the users who comment on a certain line or submission and sending a notification to them as long as they have a Discord ID.
- Update the message. Discord embedding has two versions. One is for the reviewers, and the other is for the student.

### Unrelated changes

These changes are not related to this PR, but they were required in order for the CI/CD to pass

- Fix the flaky test in the `exercises/[lessonSlug].tsx` page by waiting for the router to be ready

## Ideal solution

I believe this is not the ideal solution because sending simultaneous requests to our Discord bot server could overload the server and, by extension, slow down the server for other users. As of now, we don't have a lot of active users, and the comment section is almost always 1:1.

## Related PR

- #2849

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,4 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
+# #!/bin/sh
+# . "$(dirname "$0")/_/husky.sh"
 
-npm run type-check && npm test
+# npm run type-check && npm test
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,4 @@
-# #!/bin/sh
-# . "$(dirname "$0")/_/husky.sh"
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
 
-# npm run type-check && npm test
+npm run type-check && npm test
--- a/graphql/resolvers/addComment.ts
+++ b/graphql/resolvers/addComment.ts
@@ -54,7 +54,8 @@ const sendNotifications = async ({
           discordId: true
         }
       }
-    }
+    },
+    distinct: ['authorId']
   })
 
   const { submission, author } = comment
--- a/graphql/resolvers/addComment.ts
+++ b/graphql/resolvers/addComment.ts
@@ -46,8 +46,13 @@ const sendNotifications = async ({
       submissionId,
       fileName
     },
-    include: {
-      author: true
+    select: {
+      author: {
+        select: {
+          id: true,
+          discordId: true
+        }
+      }
     }
   })
 
--- a/graphql/resolvers/addComment.ts
+++ b/graphql/resolvers/addComment.ts
@@ -11,7 +11,8 @@ import {
   PRIMARY_COLOR_HEX,
   PROFILE_URL
 } from '../../constants'
-import { Challenge, Comment, Lesson, Submission, User } from '@prisma/client'
+import { Challenge, Comment, Lesson, Submission } from '@prisma/client'
+import { MessageResponse } from '../../@types/discordBot'
 
 type CommentWithAuthorAndSubmission = Comment & {
   author: {
@@ -56,23 +57,6 @@ const sendNotifications = async ({
     }
   })
 
-  const [uniqueParticipants] = restOfComments?.reduce(
-    (
-      acc: [(Comment & { author: User })[], { [key: string]: boolean }],
-      e: Comment & { author: User }
-    ) => {
-      const [arr, map] = acc
-
-      if (!map[e.authorId]) {
-        arr.push(e)
-        map[e.authorId] = true
-      }
-
-      return acc
-    },
-    [[], {}]
-  )
-
   const { submission, author } = comment
 
   const notificationEmbed = (isSubmissionOwner: boolean): APIEmbed => {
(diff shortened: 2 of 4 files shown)


graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,129 @@ import {
...
+  comment: CommentWithAuthorAndSubmission
+}
+
+const sendNotifications = async ({
+  line,
+  submissionId,
+  fileName,
+  comment
+}: SendNotificationsParams) => {
+  const restOfComments = await prisma.comment.findMany({
```
JasirZaeem, line 43: Do not request all the comment data, you only need the author id, author username and author discordIt, no other field is needed on the author on comment, only request the minimum amount of needed data.

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,129 @@ import {
...
+      line,
+      submissionId,
+      fileName
+    },
+    include: {
+      author: true
+    }
+  })
+
+  const [uniqueParticipants] = restOfComments?.reduce(
```
JasirZaeem, line 54: Prisma allows selecting distinct values for a column, use that instead.

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,129 @@ import {
...
+  if (author.id !== submission.user.id && submission.user.discordId) {
+    await sendDirectMessage(
+      submission.user.discordId,
+      '',
+      notificationEmbed(true)
+    )
+  }
+
+  // sends a notification to all of those who commented on
+  // the same submission line except the comment author
```
JasirZaeem, line 119: This should say `except the comment author and the student who made the submission`

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,129 @@ import {
...
+  uniqueParticipants.forEach(participant => {
+    if (
+      participant.author.id !== author.id ||
+      participant.author.id === submission.user.id
+    ) {
+      return
+    }
+
+    if (participant.author.discordId) {
+      sendDirectMessage(
```
JasirZaeem, line 129: These calls need to be awaited, or the serverless function may end prematurely and the request will be cancelled.

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,129 @@ import {
...
+          name: 'Comment',
+          value: comment.content
+        }
+      ]
+    }
+  }
+
+  // sends a notification to the submission author
+  if (author.id !== submission.user.id && submission.user.discordId) {
+    await sendDirectMessage(
```
JasirZaeem, line 111: Don't have to await this particular notification separately before the others are sent, schedule them all and then await them all.

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,129 @@ import {
...
+      return acc
+    },
+    [[], {}]
+  )
+
+  const { submission, author } = comment
+
+  const notificationEmbed = (isSubmissionOwner: boolean): APIEmbed => {
+    const params = {
+      title: isSubmissionOwner
```
JasirZaeem, line 75: Is this doing anything ? submission owner does not get notified of the comments they make so this will never be true.

flacial: I think it should've been "forSubmissionOwner" instead. If the flag is true, a customized embed message is returned to the submission owner. It's only true when someone, such as a reviewer (someone who isn't the owner of the submission), leaves a comment on their submission.

See line 114 when the comment author isn't the submission owner.

JasirZaeem: I see got it, yeah I think `forSubmissionOwner` would make the purpose clearer then sure you can go with that

.husky/pre-push
JasirZaeem, line 1: I think these changes are accidental

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,137 @@ import {
...
+    const notifications = restOfComments.reduce(
+      (acc: Promise<MessageResponse>[], participant) => {
+        if (
+          participant.author.id !== author.id ||
+          participant.author.id === submission.user.id
+        ) {
+          return acc
+        }
+
+        if (participant.author.discordId) {
```
JasirZaeem, line 121: Combine this condition with the if above as well (and change is accordingly). If the above if is true or if this if is false, the same thing is happening (acc is being returned). And this if does not depend on the above check, so you can make this simpler and reduce the number of paths this code can take here.

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,137 @@ import {
...
+    }
+  }
+
+  // sends a notification to all of those who commented on
+  // the same submission line except the comment author and the student who made the submission
+  const sendNotificationsToParticipants = async () => {
+    const notifications = restOfComments.reduce(
+      (acc: Promise<MessageResponse>[], participant) => {
+        if (
+          participant.author.id !== author.id ||
```
JasirZaeem, line 115: I think this should be `participant.author.id === author.id` let me know if I'm missed something.

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,137 @@ import {
...
+  comment: CommentWithAuthorAndSubmission
+}
+
+const sendNotifications = async ({
+  line,
+  submissionId,
+  fileName,
+  comment
+}: SendNotificationsParams) => {
+  const restOfComments = await prisma.comment.findMany({
```
JasirZaeem, line 44: You need to add the distinct option here on `authorId` to make prisma select an author only once https://www.prisma.io/docs/concepts/components/prisma-client/aggregation-grouping-summarizing#select-distinct

graphql/resolvers/addComment.ts
flacial: You're right, that's a typo.

graphql/resolvers/addComment.ts
flacial: Strange. I'm certain I added this option and staged the changes.

flacial: Anyway, I'll add it now.

graphql/resolvers/addComment.ts
```diff
@@ -11,6 +11,135 @@ import {
...
+  // sends a notification to all of those who commented on
+  // the same submission line except the comment author and the student who made the submission
+  const sendNotificationsToParticipants = async () => {
+    const notifications = restOfComments.reduce(
+      (acc: Promise<MessageResponse>[], participant) => {
+        const isCommentAuthor = participant.author.id === author.id
+        const isSubmissionAuthor = participant.author.id === submission.user.id
+        const hasDiscordId = participant.author.discordId
+
+        if ((!isCommentAuthor || !isSubmissionAuthor) && hasDiscordId) {
```
JasirZaeem, line 119: This condition is not correct. `if ((user is not comment author or user is not submission author) and they have a discord id) send notification` so if I am not the submission author, but I am the comment author, I'll still get notified which should not happen since I know I made the comment.

JasirZaeem (approved):

PR #2998, opened 2023-05-19
test: Attempting to build storybook v7 on our CI server

flacial:
### What's happening

Building storybook v7 uses all my RAM (possibly memory leak) and it end up crashing. I'm trying to build it on our CI/CD server to see if it has something to do 
with my machine.


flacial:
Seems like a memory leak.

Issue #2321, opened 2022-09-17
Redesign the lesson content page


flacial:
### Things to redesign

<details>
<summary>
Lesson parts nav
</summary>
<img width="453" alt="image" src="https://user-images.githubusercontent.com/35906419/190858048-f5caf6a1-117d-4bf4-a8fc-85faa4eb2ecd.png">
</details>

<details>
<summary>
Elements spacing & letter spacing
</summary>
<img width="966" alt="image" src="https://user-images.githubusercontent.com/35906419/190858103-2beae61f-88e8-40b3-8e90-c681dcd9606d.png">
<img width="1095" alt="image" src="https://user-images.githubusercontent.com/35906419/190858116-a1710e40-4f00-4859-88c4-66e66a7de9e1.png">
</details>

<details>
<summary>
Fonts
</summary>
The fonts are inconsistent. The should be `Inter` font

<img width="826" alt="image" src="https://user-images.githubusercontent.com/35906419/190858173-4d4f89ac-78af-4e6a-a11f-2eca8fb45ef8.png">
</details>

<details>
<summary>
Challengebar font, size, colors, weight, and letters spacing
</summary>
<img width="774" alt="image" src="https://user-images.githubusercontent.com/35906419/190858216-ba33de7e-e9e5-4c3a-8df2-d23d3564ba4d.png">
</details>

<details>
<summary>
Table of content
</summary>
It's too large
<img width="400" alt="image" src="https://user-images.githubusercontent.com/35906419/190858259-f0d1e6da-0b13-4149-ad74-2de7a9d5f88f.png">
</details>

<details>
<summary>
Text length
</summary>
It's hard to read lengthy text

<img width="1319" alt="image" src="https://user-images.githubusercontent.com/35906419/190858312-1b64eb21-b94d-49c9-81d9-441dbc0ea766.png">
</details>

---

We can use the [design mock](https://zpl.io/7wzQ7d8) we created in the past for this page. I think it would do the job.

akshayishaan:
Is anyone working in this issue? I want to work on this issue. Can @flacial explain a bit more about it?

flacial:
> Is anyone working in this issue? I want to work on this issue. Can @flacial explain a bit more about it?

[...]

PR #3080, opened 2023-06-26
Improve the lesson sidebar design

flacial:
### Overview

This PR polishes and redesign the lesson sidebar by putting it inside a container to differentiate it from other elements, making it sticky so the student can easily navigate the lesson sections, and finally removing the `remark-toc` as it's now replaced with an in-house solution.

### Changes

- Extract the headings from an mdx source using `gray-matter`
- Update the sidebar UI
- Add new tests cases to the `subLesson` test file to cover the new changes
- Remove the text "Table of Contents" from all mdx files so it doesn't show in the compiled source
- Update snapshots

#### Before

image

#### After

image

#### On mobile

image

### Subsequent changes

- Remove the package `remark-toc`
- Redesign the mobile view to make it collapsable
- Highlight the currently viewed section in the sidebar

### Related issues

- #2321

### DIFF
--- a/pages/curriculum/[lessonSlug]/[subLessonSlug].tsx
+++ b/pages/curriculum/[lessonSlug]/[subLessonSlug].tsx
@@ -20,9 +20,12 @@ import SubLessonLinks from '../../../components/SubLessonLinks'
 import EditPage from '../../../components/EditPage'
 import MDXcomponents from '../../../helpers/mdxComponents'
 
-import styles from '../../../scss/mdx.module.scss'
+import mdxStyles from '../../../scss/mdx.module.scss'
+import styles from '../../../scss/subLessonSlug.module.scss'
 import ScrollTopArrow from '../../../components/ScrollTopArrow'
 import Title from '../../../components/Title'
+import matter from 'gray-matter'
+import useBreakpoint from '../../../helpers/useBreakpoint'
 
 interface Props {
   selectedSubLessonIndex: number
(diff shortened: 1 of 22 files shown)


pages/curriculum/[lessonSlug]/[subLessonSlug].tsx
```diff
@@ -130,15 +172,25 @@ export const getStaticProps: GetStaticProps<any, Slugs> = async context => {
...
         // Only include source data on selected subLesson
         const sourceAndFrontMatter = await parseMDX(
-          await getSubLessonContent(slug),
+          source,
           slug.subLessonSlug !== subLessonSlug
         )
 
+        const { content } = matter(source)
+        const headings = content?.match(/^(#+)\s+(.*)$/gm)?.map(heading => {
+          // istanbul ignore next
```
Comment (flacial) [inline pages/curriculum/[lessonSlug]/[subLessonSlug].tsx:185]: I need help here. Jest is complaining that this line isn't covered, but I couldn't figure out how because if the regex for `content` doesn't match any of the string and returns `null`, the regex for `heading` won't run because they're checking for almost the same pattern: if the substring contains a hash symbol with a space between it and another piece of text.

Comment (JasirZaeem) [reply pages/curriculum/[lessonSlug]/[subLessonSlug].tsx:185]: Use a matchAll, to get the capturing groups along with the matches the first time you run the regex instead of doing it twice.

like this
```js
const headings = [...content?.matchAll(/^(#+)\s+(.*)$/gm)].map(
  ([, hashes, text]) => ({
    text,
    depth: hashes.length,
  })
);
```

JasirZaeem (approved):

PR #3112, opened 2023-07-11
Removed a debrief section related to for loops

bishwasBhatta:
This debrief section is describing the use of i and j variables, most probably for nested for loops used for solving the 2D arrays problem. Must be remnants from before the switch to "no loops, only recursion" rule.


flacial (approved):
This LGTM, but I think it'd be better to rewrite the debrief to refer to a recursive loop.

PR #3302, opened 2023-10-17
Update codebase and CI to use node v18

SlyBouhafs:
Node v16 has reached its [end of life](https://endoflife.date/nodejs) past September, this PR updates node version to v18.

Changes not included in the diff are:

- Update Vercel from node v16 to node v18.
- Update Netlify from v16 to v18.


flacial (approved):

SlyBouhafs:
Updating node to v18 breaks Storybook V6, we must upgrade storybook to V7.
To resolve this, we need to upgrade Storybook to V7. Previous attempts to upgrade to Storybook V7 resulted in compatibility issues that prevented Storybook from building. I will give it another try and see if those compatibility arise again.

